PR #1, opened 2016-10-14
Make it possible to specify the jmx port through the UI

michaelsembwever:
The seed node becomes "hostname|ip[:jmx-port]"

No hurry to merge this. Think it needs further thought on simplest approach forward.


beevee:
Is this going to be merged anytime soon? I have a very real usecase: two Cassandra nodes on the same host, with different JMX ports. Right now I have to use several instances of Reaper, because YAML config allows to specify only one JMX port per IP address.

adejanovski:
Hi @beevee,

you can override ports for specific ip addresses as it is done in the resource files here : https://github.com/thelastpickle/cassandra-reaper/blob/master/resource/cassandra-reaper-cassandra.yaml#L17-L25

Let us know if that suits your requirements.

beevee:
Sure, but can I specify two ports for one IP address this way? I have two separate Cassandra nodes on one physical machine, with JMX ports confugured to be 7199 and 7299.

adejanovski:
Both nodes are using the same ip and hostname ? 
Did you configure different ports for all interfaces (internode, rpc, jmx) ? 

What is your use case for running several nodes on the same machine ? Is it for test purposes only ? Are those two nodes part of the same cluster ?

beevee:
Yes, both nodes use same ip and hostname, but different ports for all interfaces.

We have several lightly loaded 3-node clusters, that are too small to buy a separate hardware machine for. And we've had some negative experience running Cassandra on VMs, so we just run several nodes from different clusters on one physical host with different ports. It's pretty easy and convenient to install, maintain and monitor with our Ansible scripts. These clusters are in production environment.

Of course, we can run separate Reapers for each cluster, but single web UI would be much more convenient.

Issue #8, opened 2016-10-20
problem starting incremental repair when full repair was executed earlier or vice-versa


aggarwalabhi23:
while creating new run, if the passed value of incremental repair flag is different from the existing value  then it should allow to create new repair_unit instead of getting repair_unit based on cluster name/ keyspace /column combination.

so, if the full run was executed for the clustername/keyspace/column was executed earlier, not able to start new incremental run for the same combination. Reverse is also true.

So as per current solution if i delete the repair_unit then due to referential constraints i need to delete repair_segment and repair_run as well which will delete the run history corresponds to the repaid_unit.

zznate:
Given the potential performance impact (run-time and per repair) of flipping to and from incremental, treating this as a completely new `repair_unit` seems reasonable to me.

adejanovski:
@aggarwalabhi23 : I've fixed this in PR #9 
Can you check that the fix works for you before we merge ?

adejanovski:
Fix merged

PR #9, opened 2016-10-21
Fix for issue #8 (problem starting incremental repair when full repair was executed earlier or vice-versa)

adejanovski:
Allows creating different repair units for the same cluster/keyspace with different incremental repair settings.

Added a scenario in acceptance tests to validate the behavior, and also forced repair parallelism to PARALLEL with incremental repair, given it's the only one allowed.

### DIFF
--- a/src/main/java/com/spotify/reaper/resources/RepairRunResource.java
+++ b/src/main/java/com/spotify/reaper/resources/RepairRunResource.java
@@ -148,7 +148,7 @@ public class RepairRunResource {
       RepairUnit theRepairUnit =
           CommonTools.getNewOrExistingRepairUnit(context, cluster, keyspace.get(), tableNames, incrementalRepair);
 
-      if(theRepairUnit.getIncrementalRepair() != incrementalRepair) {
+      if (theRepairUnit.getIncrementalRepair() != incrementalRepair) {
     	  return Response.status(Response.Status.BAD_REQUEST).entity(
                   "A repair run already exist for the same cluster/keyspace/table but with a different incremental repair value." 
                   + "Requested value: " + incrementalRepair + " | Existing value: " + theRepairUnit.getIncrementalRepair()).build();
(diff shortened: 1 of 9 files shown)


src/main/java/com/spotify/reaper/resources/RepairRunResource.java
```diff
@@ -161,12 +161,10 @@ public Response addRepairRun(
         parallelism = RepairParallelism.valueOf(repairParallelism.get().toUpperCase());
       }
       
-      if(!parallelism.equals(RepairParallelism.PARALLEL) && incrementalRepair) {
-          return Response.status(Response.Status.BAD_REQUEST).entity(
-                  "It is not possible to mix sequential repair and incremental repairs. parallelism " 
-                  + parallelism + " : incrementalRepair " + incrementalRepair).build();
+      if (incrementalRepair) {
+        parallelism = RepairParallelism.PARALLEL;
```
zznate, line 165: Should we just step on the parallelism option or return with a validation error when incremental and parallel are true? Either way, we should add in to javadoc saying we change it.

src/main/java/com/spotify/reaper/service/SegmentGenerator.java
```diff
@@ -156,7 +157,7 @@ public static boolean greaterThanOrEqual(BigInteger a, BigInteger b) {
       BigInteger size = segment.span(RANGE_SIZE);
       total = total.add(size);
     }
-    if (!total.equals(RANGE_SIZE)) {
+    if (!total.equals(RANGE_SIZE) && !incrementalRepair) {
```
zznate, line 160: Is this necessarily the case? And feels weird that this is the only place we use incrementalRepair argument.

zznate (commented):
Testing looks fine. Im not quite sure I understand SegmentGenerator#generateSegments and it's need for/use of `incrementalRepair`

src/main/java/com/spotify/reaper/service/SegmentGenerator.java
adejanovski: The thing is that we are tricking reaper with incremental repair, as our repair unit is no longer segments but nodes, and generate one bogus segment per node. 
This check started to fail after I added a test scenario including incremental repair and as it was not relevant I chose to disable the check with incremental repair.

src/main/java/com/spotify/reaper/resources/RepairRunResource.java
adejanovski: I feel like if PARALLEL is the only option for incremental repair, we should just force it instead of asking for the user to pick the right one.
Furthermore, the UI doesn't require this field to be set and then SEQUENTIAL is selected as the default, which is not compatible with inc repair.

michaelsembwever:
fixed title. it's nice to have readable email subject lines :-)

zznate:
Those both sound fair. Thanks @adejanovski

PR #25, opened 2016-12-01
Include UI sources and maven profile to rebuild it

adejanovski:
Now that we're embedding the UI we need to make it live inside the same repository.
I've created a separate maven profile so that people not wanting to mess with NPM and Bower won't have to. 

Readme is updated to include build commands for various needs.


spodkowinski:
How would you be able to merge changes from the ui project at this point? If you really like to integrate the ui sources, you could as well add them as git sub-module.

adejanovski:
We actually feel that the UI is fully part of the project now instead of being an addon and that its sources shouldn't live outside the repo. 
We won't tie it to the original project anymore, and changes will be contributed directly in this repo.
This should allow to make changes faster if we add features that require both the backend and the frontend to be updated.

PR #36, opened 2016-12-23
Fix a NullPointerException when postponing

vrischmann:
Hello,

I'm trying to use the Cassandra storage for Reaper and just encountered this bug: `SegmentRunner.postpone` would always fail because the method `updateRepairSegment` never checks that `newRepairSegment.getStartTime()` returns something.

### DIFF
--- a/src/main/java/com/spotify/reaper/storage/CassandraStorage.java
+++ b/src/main/java/com/spotify/reaper/storage/CassandraStorage.java
@@ -5,6 +5,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.Date;
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


vrischmann:
The build failed but it doesn't look like it's from my modifications.

    [ERROR] Failed to execute goal org.sonarsource.scanner.maven:sonar-maven-plugin:3.2:sonar (default-cli) on project cassandra-reaper: You're not authorized to execute any SonarQube analysis. Please contact your SonarQube administrator. -> [Help 1]

adejanovski:
@vrischmann : the build fails because Travis prevents using env vars that contain private tokens on external PRs. 
That's understandable for security reasons (anyone could expose the tokens), but sucks big time as we cannot perform Sonar analysis on the submitted code.

Anyway, thanks for the fix ! 

Merged.

PR #69, opened 2017-03-30
Including jackson-datatype-jdk8 and the JDK8Module

chrislbs:
This might not be necessary if we go ahead and merge #66


adejanovski:
Hi @chrislbs , 

I've merged #66 : could you check that it fixes your issue ?

Thanks !

chrislbs:
That merge also resolved the issue. Thanks @adejanovski

Issue #72, opened 2017-04-03
Update Documentation for Cassandra Storage with Auth


skhatri:
Update documentation to show configuration for reaper with storage type of cassandra.

PR #73, opened 2017-04-03
additional example and doc around cassandra storage for reaper

skhatri:
For issue: https://github.com/thelastpickle/cassandra-reaper/issues/72


adejanovski:
Thanks for the PR @skhatri, merged !

Issue #94, opened 2017-05-08
Cassandra performance improvements


michaelsembwever:
Relevant to https://github.com/thelastpickle/cassandra-reaper/issues/85
---

**Observations:**
 - frequent full table scans on `cluster` 
 - UI pages hit the rest endpoints like  `/repair_run/state`, causing scheduled 10s repeated selects on `repair_run`, `repair_unit`, and `repair_run_by_cluster`,
 - small granularity of tables: no/little utilisations of clustering keys and common keys,
 - manual denormalisation over 2nd indexes,
 - large volumes of cql requests (selects and inserts) during creation of a repair run (before activation).

The rest endpoint `/repair_run/state` repeats the follow every 10 seconds…
 ```
SELECT * FROM cluster;
  1-> SELECT * FROM repair_run_by_cluster WHERE cluster_name = ?;
    N-> SELECT * FROM repair_run WHERE id = ?; 
    N-> SELECT * FROM repair_unit WHERE id = ?;
      1-> SELECT * FROM repair_segment_by_run_id WHERE run_id = ?;
      1-> SELECT * FROM repair_segment WHERE id = ?;
```

During the creation of a repair run, non-incremental with 1467 segments, ~14k requests were logged.
A distribution of these requests is as follows:
```
   8235  SELECT * FROM repair_segment WHERE id = ?;
   1467  UPDATE repair_id SET id=N WHERE id_type = 'repair_segment' IF id = N;
   1467  SELECT id FROM repair_id WHERE id_type = 'repair_segment';
   1467  INSERT INTO repair_segment(id, repair_unit_id, run_id, start_token, end_token, state, coordinator_host, start_time, end_time, fail_count) VALUES(?, ?, ?, ?, ?, ?, ?, ?, ?, ?);
   1467  INSERT INTO repair_id (id_type, id) VALUES('repair_segment', N) IF NOT EXISTS;
     14  SELECT * FROM cluster;
[...]

PR #96, opened 2017-05-09
Cassandra performance: Add LCS and row cache

michaelsembwever:
ref:
 - https://github.com/thelastpickle/cassandra-reaper/pull/96
 - https://github.com/thelastpickle/cassandra-reaper/issues/94

### DIFF
--- a/src/main/resources/db/cassandra/002_table_properties.cql
+++ b/src/main/resources/db/cassandra/002_table_properties.cql
@@ -1,0 +1,41 @@
+
+-- Optimal table properties
+
+ALTER TABLE cluster
+    WITH compaction = {'class': 'LeveledCompactionStrategy'}
+    AND caching = {'rows_per_partition': 'ALL'};
+
+ALTER TABLE repair_unit
+    WITH compaction = {'class': 'LeveledCompactionStrategy'}
+    AND caching = {'rows_per_partition': 10};
+
+ALTER TABLE repair_run
+    WITH compaction = {'class': 'LeveledCompactionStrategy'}
+    AND caching = {'rows_per_partition': 10};
+
+ALTER TABLE repair_run_by_cluster
+    WITH compaction = {'class': 'LeveledCompactionStrategy'}
+    AND caching = {'rows_per_partition': 10};
+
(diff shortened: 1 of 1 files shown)


adejanovski:
LCS makes sense since we're mostly reading. 
The way Reaper is implemented we would still have a lot of queries sent when retrieving all segments to get generate new tasks but we could be reading faster overall.
I thought TWCS would be a good fit here but since it's not available by default, it's not possible to add it straight away.

Cache is a smart idea too but unless activated in the cluster's config, we won't benefit from it (and it's turned off by default). 
OK with keeping it since it will do nothing in the worst case, and bring improvements in the best one.

adejanovski:
I have some concerns about the recompactions involved in altering the schema on existing clusters. Not that the reaper db should be very big anyway, but we could trigger a lot of compactions all at once in the worst case.

We could instead modify the existing CQL file and affect only new installs this way. Then users could do the change themselves if required.

What do you think ?

rustyrazorblade:
> I have some concerns about the recompactions involved in altering the schema on existing clusters. Not that the reaper db should be very big anyway, but we could trigger a lot of compactions all at once in the worst case.

I'm not concerned about this.  I feel like it's a premature optimization to try to avoid it.  We should have release notes that warn users about changes like this though.

michaelsembwever:
> I thought TWCS would be a good fit here but since it's not available by default, it's not possible to add it straight away.

But none of the tables are time-series? (not obviously so anyway. i'd be reluctant to introduce that constraint upon the datamodel when it's not obvious to begin with)

michaelsembwever:
> I have some concerns about the recompactions involved in altering the schema on existing clusters. Not that the reaper db should be very big anyway, but we could trigger a lot of compactions all at once in the worst case.

No concerns. The size of the data involved will be small.

adejanovski:
> But none of the tables are time-series? (not obviously so anyway. i'd be reluctant to introduce that constraint upon the datamodel when it's not obvious to begin with)

Right, actually it was on the tables I've added in the ft-reaper branch. Nothing is TTLed here so nevermind :)

PR #97, opened 2017-05-09
Don't handle success callbacks after timeout (C* <= 2.1)

Bj0rnen:
This is an important fix that we merged into Spotify's production Reaper in the beginning of the year (https://github.com/spotify/cassandra-reaper/pull/148). The "Old repair API" (C* 1.2, 2.0, maybe 2.1 too) inexplicably returns `SESSION_SUCCESS` when a repair session is aborted. That's what we do on timed out segments, and this led to a race condition where there's a chance that a segment that takes too long is flagged as "done" and never gets repaired.


adejanovski:
I never ran into this one so it may be handled a bit differently in 2.0 (or never had a "spotted" timeout in 2.1).

Thanks !

PR #98, opened 2017-05-09
Bump aborted runs down in cluster run summary

Bj0rnen:
Null fields put old aborted runs high in the sort order for repair
runs history of a cluster. This moves those down to an appropriate
position.

(Port of https://github.com/spotify/cassandra-reaper/pull/151)


adejanovski:
Thanks @Bj0rnen, I need to check if we have the same issue with the Cassandra backend.

PR #101, opened 2017-05-12
simplify creating repair run and segments

michaelsembwever:
Repair runs and their segments are one unit of work in concept and the persistence layer should be designed accordingly.

 Previous they were separated because the concern of sequence generation for IDs were exposed in the code. This is now encapsulated within storage implementations.

 This work allows the CassandraStorage to implement segments as clustering keys within the repair_run table.

ref:
 - https://github.com/thelastpickle/cassandra-reaper/issues/94

PR #102, opened 2017-05-13
In CassandraStorage implement segments as clustering keys within the repair_run table

michaelsembwever:
Collapses three tables into just one `repair_run`.

Also a change as required in IStorage so to identify a segment both by runId and segmentId.

ref:
 - https://github.com/thelastpickle/cassandra-reaper/issues/94
 - https://github.com/thelastpickle/cassandra-reaper/pull/102

### DIFF
--- a/src/test/resources/cassandra-reaper-cassandra-at.yaml
+++ b/src/test/resources/cassandra-reaper-cassandra-at.yaml
@@ -8,6 +8,7 @@ repairRunThreadCount: 15
 hangingRepairTimeoutMins: 1
 storageType: cassandra
 incrementalRepair: false
+allowUnreachableNodes: true
 
 logging:
   level: DEBUG
--- a/pom.xml
+++ b/pom.xml
@@ -11,7 +11,7 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <dropwizard.version>1.1.0</dropwizard.version>
+        <dropwizard.version>1.0.7</dropwizard.version>
         <dropwizard.cassandra.version>4.1.0</dropwizard.cassandra.version>
         <cassandra.version>2.2.7</cassandra.version>
         <cucumber.version>1.2.5</cucumber.version>
--- a/src/main/java/com/spotify/reaper/ReaperApplicationConfiguration.java
+++ b/src/main/java/com/spotify/reaper/ReaperApplicationConfiguration.java
@@ -254,7 +254,7 @@ public class ReaperApplicationConfiguration extends Configuration {
   }
   
   public Boolean getAllowUnreachableNodes() {
-    return allowUnreachableNodes;
+    return allowUnreachableNodes != null ? allowUnreachableNodes : false;
   }
 
   public void setAllowUnreachableNodes(Boolean allow) {
--- a/src/main/java/com/spotify/reaper/service/RepairRunner.java
+++ b/src/main/java/com/spotify/reaper/service/RepairRunner.java
@@ -229,7 +229,7 @@ public class RepairRunner implements Runnable {
 
         // Just checking that no currently running segment runner is stuck.
         RepairSegment supposedlyRunningSegment =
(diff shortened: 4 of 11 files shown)


michaelsembwever:
This branch is looking better now.
The only thing is that for a yet unknown reason the integration tests stopped working on me unless I switch them to use `allowUnreachableNodes`, ref https://github.com/thelastpickle/cassandra-reaper/pull/102/files#diff-fa47d3fc97cdc9d63f61af8cd48c2f5bR11

src/test/resources/cassandra-reaper-cassandra-at.yaml
```diff
@@ -8,6 +8,7 @@ repairRunThreadCount: 15
 hangingRepairTimeoutMins: 1
 storageType: cassandra
 incrementalRepair: false
+allowUnreachableNodes: true
```
michaelsembwever, line 11: Integration tests wouldn't work without turning this on.  Still not sure if i broke something, if my environment changed, or if I fixed something…?

src/main/resources/db/cassandra/001_Initialize_db.cql
```diff
@@ -23,24 +23,32 @@ CREATE TABLE IF NOT EXISTS repair_unit (
...
+  end_token          varint,
+  segment_state      int,
+  coordinator_host   text,
+  segment_start_time timestamp,
+  segment_end_time   timestamp,
+  fail_count         int,
+  PRIMARY KEY (id, segment_id)
 )
   WITH compaction = {'class': 'LeveledCompactionStrategy'}
   AND caching = {'rows_per_partition': 10};
```
adejanovski: I still think we need to make no change that requires the users to drop the database then recreate it.
As schema migration is performed, we need to create a new cql file with a new version number, create a single table in it that we would call `repair_run_v2` for example, and move all queries from `repair_run` to `repair_run_v2` in the code. Otherwise the app will just crash if people with existing installs try to upgrade.

We also need more caching here since many runs could have thousands of segments. After running some tests on the old schema, setting 1000 rows_per_partition uses only a few kb in cache so I guess we can use values ranging between 1000 and 5000 to maximize cache utilization.

src/test/resources/cassandra-reaper-cassandra-at.yaml
adejanovski: I doubt you fixed something here :)
allowUnreachableNodes permits have reaper start processing a segment even if all replicas aren't reachable through JMX. This was introduced to repair multi region clusters with the JMX port closed in between regions (which should be the case for security sake).

This flag will be deprecated with the introduction of the ft-reaper branch that uses the Cassandra backend to exchange metrics between regions.

I'll try to replicate the flaky tests using this branch and see if I can make sense out of it.

adejanovski (changes requested):

src/test/resources/cassandra-reaper-cassandra-at.yaml
michaelsembwever: a) Previously the integration tests were working for me even though two of the three nodes could not be reached via JMX.

b) With this patch the integration tests then started to fail.

c) The integration tests with this patch do work with `allowUnreachableNodes=true` ,  as one would expect.

I don't know, but I'm left wondering if the problem was in (a), and not actually in this patch. It's worth checking though.

src/main/resources/db/cassandra/001_Initialize_db.cql
michaelsembwever: will fix…

adejanovski (approved):

michaelsembwever:
This PR will get re-merged again, into 'mck/cassandra-improvements-94', after #101 gets merged.

michaelsembwever:
> I doubt you fixed something here :)
> `allowUnreachableNodes` permits have reaper start processing a segment even if all replicas aren't reachable through JMX. This was introduced to repair multi region clusters with the JMX port closed in between regions (which should be the case for security sake).

I'll take it out. As I figured out I was just having problems with local iptables occasionally blocking the jmx connections to the ccm addresses.

PR #107, opened 2017-05-24
Fix CFs list broken display in repairs and schedule views.

anguenot:
See before and after screenshots for repairs and schedule pages:

before_repair_page
after_repair_page
before_schedule_page
after_schedule_page


adejanovski:
That's much better indeed :)

Thanks @anguenot !

PR #110, opened 2017-06-02
move into the next development cycle

michaelsembwever:
It's healthy to only have non-snapshot versions existing in one git SHA.
That is master should be bumped immediately into SNAPSHOT, after each release, to ensure releases are attached to unique SHA and of the one distribution.

For example, now in master a `mvn install` would overwrite the 0.6.0 release on your local machine.


adejanovski (commented):
Agreed, I forgot to bump the pom to a new snapshot version

PR #114, opened 2017-06-08
LongCollectionSQLType should hold Longs, not UUIDs

vrischmann:
Fixes #113

Issue #113, opened 2017-06-07
PostgreSQL exception since the ID changed to UUID


vrischmann:
Hi,

recently I updated reaper and noticed the schedule and repair IDs changed to a timeuuid. I'm running a build from master right now.

There is an exception when using the PostgreSQL storage:

```
ERROR  [2017-06-07 12:35:36,627] [SchedulingManagerTimer] c.s.r.s.SchedulingManager - failed managing schedule for run with id: 00000000-0000-000a-b8dc-6f0948892c1e
ERROR  [2017-06-07 12:35:36,628] [SchedulingManagerTimer] c.s.r.s.SchedulingManager - catch exception
org.skife.jdbi.v2.exceptions.UnableToExecuteStatementException: org.postgresql.util.PSQLException: ERROR: invalid input syntax for integer: "00000000-0000-000c-b8dc-6f0948892c1e" [statement:"UPDATE repair_schedule SET repair_unit_id = :re
pairUnitId, state = :state, days_between = :daysBetween, next_activation = :nextActivation, run_history = :runHistorySQL, segment_count = :segmentCount, repair_parallelism = :repairParallelism, creation_time = :creationTime, owner = :owne
r, pause_time = :pauseTime WHERE id = :id", located:"UPDATE repair_schedule SET repair_unit_id = :repairUnitId, state = :state, days_between = :daysBetween, next_activation = :nextActivation, run_history = :runHistorySQL, segment_count =
:segmentCount, repair_parallelism = :repairParallelism, creation_time = :creationTime, owner = :owner, pause_time = :pauseTime WHERE id = :id", rewritten:"/* IStoragePostgreSQL.updateRepairSchedule */ UPDATE repair_schedule SET repair_uni
t_id = ?, state = ?, days_between = ?, next_activation = ?, run_history = ?, segment_count = ?, repair_parallelism = ?, creation_time = ?, owner = ?, pause_time = ? WHERE id = ?", arguments:{ positional:{}, named:{owner:'root',followingAc
tivation:io.dropwizard.jdbi.args.JodaDateTimeArgument@4512ac38,creationTime:io.dropwizard.jdbi.args.JodaDateTimeArgument@532bfbc0,nextActivation:io.dropwizard.jdbi.args.JodaDateTimeArgument@1748b980,daysBetween:9,segmentCount:4000,runHist
[...]

PR #116, opened 2017-06-11
Adds integration test to ensure scheduled repairs can initiate repair runs, and maintain history

michaelsembwever:
Adds integration test to ensure scheduled repairs can initiate repair runs, and maintain history

 This test fails before commit https://github.com/thelastpickle/cassandra-reaper/commit/3d16533fe5be00cd1ce500225349068b33b974bf, and passes with and after it.

 Part of the requirement is that, with assertions enabled, as they are when running tests, the SchedulingManager exits the JVM on errors.
 This is ok, as fail-fast behaviour would be expected when assertions are enabled.

ref:
 - https://github.com/thelastpickle/cassandra-reaper/issues/113 – PostgreSQL exception since the ID changed to UUID
 - https://github.com/thelastpickle/cassandra-reaper/pull/114 – LongCollectionSQLType should hold Longs, not UUIDs

### DIFF
--- a/src/main/java/com/spotify/reaper/service/SchedulingManager.java
+++ b/src/main/java/com/spotify/reaper/service/SchedulingManager.java
@@ -43,7 +43,7 @@ public class SchedulingManager extends TimerTask {
         .pauseTime(DateTime.now())
         .build(schedule.getId());
     if (!context.storage.updateRepairSchedule(updatedSchedule)) {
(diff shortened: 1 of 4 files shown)


michaelsembwever:
some failures in `ReaperH2IT` i need to look into…

michaelsembwever:
> some failures in ReaperH2IT i need to look into…

fixed. (test was correct and uncovered a valid new bug)

adejanovski:
LGTM, merging.

PR #120, opened 2017-06-19
Changed the git protocol form SSH to http of startbootstrap-sb-admin

jagodevreede:
To use git@github.com: you need to have SSH set up, so only BlackrockDigital would have access. With this PR everyone can build the UI


adejanovski:
Works for me.

Thanks !

PR #121, opened 2017-06-19
Allow the auto scheduler to honour the incrementalRepair option.

jagodevreede:
Currently the auto-scheduler will always schedule a full repair, regardless of the incrementalRepair option. This PR will honour that setting, allowing to auto-schedule incremental repairs.


adejanovski:
Good catch !

Thanks

PR #130, opened 2017-07-03
storage type as capability definition

michaelsembwever:
Implement "local mode" as a "Distributed Storage" capability.


adejanovski (approved):
LGTM, modulus the failed tests which will be handled in the distributed-testing branch I guess (problems aren't coming from this very branch anyway).

michaelsembwever:
> modulus the failed tests which will be handled in the distributed-testing branch I guess (problems aren't coming from this very branch anyway).

that's correct.

PR #131, opened 2017-07-03
Make UUIDs returned from PostgresStorage deterministic to reduce confusion

Bj0rnen:
I'm not sure if there was a particular reason that that least significant 64 bits were taken from a time-based UUID. It led to surprising behavior, like all of the UUIDs (returned from various endpoints) changing every time Reaper got restarted.

Perhaps `toSequenceId` should also require the least significant 64 bits to be 0, so that only the reported UUID can be used for commands like spreaper's pause-schedule.

### DIFF
--- a/src/main/java/com/spotify/reaper/storage/PostgresStorage.java
+++ b/src/main/java/com/spotify/reaper/storage/PostgresStorage.java
@@ -25,15 +25,7 @@ import com.spotify.reaper.resources.view.RepairRunStatus;
 import com.spotify.reaper.resources.view.RepairScheduleStatus;
 import com.spotify.reaper.service.RepairParameters;
 import com.spotify.reaper.service.RingRange;
-import com.spotify.reaper.storage.postgresql.BigIntegerArgumentFactory;
-import com.spotify.reaper.storage.postgresql.IStoragePostgreSQL;
-import com.spotify.reaper.storage.postgresql.LongCollectionSQLTypeArgumentFactory;
-import com.spotify.reaper.storage.postgresql.PostgresArrayArgumentFactory;
-import com.spotify.reaper.storage.postgresql.RepairParallelismArgumentFactory;
-import com.spotify.reaper.storage.postgresql.RunStateArgumentFactory;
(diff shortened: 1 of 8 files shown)


Bj0rnen:
Didn't actually fix it, because there are multiple other places that do the same thing. I will factor them out.

michaelsembwever:
Thanks for seeing and fixing this @Bj0rnen 
Went straight past me. Of course, over multiple jvm processes we need UUIDs to be reproducible from timestamp (within the storage implementation)…

michaelsembwever:
@Bj0rnen is this ready for review now?

src/main/java/com/spotify/reaper/storage/PostgresStorage.java
```diff
@@ -496,11 +495,11 @@ public boolean updateRepairSchedule(RepairSchedule newRepairSchedule) {
     }
   }
 
-  private static UUID fromSequenceId(long insertedId) {
-    return new UUID(insertedId, UUIDs.timeBased().getLeastSignificantBits());
+  public static UUID fromSequenceId(long insertedId) {
```
michaelsembwever: The problem with re-using these static methods is it creates a cyclic dependency between the `com.spotify.reaper.storage` and `com.spotify.reaper.storage.postgresql` packages.

If the two methods are moved something into the `com.spotify.reaper.storage.postgresql` package then the dependency direction remains only from `com.spotify.reaper.storage` to `com.spotify.reaper.storage.postgresql` as it is today.

michaelsembwever (changes requested):

src/main/java/com/spotify/reaper/storage/PostgresStorage.java
Bj0rnen: Good point! I'll split those to a separate class.

michaelsembwever (approved):

michaelsembwever:
Pushed branch to https://github.com/thelastpickle/cassandra-reaper/tree/DeterministicPostgresUUID
 and made a few small edits.

From there it was merged to master.

PR #46, opened 2017-02-01
Attempt building debian, jar, and rpm packages using Dockerfiles

joaquincasares:
Pushing my Dockerfiles for review.

Experiencing these issues when installing the rpm package:

```
xxx -  -  [01/Feb/2017:01:17:51 +0000] "GET /webui HTTP/1.1" 200 - "-" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_2) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/55.0.2883.95 Safari/537.36" 10
xxx -  -  [01/Feb/2017:01:17:51 +0000] "GET /deps.js HTTP/1.1" 404 - "http://52.33.90.49:8080/webui" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_2) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/55.0.2883.95 Safari/537.36" 1
xxx -  -  [01/Feb/2017:01:17:51 +0000] "GET /index.js HTTP/1.1" 404 - "http://52.33.90.49:8080/webui" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_2) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/55.0.2883.95 Safari/537.36" 1
xxx -  -  [01/Feb/2017:01:17:55 +0000] "GET /index.html HTTP/1.1" 404 - "http://52.33.90.49:8080/webui" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_2) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/55.0.2883.95 Safari/537.36" 0
```

Experiencing this issue when attempting to build the jar while inside a Docker container:

```
[INFO] --- exec-maven-plugin:1.5.0:exec (exec-bower-install) @ cassandra-reaper ---
bower ESUDO         Cannot be run with sudo

Additional error details:
Since bower is a user command, there is no need to execute it with superuser permissions.
If you're having permission errors when using bower without sudo, please spend a few minutes learning more about how your system should work and make any necessary repairs.

http://www.joyent.com/blog/installing-node-and-npm
https://gist.github.com/isaacs/579814

You can however run a command with sudo using --allow-root option
```

The debian packages build successfully, even if they take a while to build.

PR #135, opened 2017-07-07
Add jmxAuth keys to default configs.

joaquincasares:
While the majority of these changes merely expose the jmxAuth variable, I'm concerned about the `src/main/docker/Dockerfile` change.

I'm opening this PR as I continue to deprecate #46.


joaquincasares:
This has been tested by #136 to work as expected.

joaquincasares:
@ossarga /cc @adejanovski I'll be merging this in since #136 will rely on this change and the Docker-related files have been tested via #136.

PR #134, opened 2017-07-07
Add docker-build process.

joaquincasares:
This is the first pull request in an attempt to resolve #46, which has partially, and more accurately, been taken care of through #76.

The bulk of these changes appear in the README.

The intent of these files is to build the RPM and Debian packages without any system dependencies other than Docker.

PR #136, opened 2017-07-08
Docker compose environment

joaquincasares:
This PR is dependent on #135.

This is the last PR in an effort to to deprecate #46.

This adds a sample Cassandra node, including all the required settings and Docker Compose services in order to run Reaper in a clean and isolated environment with the only dependencies being Docker and Docker Compose.

I've also included a semi-lengthy, in-depth view of Docker Compose infrastructure within `./docker/README.md`. At the end of this README, are instructions for how to run Dockerized Reaper within production environments.

### DIFF
--- a/docker/initialize-reaper_db/Dockerfile
+++ b/docker/initialize-reaper_db/Dockerfile
@@ -1,6 +1,8 @@
 FROM cassandra:3.11
 
-# create the required reaper_db keyspace to allow Reaper to store scheduling data
-CMD set -x \
-    && cqlsh cassandra -e \
-    "CREATE KEYSPACE IF NOT EXISTS reaper_db WITH replication = {'class': 'SimpleStrategy', 'replication_factor': 1 };"
+# create the reaper_db keyspace
+COPY docker-entrypoint.sh /
+ENTRYPOINT ["/docker-entrypoint.sh"]
+
+# use a default replication factor of 1
+CMD ["1"]
--- a/docker/initialize-reaper_db/Dockerfile
+++ b/docker/initialize-reaper_db/Dockerfile
@@ -1,0 +1,6 @@
+FROM cassandra:3.11
+
+# create the required reaper_db keyspace to allow Reaper to store scheduling data
+CMD set -x \
+    && cqlsh cassandra -e \
+    "CREATE KEYSPACE IF NOT EXISTS reaper_db WITH replication = {'class': 'SimpleStrategy', 'replication_factor': 1 };"
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,6 @@ packages/
 # frontend dev stuffs.
 reaper_ui/bower_components/
 reaper_ui/node_modules/
+
+# Docker Compose persistent storage directory
+data/
--- a/README.md
+++ b/README.md
@@ -431,9 +431,13 @@ node's status:
 
 Once the Cassandra node is online and accepting CQL connections,
 create the required `reaper_db` Cassandra keyspace to allow Reaper to save
-its cluster and scheduling data:
+its cluster and scheduling data.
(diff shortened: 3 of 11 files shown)


docker/initialize-reaper_db/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM cassandra:3.11
+
+# create the required reaper_db keyspace to allow Reaper to store scheduling data
+CMD set -x \
+    && cqlsh cassandra -e \
+    "CREATE KEYSPACE IF NOT EXISTS reaper_db WITH replication = {'class': 'SimpleStrategy', 'replication_factor': 1 };"
```
rustyrazorblade: I don't think we should be assuming RF=1 anywhere.

michaelsembwever:
@joaquincasares you can't merge master into branches.
ref 01c939ae00e789b088f7e44c4fff3a7806b95815

Branches (that are intended for master) must be rebased off master.

michaelsembwever:
I don't think it's related to this PR, but 
`make all` fails for me (linux) with 
```
fpm -s dir -t deb -n reaper -v `mvn org.apache.maven.plugins:maven-help-plugin:2.1.1:evaluate -Dexpression=project.version | grep -v '\['` --pre-install debian/preinstall.sh -C build .
```

ossarga:
Fails for me on (Mac OS X). You are right @michaelsembwever it is unrelated because I see the same problem on the *master* branch.

michaelsembwever:
The docker commands now work for me! (a couple of local pax issues i had to overcome)

docker/initialize-reaper_db/Dockerfile
michaelsembwever: this RF (i presume) it only for the one-node-cassandra within the docker container.

Maybe it's the applicaton's responsibility to log a warning at startup if RF=1 and N>1
wdyt?

joaquincasares: +1 on having an application logging if the RF is set lower than expected.

joaquincasares:
## `reaper_db` Replication Factor

@rustyrazorblade @michaelsembwever `reaper_db` now defaults to an RF of 1, but providing an additional argument will overwrite the default of 1, for 3, for example:

```bash
docker-compose run initialize-reaper_db 3
```

## Rebased

@michaelsembwever I've rebased and force pushed into the same branch.

## Makefile

@michaelsembwever @ossarga I did change the Make command on #134 to move from `$()` to using backticks (`) in order to get the build command working within an Ubuntu Docker container. I know it definitely works within Docker, but not sure how to support both, or to instead just name Docker as the only prerequisite for the build process.

I'd vote for the latter to simplify all documentation around the Make command and ensure the process works across all flavors of Linux and versions of OS X, without the need for many dependencies as the dependency list is currently quite high:

* https://github.com/thelastpickle/cassandra-reaper/blob/master/docker-build/Dockerfile#L8-L25

While I do notice that `nodejs` is only needed for building the UI, and that `git` and `ruby-dev` are downstream dependencies for also building the UI, if we were to not go along the Docker-first path for building our packages, we'll have to update our documentation for the following processes:

* How to build on OS X
* How to build on Linux
* How to build with Docker
* How to build the UI in OS X
* How to build the UI in Linux
* How to build the UI in Docker

The current documenation is essentially:

* Install Docker and Docker Compose.
* Run the following command: 
    * https://github.com/thelastpickle/cassandra-reaper#docker

## Local pax Issues

@michaelsembwever could you elaborate please?

michaelsembwever:
> @michaelsembwever could you elaborate please?

i've solved this, it was a local issue around PaX controls.

michaelsembwever (approved):

PR #140, opened 2017-07-19
Fix SSL encrypted and client authenticating connections to Cassandra's jmx

michaelsembwever:
Previously if the Cassandra node's had `com.sun.management.jmxremote.registry.ssl=true` then reaper would not be able to connect. This was because the rmi port becomes encrypted but the
jmx client in reaper was not configured to use ssl here.

This fix permits jmx rmi encrypted connections when the jvm option `-Dssl.enable=true` is provided.

### DIFF
--- a/src/main/java/com/spotify/reaper/cassandra/JmxProxy.java
+++ b/src/main/java/com/spotify/reaper/cassandra/JmxProxy.java
@@ -20,6 +20,8 @@ import java.lang.reflect.UndeclaredThrowableException;
 import java.math.BigInteger;
 import java.net.InetSocketAddress;
 import java.net.MalformedURLException;
+import java.rmi.server.RMIClientSocketFactory;
+import java.rmi.server.RMISocketFactory;
 import java.util.AbstractMap;
 import java.util.Collection;
 import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


ossarga:
Nice work. This looks similar to the way it is implemented in `nodetool`. Do we need a unit test or integration test for this?

michaelsembwever:
> Do we need a unit test or integration test for this?

That would be nice. Any ideas?

zznate:
> That would be nice. Any ideas?

Meh? We would be largely testing functionality of the JDK. Maybe just ensuring we return the right instance if `ssl.enable` is true.

michaelsembwever:
Let's put it in the too hard basket for now. It would be a full integration test run against a newly ssl configured ccm cluster. That's a lot of overhead.

michaelsembwever:
any objections to merging?

zznate:
Nope.

ossarga:
No. These changes look fine. As mentioned, this is pretty much what `nodetool` does.

ossarga (approved):
@michaelsembwever these changes are really good. You and @zznate have good points about the tests. As mentioned, this is pretty much what `nodetool` does. So lets proceed with the merge.

PR #141, opened 2017-07-19
Add instructions to install and run reaper as a service

adejanovski:
Remove unit tests from Makefile package phase (already done by the docker container)
Change package path to move from spotify to cassandra-reaper

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,15 +5,18 @@ VERSION := `mvn org.apache.maven.plugins:maven-help-plugin:2.1.1:evaluate -Dexpr
 # VERSION := `python -c "import xml.etree.ElementTree as ET; print(ET.parse(open('pom.xml')).getroot().find('{http://maven.apache.org/POM/4.0.0}version').text)"`
 
 package:
-	mvn package
+	mvn package -DskipTests
 
 prepare:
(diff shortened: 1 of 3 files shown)


joaquincasares:
@adejanovski just a heads up that the `docker-build` image doesn't run tests, but strictly uses the Makefile.

If we removed the tests from the Makefile, we'll need to add them to the `docker-build`'s entrypoint:

https://github.com/thelastpickle/cassandra-reaper/blob/master/docker-build/docker-entrypoint.sh#L10-L13

I would personally opt for the tests in the Makefile, perhaps as a target, then add that target to the entrypoint.

adejanovski:
@joaquincasares : when you call `mvn clean package -Pbuild-ui` in the docker-entrypoint.sh file, you run the unit tests. Then, the makefile calls `mvn package` which runs both the build and the unit tests for the second time.
Since the docker container builds the UI, which requires some dependencies that everyone might not have, it was fair to keep the build phase there.
I could've removed the tests from that build instead of the makefile's one, but it's no big deal anyway.

joaquincasares:
Ah, thanks for explaining. As long as the docker-build path still runs tests, all seems great! :)

PR #145, opened 2017-07-28
Add ability to have both a schedule for full and incremental repairs …

### LINKED ISSUE #144: Schedule 2 repairs for the same cluster and keyspace with different incrementalRepair value
Hi all!

We are trying to schedule two repairs for a keyspace, one with incrementalRepair=true and other with incrementalRepair=false, to follow the recommended ways explained [here](http://docs.datastax.com/en/cassandra/3.0/cassandra/operations/opsRepairNodesWhen.html), but it fails with this error message: 
`A repair schedule already exists for cluster "testcluster", keyspace "testkeyspace", and column families: []`

Inside code, this message is located [here](https://github.com/thelastpickle/cassandra-reaper/blob/master/src/main/java/com/spotify/reaper/resources/CommonTools.java#L249). Could be possible to change this behaviour to include incrementalRepair property in comparison?

adejanovski:
…at the same time for the same keyspace.

This PR fixes #144


sergiogarcia94:
Hey @adejanovski we built this branch yesterday and it works fine for us, we are able to create 2 repairs at the same time now. Thanks for the fast job :D. 
Btw, we found a typo on GUI, the "Parallism" on "Schedules/Repair" form, dont know if this is the right place to fix that . Anyway, as i said good job.

PR #146, opened 2017-07-28
Fix coordinator node being nulled out by abort() for incremental repairs

adejanovski:
`SegmentRunner.abort()` was nulling the coordinator node as it didn't provide the repair unit to `postpone()`.
This is a problem for incremental repair as the coordinator node is never recomputed after repair run creation.

The fix provides the RepairUnit which allows postpone() to detect incremental repairs and avoid nulling `coordinator_host`.


michaelsembwever:
A unit test to prevent this from regressing, please.

Issue #151, opened 2017-08-01
ClassCastException in RepairParametersMapper when using H2 database


smarsching:
There is a bug in the RepairParametersMapper when using the H2 database (instead of PostgreSQL). The code tries to cast Object[] to String[], which results in a ClassCastException:

> java.lang.ClassCastException: [Ljava.lang.Object; cannot be cast to [Ljava.lang.String;
>         at com.spotify.reaper.storage.postgresql.RepairParametersMapper.map(RepairParametersMapper.java:33)
>         at com.spotify.reaper.storage.postgresql.RepairParametersMapper.map(RepairParametersMapper.java:28)
>         at org.skife.jdbi.v2.Query$4.munge(Query.java:183)
>         at org.skife.jdbi.v2.QueryResultSetMunger.munge(QueryResultSetMunger.java:41)
>         at org.skife.jdbi.v2.SQLStatement.internalExecute(SQLStatement.java:1344)
>         at org.skife.jdbi.v2.Query.fold(Query.java:173)
>         at org.skife.jdbi.v2.Query.list(Query.java:82)
>         at org.skife.jdbi.v2.sqlobject.ResultReturnThing$IterableReturningThing.result(ResultReturnThing.java:253)
>         at org.skife.jdbi.v2.sqlobject.ResultReturnThing.map(ResultReturnThing.java:43)
>         at org.skife.jdbi.v2.sqlobject.QueryHandler.invoke(QueryHandler.java:43)
>         at org.skife.jdbi.v2.sqlobject.SqlObject.invoke(SqlObject.java:212)
>         at org.skife.jdbi.v2.sqlobject.SqlObject$3.intercept(SqlObject.java:127)
>         at org.skife.jdbi.v2.sqlobject.CloseInternalDoNotUseThisClass$$EnhancerByCGLIB$$3580ecc6.getRunningRepairsForCluster(<generated>)
>         at com.spotify.reaper.storage.PostgresStorage.getOngoingRepairsInCluster(PostgresStorage.java:356)
>         at com.spotify.reaper.service.SegmentRunner$1.initialize(SegmentRunner.java:175)
>         at com.spotify.reaper.service.SegmentRunner$1.initialize(SegmentRunner.java:171)
>         at org.apache.commons.lang3.concurrent.LazyInitializer.get(LazyInitializer.java:100)
>         at com.spotify.reaper.service.SegmentRunner.canRepair(SegmentRunner.java:302)
[...]

PR #152, opened 2017-08-01
Fix ClassCastException when using H2 database (#151).

smarsching:
This PR fixes a ClassCastException in RepairParametersMapper when using the H2 database instead of PostgreSQL.


michaelsembwever:
thanks @smarsching. looks good. i'll take a closer look and merge this later today.

PR #158, opened 2017-08-18
Add SSL encrypted connections to Docker containers

ossarga:
- Added a common-services.yml file and reworked docker-compose.yml file so that the services are extensions of the services in common-services.yml file
- Added Generate SSL Stores service to docker-compose to generate signed certs and SSL stores
- Added Cassandra SSL service to docker-compose which only accepts encrypted JMX and Native Protocol connections
- Added Initialize Reaper DB SSL service to docker-compose to initialise keyspace in Cassandra SSL container via an encrypted connection
- Added nodetool SSL service to docker-compose to query Cassandra SSL container via an encrypted connection
- Added CQLSH SSL encrypted service to docker-compose
- Replaced intialize-reaper_db and initialize-reaper_db-ssl services to use CQLSH common service
- Added an environment variable to the Reaper Docker container to allow Reaper to use an SSL encrypted connection when connecting with Cassandra storage
- Minor updates to Docker Reaper configuration YAML so that it contains latest Reaper options
- Fixed bug where JVM options were passed to Reaper in the wrong position
- Updated site documentation to include instructions on orchestrating an SSL encrypted environment

PR #181, opened 2017-08-31
Docker Compose updates

ossarga:
- Updated top level docker-compose.yml to allow package building to be called at the _./packaging_ directory level using the commands:
   `docker-compose build reaper-build-packages && docker-compose run reaper-build-packages`

- Renamed ._/packaging/docker_ to _./packaging/docker-services_ to better reflect the directory contents and to separate it from the ./packaging/docker-build directory

- Updated documentation


joaquincasares:
Aside from what's mentioned in #158 , this looks good.

I do like the use of `docker-services` when used with `docker-build`. Good naming. :)

michaelsembwever:
in the commit messages, can the first line be more active and descriptive. (what changed is superfluous (the diff shows that already).)

the second commit message would be fine just without the first line, and the second line is a good topic sentence, and provides enough. but it probably should get merged into the first commit, so you have to find a good topic sentence that sums up why the list of points was needed.

other than that the PR looks good.

michaelsembwever (approved):

ossarga:
Thanks @joaquincasares and @michaelsembwever for the feedback. Changes made as requested. Merging this branch.

Issue #185, opened 2017-09-04
Hung repair segments are sometimes not aborted


smarsching:
Sometimes it happens that a hung repair segment is not aborted, thus keeping other repair segments (possible from other repair runs) from running.

In the logfile the following message shows up:

> SegmentRunner declined to repair segment … because one of the hosts (…) was already involved in a repair

It took me some time to figure out why this happens, but I think I understand the issue now:

Usually, Cassandra Reaper will use the timeout to determine when a repair segment is stuck and abort the segment when it takes too long. Obviously, the segment will not be aborted if Cassandra Reaper is restarted before the timeout. Usually, this is not a problem because on startup, Cassandra Reaper iterates through all running segments of all running jobs and restart them. When the job has been paused (or aborted), this will not happen (which makes sense because the segment should not be restarted in this case).

The issue is that the stale entry for the "running" segment will still keep repairs on the affected nodes from being aborted, possibly blocking future repair runs. I believe that this is due to a bug in the implementations of the`getOngoingRepairsInCluster(String)` method in `CassandraStorage` and `PostgresStorage`. The implementation in `MemoryStorage` does not have this bug, so the problem described above should not appear when using `MemoryStorage`.

I will try to put together a PR which fixes this issue.

adejanovski:
Fixed by bdb0bd1eba1c5c06c512461fb17829daaa1872bd

Issue #182, opened 2017-08-31
discuss adding java formatter


rustyrazorblade:
It would be nice to run our code through a formatter with a set of conventions we can agree on and not have to nit over style.  Something like Google's Java formatter: https://github.com/google/google-java-format, but it doesn't have to be exactly their conventions / decisions.  Style should be, in my opinion, standardized and removed from the discussion when looking at a patch.  It adds noise to the discussion and focuses on the appearance of the code rather than the quality.

michaelsembwever:
I don't think it should matter which tool we use, there's a few out there, but that the efforts of any commit are measurable.
Re-applying a formatter isn't a sustainable approach: once the code is cleaned up then a PR can simply be blocked until code style meets requirements.

Measuring code style can be done in a number of different ways: checkstyle, pmd, findbugs, sonar; which all plugin into maven. I believe @adejanovski already has set up sonar reporting for Reaper? And i believe checkstyle, pmd, and findbugs all can plug into sonar?

rustyrazorblade:
I'm not pushing for any particular tool, just that something is picked and we roll with it.  I was using the google one as an example.

michaelsembwever:
> I'm not pushing for any particular tool, just that something is picked and we roll with it. I was using the google one as an example.

Think we might be speaking past each other there? I don't think we need to pick any tool because any tool that is used should only be used the once. What i would say is more important is that we have code style rules and metrics in place. 

(For example with these things in place, after we're done one pass of code formatting, then there's no excuse that the code style deteriorates again… We also  want to avoid a code base that's littered with ongoing automated formatting commits)

rustyrazorblade:
> I don't think we need to pick any tool because any tool that is used should only be used the once. 

I disagree.  Any other project that I've worked on has required the use of the code formatter to ensure the project stays formatted properly.  Otherwise you end up with a bunch of nit comments over formatting, and that's just unnecessary noise. 

> We also want to avoid a code base that's littered with ongoing automated formatting commits

I don't see formatting commits as a negative thing, so I'm going to disagree with you on this.  Ideally any PR that goes up has been run through the formatter already.  People can squash those commits, but I really don't care if they do or don't.  Perfectly clean history isn't something I attach value to.

michaelsembwever:
Work in progress in https://github.com/thelastpickle/cassandra-reaper/pull/198

PR #188, opened 2017-09-05
Only consider segments belonging to running jobs.

smarsching:
This PR fixed issue #185 by making `CassandraStorage` and `PostgresStorage` behave like `MemoryStorage` in that only running segments belonging to running jobs will keep Cassandra Reaper from terminating the repair process on a node.

### DIFF
--- a/src/server/src/main/java/com/spotify/reaper/storage/CassandraStorage.java
+++ b/src/server/src/main/java/com/spotify/reaper/storage/CassandraStorage.java
@@ -539,6 +539,9 @@ public final class CassandraStorage implements IStorage, IDistributedStorage {
     Collection<RepairRun> repairRuns = getRepairRunsForCluster(clusterName);
 
     for(RepairRun repairRun:repairRuns){
+      if (repairRun.getRunState() != RunState.RUNNING) {
+        continue;
+      }
       Collection<RepairSegment> runningSegments = getSegmentsWithState(repairRun.getId(), State.RUNNING);
       for(RepairSegment segment:runningSegments){
         Optional<RepairUnit> repairUnit = getRepairUnit(repairRun.getRepairUnitId());
--- a/src/server/src/main/java/com/spotify/reaper/storage/postgresql/IStoragePostgreSQL.java
+++ b/src/server/src/main/java/com/spotify/reaper/storage/postgresql/IStoragePostgreSQL.java
@@ -130,7 +130,7 @@ public interface IStoragePostgreSQL {
           + "FROM repair_segment "
           + "JOIN repair_run ON run_id = repair_run.id "
           + "JOIN repair_unit ON repair_run.repair_unit_id = repair_unit.id "
-          + "WHERE repair_segment.state = 1 AND repair_unit.cluster_name = :clusterName";
+          + "WHERE repair_segment.state = 1 AND repair_run.state = 'RUNNING' AND repair_unit.cluster_name = :clusterName";
   String SQL_GET_NEXT_FREE_REPAIR_SEGMENT =
       "SELECT " + SQL_REPAIR_SEGMENT_ALL_FIELDS + " FROM repair_segment WHERE run_id = :runId "
       + "AND state = 0 ORDER BY fail_count ASC, start_token ASC LIMIT 1";


adejanovski:
Hi @smarsching,

thanks for that PR! I'll review it ASAP.

src/server/src/main/java/com/spotify/reaper/storage/CassandraStorage.java
```diff
@@ -539,6 +539,9 @@ private static RepairSegment createRepairSegmentFromRow(Row segmentRow){
     Collection<RepairRun> repairRuns = getRepairRunsForCluster(clusterName);
 
     for(RepairRun repairRun:repairRuns){
+      if(repairRun.getRunState() != RunState.RUNNING){
```
michaelsembwever: nit-pick: spaces, so like `if (repairRun.getRunState() != RunState.RUNNING) {`

(this class isn't a good example, so it's really really trivial)

michaelsembwever: wonder if making lines 539-550 into 
```
    getRepairRunsForCluster(clusterName)
            .stream()
            .filter(r -> r.getRunState() == RunState.RUNNING)
            .forEach(...);
```
would be more elegant?

smarsching: I changed the spacing. Originally, I simply made it look like the rest of the file. Is there a style-guide for Cassandra Reaper?

Regarding the functional approach: I like it, but all of the rest of the code is written in a imperative way. Therefore I think translating everything into a more functional style should be a separate change, that is not related to this bugfix.

rustyrazorblade: @smarsching we've got an issue open to discuss using a formatting tool, #182.

adejanovski:
Hi @smarsching,

Indeed, pausing a repair won't stop the running segments. I see what you're trying to do here, but I'm concerned some implications.
So the problematic case is : 
- Reaper instance A pauses a repair run but a segment is running
- that segment won't be stopped until (poorly-named) `resumeRunningRepairRuns` kicks in (every 30s) and [aborts the segment](https://github.com/thelastpickle/cassandra-reaper/blob/master/src/server/src/main/java/com/spotify/reaper/service/RepairManager.java#L75-L83)
- If Reaper instance A is restarted before the segment is aborted, then upon restart, it will have no more RepairRunner for that run and `resumeRunningRepairRuns` won't try to abort the segment
- As the segment is still being repaired by Cassandra nodes (and could possibly hang) and no Reaper instance is responsible for it, no other repair will be able to start on the same nodes until it's actually over (but it's status won't be updated anymore and it will stay at 'RUNNING')

The thing is that it looks as your fix will allow to kill the repair session on the nodes but [won't update the status of the segment in the database](https://github.com/thelastpickle/cassandra-reaper/blob/master/src/server/src/main/java/com/spotify/reaper/service/SegmentRunner.java#L453-L465), and furthermore, if there's a slight inconsistency in the repair run status, it won't try to get leadership on the segment although another instance of Reaper could very well be in charge of (remember that Reaper is distributed now and several instance can be processing the same repair run concurrently).

I'd be more inclined to either : 

- Change the way "Pause" works and abort segments straight away (instead of letting them run)
or
- Handle your edge case in `resumeRunningRepairRuns` (which could change name at the same time) and use leader election to make sure no other reaper instance is already dealing with the segment

It could even be both as a Reaper instance could be stopped between pausing the run and aborting the segment. 

What do you think ?

smarsching:
Thanks for giving me more context regarding this issue.

I had missed that `resumeRunningRepairRuns` now aborts segments for paused jobs. However, I could not find the place from where it is called every 30 s (it only seems to be called once during application startup). Another potential issue might arise when a run is not in the paused, but in one of the other non-running states. In this case, the current implementation will not take care of aborting the segments.

I am going to think about how this could be improved. The most logical solution seems to do the operation over all runs, that are not in the running state. In addition to that, we probably have to ensure that a segment that has failed or timed out is not scheduled again when the respective run is not in the running state any longer. Also the abort code would need to handle the case where aborting the segment fails because the JMX connection is temporarily unavailable.

zznate:
Marking this as closed given the trauma we just introduced with the package name change. @smarsching Thank you for the initial effort regardless and please feel free to submit any additional PRs after updating to the new package structure.

PR #193, opened 2017-09-12
Refactored Travis to allow parallel Docker builds

ossarga:
- Created shell scripts for stage in Travis that conntained commands
- Moved commands from .travis.yml into their respective shell scripts
- Added a TEST_TYPE variable to allow different commands to be run at
   each Travis stage depending on that type
- Added a Docker test to the Travis build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,85 +1,55 @@
 sudo: required
 language: java
 jdk:
-- oraclejdk8
+  - oraclejdk8
+
 env:
   global:
     - LOCAL_JMX=no
     - TEST=""
   matrix:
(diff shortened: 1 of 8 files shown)


michaelsembwever (changes requested):
i like it!

can we have the travis-ci folder moved to under the 'src/' directory, or '.travis/' ?

(a fair bit of work went into cleaning out the root directory.)

ossarga:
Thanks. Good point. I am thinking of placing the travis scripts into the directory _src/ci/_

ossarga:
Merged as discussed with @michaelsembwever

PR #196, opened 2017-09-12
Change default configurations for more safety.

adejanovski:
For Cassandra backend yamls, use the token aware policy by default
and the DC aware policy to handle multi DC setups out of the box.
Explicitely set JMX connection timeout to 5s and reduce the repair manager
polling interval to speed up light repairs.

### DIFF
--- a/src/packaging/resource/cassandra-reaper-cassandra-ssl.yaml
+++ b/src/packaging/resource/cassandra-reaper-cassandra-ssl.yaml
@@ -10,8 +10,9 @@ hangingRepairTimeoutMins: 30
 storageType: cassandra
 enableCrossOrigin: true
 incrementalRepair: false
-enableDynamicSeedList: true
-repairManagerSchedulingIntervalSeconds: 30
+repairManagerSchedulingIntervalSeconds: 10
+activateQueryLogger: false
+jmxConnectionTimeoutInSeconds: 5
 
 # datacenterAvailability has three possible values: ALL | LOCAL | EACH
 # the correct value to use depends on whether jmx ports to C* nodes in remote datacenters are accessible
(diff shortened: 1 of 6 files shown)


src/packaging/resource/cassandra-reaper-cassandra-ssl.yaml
```diff
@@ -64,6 +65,14 @@ cassandra:
...
   contactPoints: ["127.0.0.1"]
   keyspace: reaper_db
+  loadBalancingPolicy:
+    type: tokenAware
+    shuffleReplicas: true
+    subPolicy:
+      type: dcAwareRoundRobin
+      localDC:
+      usedHostsPerRemoteDC: 0
+      allowRemoteDCsForLocalConsistencyLevel: false
```
michaelsembwever, line 75: i thought these were already the default settings in the code?

if not, then yes to the change,

but if so, is there a reason to explicit repeatly the default settings into the config? 
(this would make it harder to change it in the code and to be applied in new version updates)

michaelsembwever:
This branch has had a negative impact on the test success rate on circleci. circleci is the only integration tests that we have that do multi-dc testing.

I'm running it a few times to see just how flakey/broken it is, but already it looks more flakey than master.

And i see these in [these](https://circleci.com/gh/thelastpickle/cassandra-reaper/229) logs:
```
Caused by: java.lang.AssertionError: non-leader trying to update repair segment …
```

michaelsembwever:
@adejanovski wanna rebase this off master and see if master test improvements make it any difference?

adejanovski:
@michaelsembwever : those yaml files aren't used in the integration tests, which use custom yamls placed under src/test/resources. 
I'll perform the rebase to see if that helps though.

PR #197, opened 2017-09-13
Make asserts on schedules existing (having been created) eventual, as this is done by design eventually

michaelsembwever:
Should address some test failures.

This might be necessary in some of the other methods that assert of schedules, but i'm keen to see a report of those errors first.


adejanovski:
@michaelsembwever : That should definitely fix this test, but we need to generalize this to all eventually consistent reads. Any test that starts with "reaper_has" should use that kind of loop as we're using CL.LOCAL_ONE behind the scenes.

Do you think it should be done in this PR or as tests fail (you already have other failed tests in your Travis run on this PR) ?

michaelsembwever:
> Do you think it should be done in this PR or as tests fail (you already have other failed tests in your Travis run on this PR) ?

a lot of the tests have this already.
and those that don't _might_ not be as obvious as that, although it is quite likely most were just overlooked during the ft branch work.

so yeah, i was hoping to see a few more stack traces from test results.

PR #198, opened 2017-09-13
add maven-checkstyle-plugin and a ruleset based off google's codestyle (which seems to be what Spotify was using)

michaelsembwever:
if we can run a formatter once over the code, using this ruleset, we can then make the build break on new violations.


michaelsembwever:
The main code is formatted. 

Left to do is a test failure against postgres, and the test code to format.

michaelsembwever:
Complete. Open for review…

src/server/src/main/java/com/spotify/reaper/cassandra/JmxProxy.java
```diff
@@ -287,14 +289,14 @@ public String getDataCenter(String host) {
...
-            Lists.newArrayList(((StorageServiceMBean) ssProxy).getTokenToEndpointMap().keySet()), s -> new BigInteger(s));
+        Lists.newArrayList(((StorageServiceMBean) ssProxy).getTokenToEndpointMap().keySet()), s -> new BigInteger(s));
   }
 
   public Map<List<String>, List<String>> getRangeToEndpointMap(String keyspace) throws ReaperException {
     checkNotNull(ssProxy, "Looks like the proxy is not connected");
     try {
       return ((StorageServiceMBean) ssProxy).getRangeToEndpointMap(keyspace);
-    } catch (Exception e) {
+    } catch (RuntimeException e) {
```
adejanovski, line 299: This could leave us with uncatched exceptions, which sucks. Why are you changing this ?

adejanovski (approved):
Looks good and I've confirmed it builds and runs fine.
Merging it.

src/server/src/main/java/com/spotify/reaper/cassandra/JmxProxy.java
michaelsembwever: No that's not possible.
Here the only other exceptions, under `Exception`, would be checked exceptions. The only checked exception that could slip through here is if `getRangeToEndpointMap(keyspace)` threw a ReaperException, otherwise the code would not compile.  `getRangeToEndpointMap(keyspace)`  does not throw any checked exceptions.

The code won't catch Errors (or Throwables), but it didn't before either.

PR #204, opened 2017-09-21
Fix flaky test due to shared mock object between tests

adejanovski:
The ClusterResourceTest.testModifyClusterSeeds() test was often failing because the mock was a class member, reused through different tests.
Each test method now has its very own mocks.

### DIFF
--- a/src/server/src/test/java/com/spotify/reaper/unit/resources/ClusterResourceTest.java
+++ b/src/server/src/test/java/com/spotify/reaper/unit/resources/ClusterResourceTest.java
@@ -26,22 +26,141 @@ import static org.fest.assertions.api.Assertions.assertThat;
 import static org.hibernate.validator.internal.util.Contracts.assertNotNull;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.doReturn;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
 public class ClusterResourceTest {
 
-  String CLUSTER_NAME = "testcluster";
-  String PARTITIONER = "org.apache.cassandra.dht.RandomPartitioner";
-  String SEED_HOST = "TestHost";
-  URI SAMPLE_URI = URI.create("http://test");
-
-  AppContext context = new AppContext();
-  UriInfo uriInfo;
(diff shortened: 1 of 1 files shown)


src/server/src/test/java/com/spotify/reaper/unit/resources/ClusterResourceTest.java
```diff
@@ -31,126 +31,169 @@
...
+    clusterResource.addCluster(mocks.uriInfo, Optional.of(SEED_HOST));
 
-    when(jmxProxy.getLiveNodes()).thenReturn(Arrays.asList(SEED_HOST+1));
-    
-    Response response = clusterResource.modifyClusterSeed(uriInfo, CLUSTER_NAME,
-        Optional.of(SEED_HOST + 1));
+    when(mocks.jmxProxy.getLiveNodes()).thenReturn(Arrays.asList(SEED_HOST + 1));
+
+    Response response =
+        clusterResource.modifyClusterSeed(mocks.uriInfo, CLUSTER_NAME, Optional.of(SEED_HOST + 1));
```
michaelsembwever: fyi: google style will request that the operator `=` be the start of the new line.

this doesn't need to be fixed here, with a big format commit around the corner.

src/server/src/test/java/com/spotify/reaper/unit/resources/ClusterResourceTest.java
```diff
@@ -31,126 +31,169 @@
...
+          throws ReaperException {
+        return jmxProxy;
+      }
+    };
+
+    return new MockObjects(context, uriInfo, jmxProxy);
+
+  }
+
+  private class MockObjects {
```
michaelsembwever: i'm not too sure why this is an instance class? there's no state in the surrounding class to use.

or was this meant as, or should be `private static class MockObjects {` ?

michaelsembwever (commented):
approved after https://github.com/thelastpickle/cassandra-reaper/pull/204/files#r140359262 is addressed.

src/server/src/test/java/com/spotify/reaper/unit/resources/ClusterResourceTest.java
adejanovski: meant to be `private static class...` indeed. I'm making the change.

PR #210, opened 2017-09-27
Rename package from com.spotify.reaper to io.cassandrareaper

adejanovski:


michaelsembwever (changes requested):
LGTM.
And thanks for the cassandra->jmx change.

A couple of small things I see…
I see the following files still have references to spotify that should be changed.

 - src/packaging/build/usr/local/bin/cassandra-reaper
 - src/packaging/build/usr/local/bin/spreaper
 - src/server/src/checkstyle/suppressions.xml (this is why you had to fix RingRangeTest. best to now just remove the suppression, lines 9-10)

adejanovski:
@michaelsembwever, `src/packaging/build` is not in version control and it is regenerated each time the deb/rpm packages are built.
The original files have moved to `io.cassandrareaper` so it should be ok next time you run the package build.

I've pushed the modification for suppressions.xml.
I'll squash commits and merge.

PR #215, opened 2017-09-29
Add JMX connection test for the whole cluster on init to speed up sub…

adejanovski:
Currently, for some of the operations, Reaper will try to connect to any node in the cluster through JMX. 
When running on multi region clusters, this can lead to a lot of connection failures as JMX is not authorized on the cross region link.

With this PR, Reaper will build a map of accessible nodes during initialization so that subsequent calls are made on accessible nodes straight away.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/jmx/JmxConnectionFactory.java
+++ b/src/server/src/main/java/io/cassandrareaper/jmx/JmxConnectionFactory.java
@@ -30,6 +30,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 
 import com.codahale.metrics.MetricRegistry;
 import com.datastax.driver.core.policies.EC2MultiRegionAddressTranslator;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Optional;
 import com.google.common.collect.Maps;
 import org.slf4j.Logger;
@@ -37,14 +38,25 @@ import org.slf4j.LoggerFactory;
 
 public class JmxConnectionFactory {
 
-  private static final ConcurrentMap<String, AtomicInteger> SUCCESSFULL_CONNECTIONS = Maps.newConcurrentMap();
   private static final Logger LOG = LoggerFactory.getLogger(JmxConnectionFactory.class);
 
+  private final MetricRegistry metricRegistry;
+  private final HostConnectionCounters hostConnectionCounters;
   private Map<String, Integer> jmxPorts;
   private JmxCredentials jmxAuth;
   private EC2MultiRegionAddressTranslator addressTranslator;
   private boolean localMode = false;
-  private MetricRegistry metricRegistry;
+
+  @VisibleForTesting
+  public JmxConnectionFactory() {
+    this.metricRegistry = new MetricRegistry();
+    hostConnectionCounters = new HostConnectionCounters(metricRegistry);
+  }
+
+  public JmxConnectionFactory(MetricRegistry metricRegistry) {
+    this.metricRegistry = metricRegistry;
+    hostConnectionCounters = new HostConnectionCounters(metricRegistry);
+  }
 
   public JmxProxy connect(Optional<RepairStatusHandler> handler, String host, int connectionTimeout)
       throws ReaperException {
@@ -86,13 +98,13 @@ public class JmxConnectionFactory {
         String host = hostIterator.next();
 
         // First loop, we try the most accessible nodes, then second loop we try all nodes
-        if (null != host && (SUCCESSFULL_CONNECTIONS.getOrDefault(host, new AtomicInteger(0)).get() >= 0 || 1 == i)) {
+        if (null != host && (hostConnectionCounters.getSuccessfulConnections(host) >= 0 || 1 == i)) {
           try {
             JmxProxy jmxProxy = connect(handler, host, connectionTimeout);
-            incrementSuccessfullConnections(host);
+            hostConnectionCounters.incrementSuccessfulConnections(host);
             return jmxProxy;
           } catch (ReaperException | RuntimeException e) {
-            decrementSuccessfullConnections(host);
+            hostConnectionCounters.decrementSuccessfulConnections(host);
             LOG.info("Unreachable host", e);
           }
         }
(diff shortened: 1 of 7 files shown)


src/server/src/main/java/io/cassandrareaper/jmx/JmxConnectionFactory.java
```diff
@@ -19,26 +19,36 @@
...
 import com.google.common.base.Optional;
 import com.google.common.collect.Maps;
+import jersey.repackaged.com.google.common.collect.Lists;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class JmxConnectionFactory {
 
-  private static final ConcurrentMap<String, AtomicInteger> SUCCESSFULL_CONNECTIONS = Maps.newConcurrentMap();
+  private static final ConcurrentMap<String, AtomicInteger> successfulConnections =
```
michaelsembwever: nit: should be `SUCCESSFULL_CONNECTIONS` and `INIT_SEEDS_EXECUTOR`.

i'll open a separate PR to add to checkstyle.xml 
```
<module name="ConstantName"/>
```
for this, so it's optional to fix here.
(although if it were fixed here it would reduce the PR size)

src/server/src/main/java/io/cassandrareaper/jmx/JmxConnectionFactory.java
```diff
@@ -163,4 +175,55 @@ private void decrementSuccessfullConnections(String host) {
...
+    try {
+      List<Future<Optional<String>>> endpointFutures =
+          initSeedsJmxExecutor.invokeAll(
+              jmxTasks, JmxProxy.JMX_CONNECTION_TIMEOUT, JmxProxy.JMX_CONNECTION_TIMEOUT_UNIT);
+
+      for (Future<Optional<String>> endpointFuture : endpointFutures) {
+        try {
+          endpointFuture.get(JmxProxy.JMX_CONNECTION_TIMEOUT, JmxProxy.JMX_CONNECTION_TIMEOUT_UNIT);
+        } catch (RuntimeException | ExecutionException | TimeoutException e) {
+          // We expect to have failures here, so using TRACE level logs
```
michaelsembwever: if the exception is expected then name the exception `e` as instead `expected`, or `ignore`, and remove this comment line. that'll explain it enough.

src/server/src/main/java/io/cassandrareaper/jmx/JmxConnectionFactory.java
```diff
@@ -163,4 +175,55 @@ private void decrementSuccessfullConnections(String host) {
...
+          this.connectAny(Optional.absent(), endpoints, JmxProxy.JMX_CONNECTION_TIMEOUT)) {
+
+        return Optional.of(endpoints.get(0));
+
+      } catch (RuntimeException e) {
+        LOG.debug("failed to connect to hosts {} through JMX", endpoints.get(0), e);
+        return Optional.absent();
+      }
+    };
+  }
```
michaelsembwever: I don't understand this method…

why couldn't it be:
```
  private Callable<Optional<String>> testJmxConnection(String endpoint) {
    return () -> {
      try (JmxProxy jmxProxy = connect(Optional.absent(), endpoint, JmxProxy.JMX_CONNECTION_TIMEOUT)) {
        incrementSuccessfullConnections(endpoint);
        return Optional.of(endpoint);
      } catch (ReaperException | RuntimeException e) {
        decrementSuccessfullConnections(host);
        LOG.debug("failed to connect to hosts {} through JMX", endpoints.get(0), e);
        return Optional.absent();
      }
    };
  }
```

src/server/src/main/java/io/cassandrareaper/jmx/JmxConnectionFactory.java
```diff
@@ -163,4 +175,55 @@ private void decrementSuccessfullConnections(String host) {
       LOG.warn("Could not decrement JMX successfull connections counter for host {}", host, e);
     }
   }
+
+
```
michaelsembwever: Could the following three methods (and the initSeedsJmxExecutor field) be encapsulated into a separate class.

If so then such a class could be AutoClosable so we could close the ExecutorService as a resource like
```
 private void initializeJmxSeedsForAllClusters() {
    LOG.info("Initializing JMX seed list for all clusters...");
    try (
         JmxConnectionsIntializer jmxConnectionsIntializer = new JmxConnectionsIntializer(context.jmxConnectionFactory);
         Timer.Context cxt = context.metricRegistry.timer(MetricRegistry.name(JmxConnectionFactory.class, "jmxConnectionsIntializer")).time()) {

      context.storage.getClusters()
          .parallelStream()
          .forEach(cluster -> jmxConnectionsIntializer.on(cluster));
    }
    LOG.info("Initialized JMX seed list for all clusters.);
  }
```


(I think adding metrics is needed here, and i think inside `jmxConnectionsIntializer.on(cluster)` and inside `JmxProxy.connect(..)`)

(given this initialisation may delay startup some time it we might start thinking about implementing a `AppContext.state` enum that the ping and healthcheck and UI can take advantage of.)

michaelsembwever (changes requested):
it would be nice to see more encapsulation of new methods, …and metrics.

src/server/src/main/java/io/cassandrareaper/jmx/JmxConnectionFactory.java
michaelsembwever: see https://github.com/thelastpickle/cassandra-reaper/pull/217

adejanovski: Fixed

src/server/src/main/java/io/cassandrareaper/jmx/JmxConnectionFactory.java
adejanovski: Fixed

src/server/src/main/java/io/cassandrareaper/jmx/JmxConnectionFactory.java
adejanovski: I tried to avoid duplicating the logic of maintaining the counters for successful connections which is already implemented in `connectAny()`.

michaelsembwever: Could we encapsulate too these counters into a separate class, like…
```
static final class HostConnectionsInfo {

  private final ConcurrentMap<String, AtomicInteger> connectionCounts = Maps.newConcurrentMap();

  void incrementSuccessfullConnections(String host) {
    try {
      AtomicInteger successes = connectionCounts.putIfAbsent(host, new AtomicInteger(1));
      if (null != successes && successes.get() <= 20) {
        successes.incrementAndGet();
      }
      LOG.debug("Host {} has {} successfull connections", host, successes);
      if (null != metricRegistry) {
        metricRegistry
            .counter(MetricRegistry.name(JmxConnectionFactory.class, "connections", host.replace('.', '-')))
            .inc();
      }
    } catch (RuntimeException e) {
      LOG.warn("Could not increment JMX successfull connections counter for host {}", host, e);
    }
  }

  void decrementSuccessfullConnections(String host) {
    try {
      AtomicInteger successes = connectionCounts.putIfAbsent(host, new AtomicInteger(-1));
      if (null != successes && successes.get() >= -5) {
        successes.decrementAndGet();
      }
      LOG.debug("Host {} has {} successfull connections", host, successes);
      if (null != metricRegistry) {
        metricRegistry
            .counter(MetricRegistry.name(JmxConnectionFactory.class, "connections", host.replace('.', '-')))
            .dec();
      }
    } catch (RuntimeException e) {
      LOG.warn("Could not decrement JMX successfull connections counter for host {}", host, e);
    }
  }
}
```

Re-using the logic: which is only calling the increment/decrement methods; makes more sense then.

src/server/src/main/java/io/cassandrareaper/jmx/JmxConnectionFactory.java
adejanovski: The UI and the REST API will be unavailable until the init phase is over, so we'll need there's no feedback we could give them anyway.

src/server/src/main/java/io/cassandrareaper/jmx/JmxConnectionFactory.java
adejanovski: I've moved the counters into a separate class as suggested.

Looks good ?

adejanovski:
Squashed and merged

PR #219, opened 2017-10-03
Tunable jmx timeouts in docker

joaquincasares:
@adejanovski I've included the jmx timeout environmental variable so that it's now exposed via Docker.

I was hoping to help out consolidating the yamls to have one set of yamls/ordering/comments going forward, but it seems as though everything is spread around. What is the "de-facto" yaml that I should always base myself on?

As far as the Docker yaml, it seems as though everything except `metrics` is currently being exposed when using `src/server/src/test/resources/cassandra-reaper.yaml` as my base yaml. Let me know if you want me to add the `metrics` key and if so, what defaults would you want?

Here's one of the defaults I found:

```
metrics:
  frequency: 1 second
  reporters:
    - type: csv
      file: target/dropwizard-metrics
```


adejanovski (changes requested):
@joaquincasares, you should use these files as a reference : https://github.com/thelastpickle/cassandra-reaper/tree/master/src/packaging/resource

But some variables could be different depending on the backend.
For Cassandra, you're mainly missing the `useAddressTranslator` setting and the `localDC` one.

joaquincasares:
Thanks @adejanovski ! The requested changes have been committed. I'll be merging this into master.

PR #220, opened 2017-10-05
Table blacklist feature

adejanovski:
The PR adds the possibility to blacklist some tables from a repair run/schedule.

This allows to repair all tables but those specified and maybe have different schedules for specific tables while not having to list all the tables.
The UI has been improved to use autocompletion and tag like input for selecting the keyspace and the tables :
pasted image at 2017_10_05 08_45 am

There are some changes in the repair screen too with animated progress bars for running repairs and displayed ETA among others :
cassandra reaper repair 2


src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
```diff
@@ -58,6 +58,7 @@
 
 import com.google.common.base.Optional;
 import jersey.repackaged.com.google.common.collect.Lists;
+import jersey.repackaged.com.google.common.collect.Maps;
```
michaelsembwever: why not use the original `Lists` and `Maps` ?

src/server/src/main/resources/db/h2/V3_0_0__table_blacklist.sql
```diff
@@ -3,4 +3,4 @@
 --
 
 ALTER TABLE repair_unit 
-ADD blacklisted_tables ARRAY NOT NULL;
\ No newline at end of file
+ADD blacklisted_tables ARRAY;
```
michaelsembwever: nit: this should be fixed in the previous commit (every sha in the PR should build and test ok, otherwise we break eg `git bisect`)

michaelsembwever:
Do any of the tests actually test against specific tables? or test using a non-empty blacklist?

michaelsembwever (approved):
LGTM. thanks for the tests :-)

PR #221, opened 2017-10-08
In cassandra storage override pooling options to allow bigger queue and longer timeout

michaelsembwever:
We fetch every repair run in a separate async execute, so the default queue length of 256 will be quickly hit.

ref: https://gitter.im/thelastpickle/cassandra-reaper?at=59d66c7be44c43700afa2fce

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
+++ b/src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
@@ -45,6 +45,7 @@ import java.util.stream.Collectors;
 import com.datastax.driver.core.BatchStatement;
 import com.datastax.driver.core.CodecRegistry;
 import com.datastax.driver.core.ConsistencyLevel;
+import com.datastax.driver.core.PoolingOptions;
 import com.datastax.driver.core.PreparedStatement;
 import com.datastax.driver.core.QueryLogger;
 import com.datastax.driver.core.QueryOptions;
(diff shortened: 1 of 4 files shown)


src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
```diff
@@ -1022,6 +1017,43 @@ public void saveHeartbeat() {
...
+    }
+    cassandraFactory.setRetryPolicy(java.util.Optional.of((RetryPolicyFactory) () -> new RetryPolicyImpl()));
+  }
+
+  private static void overridePoolingOptions(CassandraFactory cassandraFactory) {
+    PoolingOptionsFactory newPoolingOptionsFactory = new PoolingOptionsFactory() {
+      @Override
+      public PoolingOptions build() {
+        if (null == getPoolTimeout()) {
+          setPoolTimeout(Duration.seconds(20));
```
michaelsembwever: i'd like to bump this up to 60s. writes in-flight are in-memory and non-durable. 
so the longer we can hold them for: the more durable they are.

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
```diff
@@ -1022,6 +1017,43 @@ public void saveHeartbeat() {
...
+        }
+        return super.build().setMaxQueueSize(4096);
```
michaelsembwever: i'd like to bump this up to 4k. writes in-flight are in-memory and non-durable. 
so the longer we can hold them for: the more durable they are.

adejanovski (approved):
LGTM

michaelsembwever:
The trade-off with a potentially long queue like this is (up to 4k) in-flight requests can get dropped (eg if the reaper instance crashes).

This is going to most likely happen when:
 - reading segments out of the repair_run table, or
 - writing segments when first creating a repair

; as that's where the high volumes of async requests occur.

Both of these situations are non-critical. Either reading the repair run page fails, or creating a repair run fails. As creating a repair run anyway needs to be started in a subsequent step this is not critical. In Reaper the worse situation should be that a segment has to be repaired again (if may or may not have previously completed).

The potentially long queue length can also cause memory pressure. Reaper's default heap setting is large enough though that it should otherwise be able to handle it.

PR #222, opened 2017-10-09
Fix for properly handling reboots of Reaper instances with distributed storage

adejanovski:
The previous behavior was to try aborting such segments only if there was no repair runner for the segment's run.
This would fail if leader row had not expired yet (which is what will happen when the reboot is takes less than 3mn).
Now, in the case of distributed storage, we try to cancel all RUNNING segments with no leader every 30s.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/service/RepairManager.java
+++ b/src/server/src/main/java/io/cassandrareaper/service/RepairManager.java
@@ -23,10 +23,13 @@ import io.cassandrareaper.jmx.JmxProxy;
 import io.cassandrareaper.storage.IDistributedStorage;
 
 import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
 import java.util.Map;
 import java.util.UUID;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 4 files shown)


src/server/src/main/java/io/cassandrareaper/service/RepairManager.java
```diff
@@ -82,11 +85,28 @@ public void resumeRunningRepairRuns(AppContext context) throws ReaperException {
...
+                : Collections.emptyList();
+
+        abortSegments(
+            runningSegments
+                .stream()
+                .filter(segment -> !activeLeaders.contains(segment.getId()))
+                .collect(Collectors.toSet()),
+            context,
+            repairRun);
+      }
```
michaelsembwever: trivial: could we put this chunk into a private method with a descriptive name? could make this already long method a bit more readable (without resorting to using comments)

michaelsembwever (approved):
LGTM

src/server/src/main/java/io/cassandrareaper/service/RepairManager.java
adejanovski: Done and force pushed

Issue #229, opened 2017-10-12
Null pointer exception in onReadTimeout


hkroger:
We are getting an exception while running reaper with Cassandra backend.

```
ERROR  [2017-10-10 11:49:26,442] [dw-2823 - GET /repair_run] c.s.r.r.RepairRunResource - Failed listing cluster statuses
com.datastax.driver.core.exceptions.DriverInternalError: Unexpected exception thrown
        at com.datastax.driver.core.DriverThrowables.propagateCause(DriverThrowables.java:39)
        at com.datastax.driver.core.DefaultResultSetFuture.getUninterruptibly(DefaultResultSetFuture.java:245)
        at com.spotify.reaper.storage.CassandraStorage.getRepairRunsAsync(CassandraStorage.java:342)
        at com.spotify.reaper.storage.CassandraStorage.getRepairRunsForCluster(CassandraStorage.java:309)
        at com.spotify.reaper.resources.RepairRunResource.listRepairRuns(RepairRunResource.java:455)
        at sun.reflect.GeneratedMethodAccessor27.invoke(Unknown Source)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at org.glassfish.jersey.server.model.internal.ResourceMethodInvocationHandlerFactory$1.invoke(ResourceMethodInvocationHandlerFactory.java:81)
        at org.glassfish.jersey.server.model.internal.AbstractJavaResourceMethodDispatcher$1.run(AbstractJavaResourceMethodDispatcher.java:144)
        at org.glassfish.jersey.server.model.internal.AbstractJavaResourceMethodDispatcher.invoke(AbstractJavaResourceMethodDispatcher.java:161)
        at org.glassfish.jersey.server.model.internal.JavaResourceMethodDispatcherProvider$ResponseOutInvoker.doDispatch(JavaResourceMethodDispatcherProvider.java:160)
        at org.glassfish.jersey.server.model.internal.AbstractJavaResourceMethodDispatcher.dispatch(AbstractJavaResourceMethodDispatcher.java:99)
        at org.glassfish.jersey.server.model.ResourceMethodInvoker.invoke(ResourceMethodInvoker.java:389)
        at org.glassfish.jersey.server.model.ResourceMethodInvoker.apply(ResourceMethodInvoker.java:347)
[...]

PR #231, opened 2017-10-12
Fix NPE in CassandraStorage, when a read times out and the statement is null

michaelsembwever:
ref: https://github.com/thelastpickle/cassandra-reaper/issues/229


adejanovski:
Yup, I had a similar fix in the works because I ran into that issue. You've been faster than me :)

Merging

PR #238, opened 2017-10-16
Switch to a segment count per node

adejanovski:
As the segment count is a misleading notion and doesn't currently scale with the cluster, we have decided to move to a segment count per node.
By default, we will use max(16, number_of_vnodes) but it can be overriden as it used to.

To simplify Reaper's usage, all optional settings for repairs and schedules were moved into an advanced section in the UI.


michaelsembwever:
the last commit should be squashed into the first (otherwise `git bisect …` will break)

michaelsembwever (changes requested):
two small nits.

otherwise review is complete and LGTM.

adejanovski:
I've addressed the requested changes.

Merging.

Issue #240, opened 2017-10-18
Using NULL for column values show up as tombstones in Cassandra


mrwulf:
I'm using the cassandra storage engine with Cassandra 3.11.0. In my Cassandra logs there are lots of messages like:
```
WARN  [ReadStage-5] 2017-10-18 11:04:30,610 ReadCommand.java:522 - Read 1547 live rows and 3311 tombstone cells for query SELECT * FROM reaper_db.repair_run WHERE id = 71d1c7e0-b426-11e7-8477-b7c57b58f240 LIMIT 5000 (see tombstone_warn_threshold)
WARN  [ReadStage-8] 2017-10-18 11:04:30,639 ReadCommand.java:522 - Read 2306 live rows and 2838 tombstone cells for query SELECT * FROM reaper_db.repair_run WHERE id = 6c33a9c0-aae4-11e7-bc2b-155801446996 LIMIT 5000 (see tombstone_warn_threshold)
WARN  [ReadStage-2] 2017-10-18 11:04:30,784 ReadCommand.java:522 - Read 2266 live rows and 6155 tombstone cells for query SELECT * FROM reaper_db.repair_run WHERE id = 6ae688f0-b23a-11e7-9d50-79c9c6ee5ac0 LIMIT 5000 (see tombstone_warn_threshold)
WARN  [ReadStage-3] 2017-10-18 11:04:30,853 ReadCommand.java:522 - Read 2266 live rows and 6049 tombstone cells for query SELECT * FROM reaper_db.repair_run WHERE id = 99202530-b2cd-11e7-a3fa-6904da3676a4 LIMIT 5000 (see tombstone_warn_threshold)
```

I believe this is being caused by writing null values instead of leaving columns unset ([here](https://opencredo.com/cassandra-tombstones-common-issues/) is a blog referencing why not to use nulls). I could just change the warn threshold, but I would prefer to store only the limited, necessary data.

mrwulf:
Actually, TLP did an article about it as well! http://thelastpickle.com/blog/2016/09/15/Null-bindings-on-prepared-statements-and-undesired-tombstone-creation.html

mrwulf:
Can anyone speak to why the repair_run table is used for both repairs and individual segments being repaired?

michaelsembwever:
fixed in https://github.com/thelastpickle/cassandra-reaper/pull/244 which is now in master.

PR #243, opened 2017-10-18
Unset null columns to prevent creating tombstones

mrwulf:
Addresses #240


michaelsembwever:
Thanks for the report @mrwulf !

This approach won't work because we also have to support Cassandra-2.1.
Furthermore the null columns have (sometimes) meaning.

Will provide a separate pull request and am keen for your review.

PR #244, opened 2017-10-19
For Cassandra storage minimise the number of tombstones (NULL entries) in the repair_run table

michaelsembwever:
Remove RepairSegment.repairCommandId as no one was using/storing it.

Make explicit which fields in RepairSegment can be null and be nulled.
Hide the RepairSegment.Builder constructor.

Add Preconditions and asserts in ensuring lifecycle state is correct and null fields only exist as/when appropriate.
Preconditions are used when checks are fast and for providing fail-fast design. Asserts are used where checks are expensive and/or are secondary checks.

ref: https://github.com/thelastpickle/cassandra-reaper/issues/240

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/core/RepairSegment.java
+++ b/src/server/src/main/java/io/cassandrareaper/core/RepairSegment.java
@@ -18,7 +18,9 @@ import io.cassandrareaper.service.RingRange;
 
 import java.math.BigInteger;
 import java.util.UUID;
+import javax.annotation.Nullable;
 
+import com.google.common.base.Preconditions;
 import org.joda.time.DateTime;
 
 public final class RepairSegment {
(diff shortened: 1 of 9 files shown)


michaelsembwever:
@mrwulf

michaelsembwever:
The changes in CassandraStorage form the crux of the patch.

michaelsembwever:
the joy of codebases that don't treat nulls explicitly… :-/
 at least this is one step in the right direction…

adejanovski (approved):
Tested and approved.
We're not generating tombstones anymore here (or at least very few when nulling the coordinator for example).

adejanovski:
Last change most probably fixed the issue with the race condition between the JMX success notification and the abort of leaderless segments.
At least I'm unable to reproduce it.

Let's go ahead and merge.

Issue #62, opened 2017-03-10
Deploy to Maven Central


jvanzyl:
Hi, we'd like to use your project internally and it's much easier if we can retrieve your build artifacts from Maven Central. Happy to help with a pull request and help get you setup in Maven Central if you need a hand with this.

zznate:
Hi @jvanzyl  - yes we have definitely discussed this. That said, since it's **you** of all folks making the offer, we'd be psyched to take such a pull request.

adejanovski:
Deployments to Maven Central are now done for every release.

PR #246, opened 2017-10-20
Automating release process using rultor

velo:
Using rultor.com to release to maven central #62 

Just need to create a ticket with this message

```
@rultor release, tag=`0.8`
```

Rultor will automatically build and publish it

### DIFF
--- a/.rultor.yml
+++ b/.rultor.yml
@@ -1,0 +1,23 @@
+architect:
+- velo
+
+docker:
+  image: "velo/rultor8"
+
+assets:
+  secring.gpg: thelastpickle/rultor-assets#maven/secring.gpg
+  settings.xml: thelastpickle/rultor-assets#maven/settings.xml
+  pubring.gpg: thelastpickle/rultor-assets#maven/pubring.gpg
+
+env:
+  MAVEN_OPTS: -XX:MaxPermSize=256m -Xmx1g
+
+release:
+  script: |-
+    # first thing, change version on poms
+    mvn versions:set "-DnewVersion=${tag}"
+    git commit -am "${tag}"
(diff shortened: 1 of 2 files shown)


.rultor.yml
```diff
@@ -0,0 +1,39 @@
+#
+# Copyright (C) 2017 Marvin Herman Froeder (velobr@gmail.com)
```
michaelsembwever: files in this project do not include copyright statements.

(if they did we would want them all to one entity)

pom.xml
```diff
@@ -28,6 +30,49 @@
...
+    </developer>
+  </developers>
+
+  <contributors>
+    <contributor>
+      <name>Marvin Herman Froeder</name>
+      <email>velo br at gmail dot com</email>
+      <url>about.me/velo</url>
+    </contributor>
+  </contributors>
```
michaelsembwever, line 74: i'd take away developers and contributors section. is it needed for rultor?

this information is stale and superfluous. live versions of it can be found via the changelog or github insights.

velo: `developers` is required to publish on central
http://central.sonatype.org/pages/requirements.html#developer-information

contributors not.

.rultor.yml
velo: Ops, my bad.

I just used a template I have for my OSS stuff.

pom.xml
michaelsembwever: Ok.

Then can we add some of the previous developers, so it doesn't look like TLP has completely taken over.

michaelsembwever (approved):

michaelsembwever:
has been merged to master

PR #247, opened 2017-10-22
set heap sizes for travis builds to the same as circleci

michaelsembwever:
circleci has displayed better stability. re-use on travis.


michaelsembwever:
with three travis builds without any failure, enough proof is there that this improves stability.

PR #235, opened 2017-10-15
Give datacenterAvailability = EACH better chances to work, through updating all reachable node metrics as part of heartbeat

michaelsembwever:
Poorman's solution of updating NodeMetrics on all nodes, via the heartbeat mechanism.
Avoids the complexities of any queue, or distributed read-before-write approach.
Still runs the risk of outdated node metrics (although any attempt at re-implementing the notion of gossip will…)

There's a number of questions here to ask and go through…
 (eg performance of 3x jmx calls to every node in a cluster every minute…)

And tests to add.

An alternative to this was the idea of how we choose the next segment to repair, rather than making it random. The idea that parallel token ranges (with the same vnode) would be attempted across datacenters…

Before merging into master, both #248 and @249 should be merged into this. The tests for this feature are in the latter #249.

PR #248, opened 2017-10-22
refactor test packages and break apart and remove CommonTools

michaelsembwever:
Can be merged into #235 ,  when/if #235 is accepted.


adejanovski (approved):
LGTM

PR #252, opened 2017-10-26
Add request mechanism for metrics

adejanovski:
This PR allows the EACH setting to work by implementing a request mechanism when grabbing metrics to determine if a segment can be processed.
Metrics collection happens as follows now : 

- Reaper will try to access all nodes through JMX to get the number of pending compactions and check if a repair is already running
- If JMX connection failed on some nodes, Reaper will try to get the metrics from the C* backend, in the `node_metrics` table
- If some nodes metrics are still missing, Reaper will add a row for each missing node in the `node_metrics_request` table. The table is partitioned by run_id and 1 minute wide time partitions to have short lived partitions and prevent from reading a lot of tombstones over time.
- Reaper will wait at most 2 minutes, checking every 10 seconds if the requested metrics have been added to `node_metrics`
- Every 10s, the heartbeat system of all Reaper instances will check for node metrics requests within the current minute and the previous one. If there are such rows, they will try and grab the metrics through JMX, store them in the `node_metrics` table and delete the request from `node_metrics_request`
- The Reaper instance who issued the request will process the segment if all metrics were provided within 2 minutes. Otherwise, the segment will be postponed.

Writes in the `node_metrics_request` table are made at `QUORUM` and reads at `LOCAL_QUORUM` to maximize odds of success.
Reads should be pretty cheap since we're only accessing 2 short lived partitions every 10s.


src/server/src/main/java/io/cassandrareaper/service/Heart.java
```diff
@@ -93,80 +99,97 @@ public void close() {
...
+                                });
+                      })
+                  .get();
+            } catch (ExecutionException | InterruptedException | RuntimeException ex) {
+              LOG.warn("failed updateAllReachableNodeMetrics submission", ex);
+            } finally {
+              assert updatingNodeMetrics.get();
+              updatingNodeMetrics.set(false);
+            }
+          });
```
michaelsembwever, line 183: 👍  this was along the lines that i'd envisioned :-)

i'm not too sure why the nesting is deeper than before, would have expected the opposite. will take a look…

michaelsembwever:
I'm presuming the second commit "Fix minor UI issues with autocomplete fields" is not intended part of this PR.

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
```diff
@@ -578,9 +596,41 @@ private void handlePotentialStuckRepairs(LazyInitializer<Set<String>> busyHosts,
...
+          .requestNodeMetrics(
+              NodeMetricsRequest.builder()
+                  .withCluster(clusterName)
+                  .withHostAddress(hostName)
+                  .withRunId(repairRunner.getRepairRunId())
+                  .withSegmentId(segmentId)
+                  .build());
+
+      // Waiting for other Reaper instances to put the metrics in the database for that node
+      await()
```
michaelsembwever, line 617: i've never seen Awaitability used in production code before…

michaelsembwever:
Work is ongoing in https://github.com/thelastpickle/cassandra-reaper/tree/mck/datacenter_availability_EACH_heartbeats-with-alex-request-model
to see if we can take the best from this PR and #235 and create something simpler.

one of the main discoveries was that we don't need both the `node_metrics` and `node_metrics_requested` tables. that is that the `node_metrics` table only serves any purpose when running with EACH, and everything we need can fit in a datamodel more like @adejanovski designed in `node_metrics_requested`.

michaelsembwever:
This work has been incorporated back into https://github.com/thelastpickle/cassandra-reaper/pull/235

PR #256, opened 2017-10-28
Add metrics for repair progress + time since last repair. Fixes #207.

### LINKED ISSUE #207: Metrics that would be good to have
After a chat with @adejanovski about if Reaper exported metrics, he mentioned its currently in master but not many metrics were exported yet, and if I had any requests to make a GitHub issue. For us, useful metrics would be:

- Number of segments pending to be repaired (so we can chart progress over time, similar to Cassandra's own Repair PendingTasks metric
- Number of segments repaired per second (likely to be low for big repairs but still handy)
- Number of postponed repair events due to high load / repairs already running per second.

rzvoncek:
This change adds in two metrics:
* `repairProgress` exposes the percentage of segments done for the most recent repair run for a cluster and keyspace.
* `millisSinceLastRepair` exposes milliseconds elapsed since a repair run was completed for a particular keyspace and table.

The metrics look like this:
```
"gauges" : {
    "io.cassandrareaper.service.RepairRunner.millisSinceLastRepair.testcluster.keyspace1" : {
      "value" : 156732
    },
    "io.cassandrareaper.service.RepairRunner.repairProgress.testcluster.keyspace1" : {
      "value" : 0.1875
    }
}
```

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
+++ b/src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
@@ -33,6 +33,8 @@ import java.util.concurrent.ThreadLocalRandom;
 import java.util.concurrent.atomic.AtomicReferenceArray;
 import java.util.stream.Collectors;
 
+import com.codahale.metrics.Gauge;
+import com.codahale.metrics.MetricRegistry;
 import com.google.common.base.Optional;
 import com.google.common.collect.Collections2;
 import com.google.common.collect.Lists;
@@ -55,6 +57,8 @@ final class RepairRunner implements Runnable {
   private JmxProxy jmxConnection;
   private final AtomicReferenceArray<UUID> currentlyRunningSegments;
   private final List<RingRange> parallelRanges;
+  private final String metricNameForMillisSinceLastRepair;
+  private float repairProgress;
 
   RepairRunner(AppContext context, UUID repairRunId) throws ReaperException {
     LOG.debug("Creating RepairRunner for run with ID {}", repairRunId);
@@ -95,6 +99,11 @@ final class RepairRunner implements Runnable {
     parallelRanges = getParallelRanges(
         parallelRepairs,
         Lists.newArrayList(Collections2.transform(repairSegments, segment -> segment.getTokenRange())));
+
+    String repairUnitClusterName = repairUnitOpt.get().getClusterName();
+    String metricNameForRepairProgress = metricName("repairProgress",  repairUnitClusterName, repairRunId);
+    context.metricRegistry.register(metricNameForRepairProgress, (Gauge<Float>) ()  -> repairProgress);
+    metricNameForMillisSinceLastRepair = metricName("millisSinceLastRepair", repairUnitClusterName, repairRunId);
   }
 
   UUID getRepairRunId() {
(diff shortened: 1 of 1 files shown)


rzvoncek:
The travis failed because of [this](https://travis-ci.org/thelastpickle/cassandra-reaper/jobs/294063896#L5179).

src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
```diff
@@ -481,4 +501,18 @@ public void killAndCleanupRunner() {
...
+    String cleanClusterName = clusterName.replaceAll("[^A-Za-z0-9]", "");
+    String cleanKeyspaceName = keyspace.replaceAll("[^A-Za-z0-9]", "");
+    return MetricRegistry.name(RepairRunner.class, metric, cleanClusterName, cleanKeyspaceName);
+  }
+
+  private void updateRepairProgressMetric(int segmentsDone, int segmentsTotal) {
+    context.metricRegistry.remove(metricNameForRepairProgress);
+    context.metricRegistry.register(
+        metricNameForRepairProgress,
+        (Gauge<Float>) () -> (float) segmentsDone / segmentsTotal);
```
michaelsembwever: this is a constant gauge. it feels a bit clumsy to bump it this way.

why not make `segmentsDone` a field, so the gauge does not need to be removed and re-registered again and again during the one repair…

src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
```diff
@@ -96,6 +100,9 @@ public RepairRunner(AppContext context, UUID repairRunId) throws ReaperException
     parallelRanges = getParallelRanges(
         parallelRepairs,
         Lists.newArrayList(Collections2.transform(repairSegments, segment -> segment.getTokenRange())));
+
+    metricNameForMillisSinceLastRepair = metricName("millisSinceLastRepair", keyspace);
+    metricNameForRepairProgress = metricName("repairProgress", keyspace);
```
michaelsembwever: these names are not unique to the repair unit. 
ie there's no indication of cluster or tables involved. 

and it could also be future-proofed against multiple repair runners running in parallel…

rzvoncek:
Thanks @michaelsembwever for the comments! 

I've changed the repair progress gauge to be registered only once and report a field in 5289bb8.

In 5aa60e6 Ive changed how the metrics are named. The format is now:

```
    "io.cassandrareaper.service.RepairRunner.millisSinceLastRepair.testcluster.4cc7d380-bd5a-11e7-a9fc-dbed8745a5eb" : {
      "value" : 422789
    },
    "io.cassandrareaper.service.RepairRunner.repairProgress.testcluster.4cc7d380-bd5a-11e7-a9fc-dbed8745a5eb" : {
      "value" : 1.0
    },
```

So it is `.repairProgress.<cluster_name>.<repair_run_id>`. Both fields are taken straight from the `RepairUnit` fields. I've opted for `repair_run_id` because @adejanovski  suggested there can potentially be a lot of tables involved and they shouldn't show up in the metric name.

With these changes, we should be able to avoid clashes for parallel RepairRunners on the same keyspace, and even tables.

There is still the scenario of multiple Reaper instances reporting the same metrics. This, I believe, is easier to solve by the monitoring stack itself (e.g. by labelling).

michaelsembwever:
Good stuff @rzvoncek 

I'm thinking that the repair run IDs should have the hyphens stripped out of them.
Both dots and hyphens gets replaced with underscores in prometheus and cause troubles. 
As far as the metrics go we don't need those hyphens in the UUIDs.

Otherwise LGTM. (I presume you'll squash the commits into one before merging.)

michaelsembwever (approved):

rzvoncek:
I've removed the dashes from the UUIDs that make it to the metric name.

I've also squashed all my changes and rebased on master. Will merge as soon as the CI goes through.

PR #257, opened 2017-10-30
cassandra fix table properties

michaelsembwever:
Previous attempt in `002_table_properties.cql` overwrote and disabled the key cache setting.

This restores key cache to be enabled on every table.

It also completely disables `dclocal_read_repair`, which has no purpose for reaper.

### DIFF
--- a/src/server/src/main/resources/db/cassandra/010_table_properties.cql
+++ b/src/server/src/main/resources/db/cassandra/010_table_properties.cql
@@ -1,0 +1,30 @@
+
+-- Optimal table properties
+
+ALTER TABLE repair_unit_v1
+  WITH caching = {'keys':'ALL', 'rows_per_partition':'1'}
+  AND dclocal_read_repair_chance = 0.0;
+
+ALTER TABLE repair_schedule_by_cluster_and_keyspace
+  WITH caching = {'keys':'ALL', 'rows_per_partition':'10'}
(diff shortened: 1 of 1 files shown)


src/server/src/main/resources/db/cassandra/009_table_properties.cql
```diff
@@ -0,0 +1,30 @@
+
+-- Optimal table properties
+
+ALTER TABLE repair_unit_v1
+  WITH caching = {'keys':'ALL', 'rows_per_partition':'1'}
+  AND dclocal_read_repair_chance = 0.0;
+
+ALTER TABLE repair_schedule_by_cluster_and_keyspace
+  WITH caching = {'keys':'NONE', 'rows_per_partition':'10'}
```
adejanovski: Why still `NONE` for key cache here ?

michaelsembwever: thanks for catching that!

adejanovski (changes requested):
This PR now conflicts with #261 which has just been merged. They both use the same version number for schema migration.

Here it should be bumped up to `010_` instead of `009_`.

michaelsembwever:
> Here it should be bumped up to 010_ instead of 009_.

Yes, i'm on it. Presuming i'm free to merge otherwise.

adejanovski:
you are :)

PR #258, opened 2017-10-30
Split 'database' configuration setting

ossarga:
The changes in this Pull Request fix the naming consistency of the `storageType` setting and the associated JSON objects that are parsed. Specifically, the generic term `database` has been replaced with the specific database technologies that are supported; `h2` and `postgres`. This brings the storage naming in align with the other backend options `cassandra` and `memory`.

- Split 'database' conguration setting into 'h2' and 'postgres'.
- Deprecated 'driverClass' setting.
- Documentation and configuration example updtes.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/ReaperApplication.java
+++ b/src/server/src/main/java/io/cassandrareaper/ReaperApplication.java
@@ -234,14 +234,15 @@ public final class ReaperApplication extends Application<ReaperApplicationConfig
       storage = new MemoryStorage();
     } else if ("cassandra".equalsIgnoreCase(config.getStorageType())) {
       storage = new CassandraStorage(config, environment);
-    } else if ("postgres".equalsIgnoreCase(config.getStorageType()) || "h2".equalsIgnoreCase(config.getStorageType())) {
+    } else if ("postgres".equalsIgnoreCase(config.getStorageType())
+        || "h2".equalsIgnoreCase(config.getStorageType())
+        || "database".equalsIgnoreCase(config.getStorageType())) {
       // create DBI instance
       final DBIFactory factory = new DBIFactory();
 
       // instanciate store
       storage = new PostgresStorage(factory.build(environment, config.getDataSourceFactory(), "postgresql"));
       initDatabase(config);
-
     } else {
       LOG.error("invalid storageType: {}", config.getStorageType());
       throw new ReaperException("invalid storage type: " + config.getStorageType());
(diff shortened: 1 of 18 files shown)


src/server/src/main/java/io/cassandrareaper/ReaperApplication.java
```diff
@@ -233,11 +234,15 @@ private IStorage initializeStorage(ReaperApplicationConfiguration config, Enviro
...
       // create DBI instance
-      DBI jdbi;
       final DBIFactory factory = new DBIFactory();
-      jdbi = factory.build(environment, config.getDataSourceFactory(), "postgresql");
+
+      DBI jdbi = factory.build(
+              environment,
+              "postgres".equalsIgnoreCase(config.getStorageType())
+                      ? config.getPostgresDataSourceFactory() : config.getH2DataSourceFactory(),
+              "postgresql");
```
michaelsembwever: Can the `DBI` and the `DataSourceFactory` variables be provided by IoC? (instead of inline conditional logic)

michaelsembwever:
@ossarga 
i see both circleci failing on this patch (for real reasons).

src/server/src/main/java/io/cassandrareaper/ReaperApplication.java
ossarga: Good point. Fixed.

michaelsembwever (approved):

michaelsembwever:
> i see both circleci failing on this patch (for real reasons).

travis is constantly failing now only on its first task in the reaper build. this does not appear to be related to this PR.

michaelsembwever:
> travis is constantly failing now only on its first task in the reaper build. this does not appear to be related to this PR.

all green now. really only waiting on @rustyrazorblade 's review now.

adejanovski (changes requested):
When we first discussed this with @rustyrazorblade, we said it would be good to deprecate `database` and allow `postgres` and `h2` as a first step. Here, we're breaking backward compatibility and all existing configurations will fail.

2 choices here : 
- We still allow `database` but issue a big fat warning in the logs
- We don't re-introduce `database` but issue a clear error message when `database` is in use, suggesting to change it to either `postgres` or `h2`.

I'm in favor of the first option to avoid breaking people's habit for a minor concern.

rustyrazorblade:
Option 1 is my preference as well.

adejanovski (approved):

adejanovski:
I've re-introduced the `database` value, with the previous way of configuring the db connection.

Both approaches are valid and the existing config files will still work here.

PR #259, opened 2017-10-30
logging to rolled over files

michaelsembwever:
Instead of logging to the console, log to files that are rolled over daily.

### DIFF
--- a/src/packaging/resource/cassandra-reaper-cassandra-ssl.yaml
+++ b/src/packaging/resource/cassandra-reaper-cassandra-ssl.yaml
@@ -26,6 +26,7 @@ useAddressTranslator: false
 datacenterAvailability: ALL
 
 
+# any cassandra nodes on 127.0.0.x addresses we presume are ccm nodes
 jmxPorts:
   127.0.0.1: 7198
   127.0.0.2: 7200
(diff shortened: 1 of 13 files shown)


src/packaging/resource/cassandra-reaper-cassandra-ssl.yaml
```diff
@@ -46,8 +47,11 @@ logging:
     io.dropwizard: WARN
     org.eclipse.jetty: WARN
   appenders:
-    - type: console
-      logFormat: "%-6level [%d] [%t] %logger{5} - %msg %n"
+    - type: file
+      logFormat: "%-6level [%t] %logger{5} - %msg %n"
+      currentLogFilename: reaper.log
+      archivedLogFilenamePattern: reaper-%d.log.gz
+      archivedFileCount: 99
```
adejanovski: Shouldn't we specify a maximum size to control when rolling over will occur ?
Isn't 99 kind of a big number ?

michaelsembwever: this rollover is daily.

when using `%d` it's daily, when using `%i` and maxFileSize it becomes size limited.

i just went for the simpler daily rotation approach here.

adejanovski (dismissed):
I'm not crazy about changing the default behavior, which was to output the logs in the console.
I'd be in favor of adding the file appender commented for now or at least to keep the console appender too.

michaelsembwever:
> I'd be in favor of adding the file appender commented for now or at least to keep the console appender too.

@adejanovski 
the console appender is back, in addition to the file appender, but the console appender only emits the important stuff (WARN|ERROR|FATAL), so that whatever captures the console log also doesn't end up at 6Gb or more like we've witnessed. one small code change is added to ensure the application start message always appears.

adejanovski:
We should do so that packaged installs create the `reaper.log` files in `/var/log/cassandra-reaper` instead of creating them in the current directory. Other than that, I'm ok with the changes.

michaelsembwever:
> We should do so that packaged installs create the reaper.log files in /var/log/cassandra-reaper instead of creating them in the current directory.

Good point! I'll take a look into that…

PR #260, opened 2017-10-30
Fix issue with upgrades on postgres and h2 when coming from pre-0.8 

adejanovski:
This PR fixes #255 by handling null values in newly added arrays in the repair_schedule table.

PR #261, opened 2017-10-30
Fix invalid RepairSegment states stored in Cassandra / Postgres / H2

michaelsembwever:
This ensures segments read from Cassandra satisfy the preconditions in `RepairSegment.Builder.build()`

The Migration009 process could take some time on large clusters.


adejanovski (changes requested):
Fixing the Cassandra DB won't be enough as the problem also arises when upgrading with the Postgres backend (and most probably H2), as shown in PR #260

michaelsembwever:
> Fixing the Cassandra DB won't be enough as the problem also arises when upgrading with the Postgres backend (and most probably H2), as shown in PR #260

Doesn't #260 fix it for postgres and H2? While this PR fixes it just for Cassandra.

EDIT: it doesn't. #260 contains an additional comment describing a new fault.

adejanovski:
@anguenot : the PR is ready to be tested when you have time.

Thanks !

anguenot:
@adejanovski this PR definitely addresses the issue and the upgrade went through this time. Thanks!

adejanovski (approved):

PR #272, opened 2017-11-06
Don't log every jmx connection failure within JmxProxyImpl. The thrown exception is logged appropriately when caught

michaelsembwever:
Also reduces visibility of the not-for-public-use JmxConnectionFactory.connect(..) method.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/jmx/JmxConnectionFactory.java
+++ b/src/server/src/main/java/io/cassandrareaper/jmx/JmxConnectionFactory.java
@@ -54,8 +54,8 @@ public class JmxConnectionFactory {
     hostConnectionCounters = new HostConnectionCounters(metricRegistry);
   }
 
-  public JmxProxy connect(Optional<RepairStatusHandler> handler, String host, int connectionTimeout)
-      throws ReaperException, NumberFormatException, InterruptedException {
+  protected JmxProxy connect(Optional<RepairStatusHandler> handler, String host, int connectionTimeout)
+      throws ReaperException, InterruptedException {
     // use configured jmx port for host if provided
     if (jmxPorts != null && jmxPorts.containsKey(host) && !host.contains(":")) {
       host = host + ":" + jmxPorts.get(host);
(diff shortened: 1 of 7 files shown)


src/server/src/main/java/io/cassandrareaper/jmx/JmxConnectionFactory.java
```diff
@@ -100,7 +99,8 @@ public final JmxProxy connectAny(
           try {
             return connect(handler, host, connectionTimeout);
           } catch (ReaperException | RuntimeException e) {
-            LOG.info("Unreachable host", e);
+            LOG.info("Unreachable host: {}: {}", e.getMessage(), e.getCause().getMessage());
+            LOG.debug("Unreachable host: ", e);
```
michaelsembwever: this change simply keeps the informational stacktrace to debug, keeping only an informative message at the info level.

src/server/src/main/java/io/cassandrareaper/jmx/JmxProxyImpl.java
```diff
@@ -251,8 +251,7 @@ static JmxProxy connect(
 
       return proxy;
     } catch (IOException | ExecutionException | TimeoutException | InstanceNotFoundException e) {
-      LOG.error("Failed to establish JMX connection to {}:{}", host, port);
-      throw new ReaperException("Failure when establishing JMX connection", e);
+      throw new ReaperException("Failure when establishing JMX connection to " + host + ":" + port, e);
```
michaelsembwever: this is the core change of the PR: we don't need to log this exception under error, as it's caught elsewhere and logged appropriately…

adejanovski (approved):

michaelsembwever:
merged to `1.0` and `master` manually, using the following procedure…
```
git checkout 1.0
git merge  mck/jmx-less-exception-logging
git checkout master
git merge 1.0 -s ours
git cherry-pick --no-commit d75ba7b
git cia --amend
git push origin 1.0 master --atomic
```

(the `--atomic` approach isn't really needed, but i'm just re-using the C* approach)

PR #274, opened 2017-11-07
Fix documentation for DATACENTER_AWARE repair parallelism

adejanovski:
Previous description for DATACENTER_AWARE was inaccurate.


ossarga:
Yes, this wording makes more sense.

ossarga (approved):
This wording makes more sense.

PR #276, opened 2017-11-08
Fix for Elassandra's gossip info output parsing

### LINKED ISSUE #211: Failed to establish JMX connection across datacenters
I'm trying to track down why I'm getting consistent issues communicating with nodes outside the private network in which reaper is running.  The UI shows "Not available" for those datacenters, and the logs contain errors stating that reaper failed to establish a JMX connection.  And yet, I can telnet to the JMX port and even open jconsole to those nodes from the system reaper is on just fine (via an SSH tunnel).

It could be that reaper is attempting to communicate with them over the private network interface somehow.  These are EC2 instances, with datacenters in different regions, if that helps.  The public addresses show up in the log and UI, however.

I've tried using a reaper in each region/datacenter with datacenterAvailability on EACH, but that doesn't seem to have helped... each reaper still gets errors that it can't connect to nodes in other datacenters.

It's highly possible I'm just ignorant and am doing something wrong, but I've run out of leads that I can see.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-305) by [Unito](https://www.unito.io)
┆Issue Number: K8SSAND-305

adejanovski:
Fixes #211

The problem fixed isn't about JMX connections but is mentioned later in the comments. It's the cluster status view that is broken due to Elassandra adding stuff in the gossip info output.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/resources/view/NodesStatus.java
+++ b/src/server/src/main/java/io/cassandrareaper/resources/view/NodesStatus.java
@@ -85,7 +85,7 @@ public final class NodesStatus {
     Set<String> endpoints = Sets.newHashSet();
     Matcher matcher;
 
-    String[] strEndpoints = allEndpointStates.split("/");
+    String[] strEndpoints = allEndpointStates.split("(?<![0-9a-zA-Z ])/");
     Double totalLoad = 0.0;
 
     for (int i = 1; i < strEndpoints.length; i++) {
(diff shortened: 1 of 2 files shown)


adejanovski:
@adisharoon : any chance you could test the fix ?

Thanks

michaelsembwever (approved):
LGTM!

adisharoon:
@adejanovski The cluster view still looks all smooshed together for all regions but one for me with these changes, but I no longer see an erroneous "Not Available".  I played around with the CSS a little bit and was able to see what I expected, so I think it's just cosmetic.

PR #278, opened 2017-11-10
Docker Updates

ossarga:
- Updated Docker documentation to include commands to launch a Docker container without docker-compose.
- Fixed typos in the In-Memory documentation.
- Moved all the common services from the _common-services.yml_ into the _docker-compose.yml_ file.
- Added the `reaper-in-memory` service to docker-compose to launch a memory backed Reaper instance in a docker container.
- Removed `reaper-build-packages` from the main docker-compose file.
- Fixed broken Docker build command in Travis CI script.
- Changed links to use markdown instead of HTML.


ossarga:
This branch needs to be squashed (and possibly rebased) prior to merging.

src/docs/content/docs/download/docker.md
```diff
@@ -13,28 +13,99 @@ parent = "download"
...
+docker-compose build
+docker-compose run build
 ```
 
 ## Building Reaper Docker Image
 
-To build the Reaper Docker Image which is then added to the local image cache using the `cassandra-reaper:latest` tag, run the following commands from the top level directory:
+### Prerequisite
+
+The generation of the Docker image requires that the JAR file be built and placed in the _src/packages_ directory. If the JAR package is missing from the directory then it can built using either the steps in the [Docker package build](building-reaper-packages-with-docker) section (above), or in the <a href="../building#building-jars-from-source">Building from Source</a> section.
```
joaquincasares: @ossarga can you do a proper Markdown link rather than href link?

joaquincasares (commented):
Everything else lgtm

src/docs/content/docs/download/docker.md
ossarga: Unsure how to link across to a different section in the Markdown.

joaquincasares: [Building from Source](../building#building-jars-from-source) should work.

```
[Building from Source](../building#building-jars-from-source) should work.
```

ossarga: Hmmm ok. I probably specified it wrong when I was trying it. I have changed the link to use the markdown recommended in the Hugo docs
```
[Building from Source]({{< relref "building.md#building-jars-from-source" >}})
```

Issue #281, opened 2017-11-21
if endTime is set, so must startTime be set


adisharoon:
For some of our cassandra reaper instances, the web UI has stopped displaying running repairs, and the javascript console shows 500 errors.  The only clue I've found is the following exception which occurs frequently in the log.  This is for the new version 1.0.0  
```
ERROR [2017-11-21 16:42:42,853] io.dropwizard.jersey.errors.LoggingExceptionMapper: Error handling a request: 61b8c4120d508373
! java.lang.IllegalStateException: if endTime is set, so must startTime be set
! at com.google.common.base.Preconditions.checkState(Preconditions.java:174)
! at io.cassandrareaper.core.RepairSegment$Builder.build(RepairSegment.java:182)
! at io.cassandrareaper.storage.CassandraStorage.createRepairSegmentFromRow(CassandraStorage.java:693)
! at io.cassandrareaper.storage.CassandraStorage.getRepairSegmentsForRun(CassandraStorage.java:663)
! at io.cassandrareaper.storage.CassandraStorage.getClusterRunStatuses(CassandraStorage.java:935)
! at io.cassandrareaper.resources.ClusterResource.viewCluster(ClusterResource.java:127)
! at io.cassandrareaper.resources.ClusterResource.getCluster(ClusterResource.java:112)
! at sun.reflect.GeneratedMethodAccessor25.invoke(Unknown Source)
! at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
! at java.lang.reflect.Method.invoke(Method.java:498)
! at org.glassfish.jersey.server.model.internal.ResourceMethodInvocationHandlerFactory$1.invoke(ResourceMethodInvocationHandlerFactory.java:81)
! at org.glassfish.jersey.server.model.internal.AbstractJavaResourceMethodDispatcher$1.run(AbstractJavaResourceMethodDispatcher.java:144)
! at org.glassfish.jersey.server.model.internal.AbstractJavaResourceMethodDispatcher.invoke(AbstractJavaResourceMethodDispatcher.java:161)
! at org.glassfish.jersey.server.model.internal.JavaResourceMethodDispatcherProvider$ResponseOutInvoker.doDispatch(JavaResourceMethodDispatcherProvider.java:160)
! at org.glassfish.jersey.server.model.internal.AbstractJavaResourceMethodDispatcher.dispatch(AbstractJavaResourceMethodDispatcher.java:99)
[...]

PR #282, opened 2017-11-23
fix repair segment starttime

michaelsembwever:
addresses https://github.com/thelastpickle/cassandra-reaper/issues/281


michaelsembwever:
An additional commit 5a14d92c3ae840383505f91f5187dcdb71c65e3f was added to fix existing databases that have been corrupted in this way.

PR #286, opened 2017-11-28
RPM packaging improvements

internetstaff:
Protect the config file from overwrite and change user creation to avoid error output from 'id'.

Fixes issue 89.


adejanovski:
Thanks a lot for the PR !

Merged

PR #287, opened 2017-11-29
Prefer absolute URLs in webui (html)

mcoolive:
Today, the webui is not correctly displayed when we use an URL
with no ending slash. It is because JS links are relative.
- works:         http://server:port/webui/
- does not work: http://server:port/webui
I want to make both URLs work.
In this patch, I suggest to hardcode the base URL for all relative URLs
Alternatively, we could replace relative URL by absolute ones
In my opinion, we should use JSP (${pageContext.request.contextPath})
to not hardcode '/webui/' twice (it is already hardcoded in Java), but
I do not want to make a so big change (it can be a further improvment)

### DIFF
--- a/src/server/src/main/resources/assets/index.html
+++ b/src/server/src/main/resources/assets/index.html
@@ -11,6 +11,8 @@
 
     <title>Cassandra Reaper - Clusters</title>
 
+    <base href="/webui/">
+
     <!-- HTML5 Shim and Respond.js IE8 support of HTML5 elements and media queries -->
     <!-- WARNING: Respond.js doesn't work if you view the page via file:// -->
     <!--[if lt IE 9]>
--- a/src/server/src/main/resources/assets/repair.html
+++ b/src/server/src/main/resources/assets/repair.html
@@ -11,6 +11,8 @@
 
     <title>Cassandra Reaper - Repair</title>
 
+    <base href="/webui/">
+
     <!-- HTML5 Shim and Respond.js IE8 support of HTML5 elements and media queries -->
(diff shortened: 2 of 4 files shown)


mcoolive:
Ok. I did not see these files was generated. So, I patched html_template.ejs and the generated HTML files. But I did not commit changes on JS.
**Should I?**

I do a quick test: webui seams work, but the repair does not progress... I don't think it's related to my change. The reaper server prints this error message just after a click on 'activate' button: 

>    ERROR  [2017-12-01 12:25:12,165] [cstars01:43a335d0-d68a-11e7-b2cb-5bd1ba2b292c:43a335d1-d68a-11e7-b2cb-5bd1ba2b292c] i.c.s.RepairRunner - Executing SegmentRunner failed 
>    java.lang.IllegalStateException: startTime must be set if segment is RUNNING or DONE
>        at com.google.common.base.Preconditions.checkState(Preconditions.java:174)
>        at io.cassandrareaper.core.RepairSegment$Builder.build(RepairSegment.java:185)

mcoolive:
So I tested again. It works well if I cherry-pick my change on eb2fc0c (Bump version number to 1.0.1).
With the HEAD version, I got an issue (startTime must be set if segment is RUNNING or DONE) just after I click on "activate" button. It seams relative to the commit ed418ce (Remove precondition checks that break Reaper when endTime/startTime ) that was reverted after the release. So I assume it's a known issue, and I am very confident with my own change.

adejanovski:
Thanks for the PR @mcoolive.

Merging

PR #289, opened 2017-11-30
Stop throwing errors on a keyspace with RF=1

internetstaff:
There may be a more elegant way you want to handle this scenario (e.g. skip the entire repair - I don't understand the commandId=0 scenarios well enough), but this worked for me. :)


adejanovski:
This was handled as part of https://github.com/thelastpickle/cassandra-reaper/pull/317. 

Thanks a lot for sending that PR and sorry for the time it took us to process it.

Issue #291, opened 2017-12-05
if endTime is set, so must startTime be set (v1.0.1)


ludekdolejsky:
I am unfortunately seeing issue similar to #281 even after upgrading to 1.0.1.

I am running one Reaper per DC and the `endTime is set, so must startTime be set` error is happening occasionally (I'd say less frequently than prior to upgrading to 1.0.1).

please see attached full log files from both DCs:
[logs.zip](https://github.com/thelastpickle/cassandra-reaper/files/1530109/logs.zip)

data causing the issue (you can see they come from dc2 log file):
```
cqlsh> select segment_start_time, segment_end_time from reaper_db.repair_run where id=231ee1e0-d871-11e7-8db9-7bb3e751dc82 and segment_id=23212c16-d871-11e7-8db9-7bb3e751dc82;

 segment_start_time | segment_end_time
--------------------+---------------------------------
               null | 2017-12-04 00:36:45.261000+0000

(1 rows)
```

everything goes back to normal when I fix the data manually by running:
```
update reaper_db.repair_run set segment_end_time=null where id=231ee1e0-d871-11e7-8db9-7bb3e751dc82 and segment_id=23212c16-d871-11e7-8db9-7bb3e751dc82;
```

adejanovski:
@ludekdolejsky : It appears that you ran into a race condition where the `SUCCESS` JMX notification was processed after the `COMPLETE` one.
Release 1.0.2 hides the issue and allows not to break Reaper in case that happens.

We're working on a proper fix that should be available shortly.

ludekdolejsky:
Thanks - will try 1.0.2 out!

vdemonchy:
Hello, just wanted to let you know that we encountered the issue while using Reaper in my company in a version <= 1.0.1 but no longer in 1.0.2, and even if I'm aware that it's a temporary workaround until it's properly fixed, thank you !

michaelsembwever:
The precondition checks are back in place (and will be released in 1.0.3).
If people hit this problem again the checks can be explicitly disabled by starting reaper with the jvm system property `-Dreaper.disableSegmentChecks=true`.

This shouldn't be necessary though as the underlying bug (caused by d6e5a10af05382b25a9621b5f0dcd6d0c5f2ef56) has been fixed.

sergeyyegournov:
Probably there is another path how to face another precondition failure:

```
[...]

PR #297, opened 2017-12-11
Fix issue #291 with out of order JMX notifications

adejanovski:
Instead of allowing to write segments with inconsistent startTime, endTime and state, we now reset any segment with that kind issue so that it gets reprocessed later.
This allows us to remove the precondition checks in the RepairSegment builder.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/core/RepairSegment.java
+++ b/src/server/src/main/java/io/cassandrareaper/core/RepairSegment.java
@@ -23,8 +23,11 @@ import javax.annotation.Nullable;
 
 import com.google.common.base.Preconditions;
 import org.joda.time.DateTime;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 public final class RepairSegment {
+  private static final Logger LOG = LoggerFactory.getLogger(RepairSegment.class);
 
   private final UUID id;
   private final UUID runId;
@@ -109,6 +112,24 @@ public final class RepairSegment {
     DONE
   }
 
+  public boolean isValid() {
+    if (null == startTime && null != endTime) {
+      LOG.debug("Invalid repair segment {} : startTime is null but endTime is not", this.id);
+      return false;
+    }
+    if (null != endTime && State.DONE != state) {
+      LOG.debug("Invalid repair segment {} : endTime is not null but state is not DONE", this.id);
+      return false;
+    }
+    if (null == startTime && State.NOT_STARTED != state) {
+      LOG.debug(
+          "Invalid repair segment {} : startTime is null but state is not NOT_STARTED", this.id);
+      return false;
+    }
+
+    return true;
+  }
+
   public static final class Builder {
 
     public final RingRange tokenRange;
@@ -171,7 +192,7 @@ public final class RepairSegment {
       return this;
     }
 
-    public Builder endTime(DateTime endTime) {
+    public Builder endTime(@Nullable DateTime endTime) {
       this.endTime = endTime;
       return this;
     }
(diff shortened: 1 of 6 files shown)


src/server/src/main/java/io/cassandrareaper/core/RepairSegment.java
```diff
@@ -172,29 +172,26 @@ public Builder startTime(@Nullable DateTime startTime) {
...
-      "startTime must be set if segment is RUNNING or DONE"); */
-
-      if (null != endTime && State.DONE != state) {
-        endTime = null;
-      }
-
-      if (startTime == null && endTime != null) {
-        startTime = endTime.minusSeconds(1);
-      }
-
```
michaelsembwever, line 196: I would want to put the preconditions back in place now.
They are there to check our design, and prevent regressions.
I see the clash with how this PR is coded though… need some further thought on this…

src/server/src/main/java/io/cassandrareaper/core/RepairSegment.java
```diff
@@ -172,29 +172,26 @@ public Builder startTime(@Nullable DateTime startTime) {
     }
 
     public Builder endTime(DateTime endTime) {
-      Preconditions.checkNotNull(endTime);
```
michaelsembwever: needs `@Nullable` annotation

src/server/src/main/java/io/cassandrareaper/core/RepairSegment.java
```diff
@@ -172,29 +172,26 @@ public Builder startTime(@Nullable DateTime startTime) {
...
-
-      if (startTime == null && endTime != null) {
-        startTime = endTime.minusSeconds(1);
-      }
-
       return new RepairSegment(this, segmentId);
     }
   }
+
+  public boolean isConsistentOnTimesAndState() {
```
michaelsembwever: what about the method name `isValid()`

src/server/src/test/java/io/cassandrareaper/service/SegmentRunnerTest.java
```diff
@@ -277,6 +274,141 @@ protected JmxProxy connect(final Optional<RepairStatusHandler> handler, String h
     assertEquals(0, storage.getRepairSegment(runId, segmentId).get().getFailCount());
   }
 
+  @Test
+  public void disorderedNotificationsTest()
```
michaelsembwever, line 278: this is failing a lot in travis tests.

for example:
 - https://travis-ci.org/thelastpickle/cassandra-reaper/jobs/314826143
 - https://travis-ci.org/thelastpickle/cassandra-reaper/jobs/314825786

In CircleCI it was a different test failure in `RepairRunnerTest.testHangingRepair`, ref: https://circleci.com/gh/thelastpickle/cassandra-reaper/710

michaelsembwever (changes requested):

src/server/src/main/java/io/cassandrareaper/core/RepairSegment.java
michaelsembwever: i took a stab at this, looking for a way to do the same thing with less code and came up with the following…
 https://github.com/thelastpickle/cassandra-reaper/commit/45e67f2aadad55baa937f8fbe7ae4aea9aefac9a

this is a smaller patch, although CassandraStorage still needs a way to re-null the endTime, and there's probably some other small things required, but it demonstrates i think a simpler way.

that said: having coded up an alternative; i've become less at ease with the approach.
my initial feeling here is that we're making the code more complicated, rather than directly solving the problem.

i'd like to put together a PR that a) directly solves the problem and b) provides a way to disable the precondition checks via the use of a system property. hope this sounds ok, then we can compare approaches and re-evaluate…

adejanovski:
This was fixed in another PR, closing.

PR #300, opened 2017-12-13
issue 291  start time end time bug

### LINKED ISSUE #291: if endTime is set, so must startTime be set (v1.0.1)

michaelsembwever:
Fix #291 if endTime is set, so must startTime be set (v1.0.1) 

Commit d6e5a10 created the fault.
When the `updateRepairSegmentPrepStmt` statement was split into two, creating `insertRepairSegmentEndTimePrepStmt`,
it meant that calls to `SegmentRunner.postponse(..)` were no longer nulling endTime when resetting the RepairSegment to NOT_STARTED state.

The fix here is to

 - use `insertRepairSegmentEndTimePrepStmt` to null the endTime if state is NOT_STARTED (and live with tombstones created),
 - add method `RepairSegment.reset()` for the explicit purpose of resetting a segment to NOT_STARTED state with null startTime and endTime,
 - not permit nulls passed into `RepairSegment.Builder.startTime(..)`
 - add migration scripts (one more time) to fix any nulls in storage

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/core/RepairSegment.java
+++ b/src/server/src/main/java/io/cassandrareaper/core/RepairSegment.java
@@ -104,6 +104,23 @@ public final class RepairSegment {
     return new Builder(this);
   }
 
+  public boolean hasStartTime() {
+    return null != startTime;
+  }
+
+  public boolean hasEndTime() {
+    return null != endTime;
+  }
+
+  /** Reset to NOT_STARTED state, with nulled startTime and endTime. */
+  public Builder reset() {
+    Builder builder = new Builder(this);
+    builder.state = State.NOT_STARTED;
+    builder.startTime = null;
+    builder.endTime = null;
+    return builder;
+  }
+
   public enum State {
     NOT_STARTED,
     RUNNING,
@@ -163,7 +180,7 @@ public final class RepairSegment {
       return this;
     }
 
-    public Builder startTime(@Nullable DateTime startTime) {
+    public Builder startTime(DateTime startTime) {
       Preconditions.checkState(
           null != startTime || null == endTime,
           "unsetting startTime only permitted if endTime unset");
--- a/src/server/src/main/java/io/cassandrareaper/core/RepairSegment.java
+++ b/src/server/src/main/java/io/cassandrareaper/core/RepairSegment.java
@@ -18,13 +18,14 @@ import io.cassandrareaper.service.RingRange;
 
 import java.math.BigInteger;
 import java.util.UUID;
-
 import javax.annotation.Nullable;
 
 import com.google.common.base.Preconditions;
 import org.joda.time.DateTime;
 
 public final class RepairSegment {
+
+  private static final boolean STRICT = !Boolean.getBoolean("reaper.disableSegmentChecks");
 
   private final UUID id;
   private final UUID runId;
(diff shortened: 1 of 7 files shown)


src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
```diff
@@ -630,14 +630,16 @@ assert hasLeadOnSegment(segment.getId())
             null != segment.getStartTime() ? segment.getStartTime().toDate() : null,
             segment.getFailCount()));
 
-    if (null != segment.getEndTime()) {
-      assert RepairSegment.State.DONE == segment.getState();
+    if (null != segment.getEndTime() || State.NOT_STARTED == segment.getState()) {
+      assert RepairSegment.State.RUNNING != segment.getState();
+      assert RepairSegment.State.NOT_STARTED != segment.getState() || null == segment.getEndTime();
+      assert RepairSegment.State.DONE != segment.getState() || null != segment.getEndTime();
```
adejanovski: I find these tests hard to read because of the negations and the `||` operator. It's not obvious what is tested unless you give it a good thought.

Could we trade these against individual methods in the RepairSegment object ?
Something like : 
```
public boolean isNotStarted() {
  return RepairSegment.State.NOT_STARTED == getState();
}

public boolean isRunning() {
  return RepairSegment.State.RUNNING == getState();
}

public boolean isDone() {
  return RepairSegment.State.RUNNING != getState();
}


public boolean hasEndTime() {
  return null != getEndTime();
}
...
```
Then we would use those methods combined together in the assertions.
It wouldn't change much but the code intent would be easier to understand.

src/server/src/main/java/io/cassandrareaper/core/RepairSegment.java
```diff
@@ -180,18 +190,20 @@ public Builder endTime(DateTime endTime) {
...
-      null != startTime || State.NOT_STARTED == state,
-      "startTime must be set if segment is RUNNING or DONE"); */
-
-      if (null != endTime && State.DONE != state) {
-        endTime = null;
-      }
-
-      if (startTime == null && endTime != null) {
-        startTime = endTime.minusSeconds(1);
+      if (STRICT) {
```
adejanovski, line 201: I like the idea that the checks can be disabled.

src/server/src/main/java/io/cassandrareaper/core/RepairSegment.java
```diff
@@ -103,6 +104,15 @@ public Builder with() {
     return new Builder(this);
   }
 
+  /** Reset to NOT_STARTED state, with nulled startTime and endTime. */
+  public Builder reset() {
```
adejanovski: That's better than what I came up with. Nice !

adejanovski (approved):
Aside from minor possible code changes to make it easier to read, LGTM

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
michaelsembwever: yeah fair point. i'll add those methods.
i'll also add assert messages, as this helps readability too…

michaelsembwever: done. and changed the `assert` statements to `Preconditions.checkArgument(..)` as there's no performance overhead on these checks.

michaelsembwever:
I'm happy with merging this. But will wait til I touch base with @adejanovski tonight.

PR #302, opened 2017-12-15
Add ability to list and abort segments

### LINKED ISSUE #299: Ability to list segments
Reaper doesn't currently allow to track segments, just the overall progress.
It would be good to have the ability to display all segments for a specific repair, and now how many segments are currently being processed for each run.

adejanovski:
Fixes #299 

The PR adds a REST endpoint to list the segments for a given repair run id, and another one to abort/reset a segment.
If the segment is running, it will be aborted through JMX and postponed for future reprocessing.
If the segment was done already, it will be reset and Reaper will reprocess it.

The spreaper command line tool has been updated to support both operations (list-segments and abort-segment), and the UI now shows a "View segments" button on the Repairs screen : 
reaper_view_segments_run

Clicking it will open a modal window that lists the segments, grouped by state : 
reaper_segment_list

The abort feature will allow to stop stuck segments that have been reported to make some repairs hang, until we can develop a segment auditor to handle this automatically.


src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
```diff
@@ -466,6 +466,53 @@ public Response getRepairRun(
...
+
+    LOG.debug("abort segment called with: run id = {} and segment id = {}", repairRunId, segmentId);
+    final Optional<RepairRun> repairRun = context.storage.getRepairRun(repairRunId);
+    if (repairRun.isPresent()) {
+      if (RepairRun.RunState.RUNNING == repairRun.get().getRunState()
+          || RepairRun.RunState.PAUSED == repairRun.get().getRunState()) {
+        RepairSegment segment = context.repairManager.abortSegment(repairRunId, segmentId);
+        return Response.ok().entity(segment).build();
+      } else {
+        return Response.status(Response.Status.FORBIDDEN)
```
michaelsembwever: I don't think forbidden is the correct response here.
This is not a problem with client authentication/authorization.

The request is illegal due to state.
405 (method not allowed) or 409 (conflict) would seem more appropriate. my preference would be would 409.

adejanovski:
@michaelsembwever, I've made the requested minor changes.

That branch has been living long enough for rebase to get tedious now, as it conflicts with at least two other merged PRs.
Let me know if we can merge this ASAP.

Thanks !

rustyrazorblade:
+1, ship it

Issue #301, opened 2017-12-15
Opportunity to change intensity during execution


kipwoker:
Hello.

I want to change intensity during execution.
I have a highload cassandra cluster during 8 working hours every weekday.
During working hours I want to have small intensity, like 0.2.
During maintenance window (night, day off) I want to have intesity like 0.8.

Am I right that there is no way to do it with [autoScheduling](http://cassandra-reaper.io/docs/configuration/reaper_specific/)?
Should I make a pull request for this feature?

adejanovski:
Hi @kipwoker, 

I confirm that intensity is a fixed value for each repair run and cannot currently vary based on the current time.
Any PR to add that feature would be much appreciated indeed :)

How far do you see this going in flexibility ? Would you add just 2 possible intensities with 2 time ranges for each repair ? Or add a high load and a low load time range on the clusters, and then 2 distinct intensities for repair run ?

kipwoker:
@adejanovski I want to do it very flexible. :)
Actually I just want to implement feature: «Change intesity when repair paused»
After that I can do whatever I want though using api + cron. It's very flexible.
Also for web interface I see it like this
![intensity](https://user-images.githubusercontent.com/1760443/34038162-6f139ada-e19c-11e7-8529-fdb4967e608b.PNG)
1. If pressed «Pause» button, show «Edit» button.
2. If pressed «Edit» button, show settings which can be changed (for now only intensity)
3. If field with value blurred, send ajax query to server with new value
4. If activate button pressed again, repair run again but with new settings

What do you mind about this?

kipwoker:
I've tried to make something. You may check it out: https://github.com/kipwoker/cassandra-reaper/commit/bdb3947aa6734bc998113d5fcc0171d0eec4efd6

adejanovski:
Merged as part of #303

Issue #163, opened 2017-08-25
Untangle project structure


rustyrazorblade:

michaelsembwever:
i've been eyeing this off. i've no opinion on using `.spotify.`, it is a compliment and recognition, but i see the need to clean up the packaging structure. there's class dependency tangles, some not very intuitive placement, and "layer" packages instead of "feature" packages.

http://www.javapractices.com/topic/TopicAction.do?Id=205

michaelsembwever:
The following diagram (from structure101) shows that the whole project is tangled but ReaperException and utils. (and utils is a layer package we don't want)

<img width="418" alt="screen shot 2017-08-26 at 09 58 55" src="https://user-images.githubusercontent.com/559444/29736551-969b69b0-8a45-11e7-8cc7-2ffca7da729b.png">

michaelsembwever:
This is the diagram from the code when it was at spotify…
<img width="596" alt="screen shot 2017-08-27 at 11 07 13" src="https://user-images.githubusercontent.com/559444/29746276-4295c99a-8b18-11e7-97b5-24f14226a0f9.png">

(you can see that it's basically the same, the `ft-reaper-improvements-final` branch did though introduce extra tangles. tangles, or circular dependencies are represented by the red arrows.)

michaelsembwever:
I took a quick [experiment](https://github.com/thelastpickle/cassandra-reaper/tree/mck/refactor-packages-by-feature-0) at refactoring to package-by-feature, so that packages look like:

<img width="315" alt="screen shot 2017-08-27 at 10 53 48" src="https://user-images.githubusercontent.com/559444/29746302-c2bcda32-8b18-11e7-9df6-0c00e9a77a8e.png">

This gave the following tangles:

<img width="513" alt="screen shot 2017-08-27 at 10 51 57" src="https://user-images.githubusercontent.com/559444/29746334-d00d0f44-8b19-11e7-98c1-caf31eba54bc.png">

This starts to look better, but still shows that there's a far bit of untangling to do.
In the [refactor-packages-by-feature-0](https://github.com/thelastpickle/cassandra-reaper/tree/mck/refactor-packages-by-feature-0) branch the commit message is keeping a list of easy-wins to untangles.

michaelsembwever:
> rustyrazorblade changed the title from switch packages structure to com.thelastpickle to Untangle project structure 2 hours ago

Thanks for the ticket rename. My apologies for hijacking the ticket.

[...]

PR #303, opened 2017-12-18
Change intensity during execution

kipwoker:
Pull request for https://github.com/thelastpickle/cassandra-reaper/issues/301

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
+++ b/src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
@@ -17,10 +17,12 @@ package io.cassandrareaper.resources;
 import io.cassandrareaper.AppContext;
 import io.cassandrareaper.ReaperException;
 import io.cassandrareaper.core.Cluster;
+import io.cassandrareaper.core.Intensity;
 import io.cassandrareaper.core.RepairRun;
 import io.cassandrareaper.core.RepairRun.RunState;
 import io.cassandrareaper.core.RepairSegment;
 import io.cassandrareaper.core.RepairUnit;
+import io.cassandrareaper.core.Result;
 import io.cassandrareaper.resources.view.RepairRunStatus;
 import io.cassandrareaper.service.RepairRunService;
 import io.cassandrareaper.service.RepairUnitService;
@@ -281,21 +283,17 @@ public final class RepairRunResource {
         return Response.status(Response.Status.BAD_REQUEST).entity(ex.getMessage()).build();
       }
     }
+
     if (intensityStr.isPresent()) {
-      try {
-        final Double intensity = Double.parseDouble(intensityStr.get());
-        if (intensity <= 0.0 || intensity > 1.0) {
-          return Response.status(Response.Status.BAD_REQUEST)
-              .entity("query parameter \"intensity\" must be in half closed range (0.0, 1.0]: " + intensityStr.get())
-              .build();
-        }
-      } catch (NumberFormatException ex) {
-        LOG.error(ex.getMessage(), ex);
-        return Response.status(Response.Status.BAD_REQUEST)
-            .entity("invalid value for query parameter \"intensity\": " + intensityStr.get())
+      final Result<Intensity, String> intensityCreateResult = Intensity.tryCreate(intensityStr.get());
+      if (intensityCreateResult.isFailed()) {
+        return Response
+            .status(Response.Status.BAD_REQUEST)
+            .entity(intensityCreateResult.getFailure())
             .build();
       }
     }
+
     if (incrementalRepairStr.isPresent()
         && (!incrementalRepairStr.get().toUpperCase().contentEquals("TRUE")
         && !incrementalRepairStr.get().toUpperCase().contentEquals("FALSE"))) {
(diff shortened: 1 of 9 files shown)


michaelsembwever:
@kipwoker  what's the first "init" commit about?

src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
```diff
@@ -320,6 +311,25 @@ static Response checkRequestForAddRepair(
     return null;
   }
 
+  private static Response validateIntensity(Optional<String> intensityStr) {
```
michaelsembwever: needs `@Nullable` if method can return `null`.

src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
```diff
@@ -334,7 +344,8 @@ static Response checkRequestForAddRepair(
   public Response modifyRunState(
       @Context UriInfo uriInfo,
       @PathParam("id") UUID repairRunId,
-      @QueryParam("state") Optional<String> state)
+      @QueryParam("state") Optional<String> state,
+      @QueryParam("intensity") Optional<String> intesityStr)
```
michaelsembwever: typo in parameter name…

src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
```diff
@@ -343,22 +354,24 @@ public Response modifyRunState(
       return Response.status(Response.Status.BAD_REQUEST.getStatusCode()).entity("\"state\" argument missing").build();
     }
 
-    final Optional<RepairRun> repairRun = context.storage.getRepairRun(repairRunId);
-    if (!repairRun.isPresent()) {
+    final Optional<RepairRun> maybeRepairRun = context.storage.getRepairRun(repairRunId);
```
michaelsembwever: not too sure why you're introducing and renaming `maybeRepairRun`… (or `maybeRepairUnit` below)
these touches feel superfluous to the patch's objective…

src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
```diff
@@ -379,22 +392,33 @@ public Response modifyRunState(
           .entity("invalid \"state\" argument: " + state.get())
           .build();
     }
-    final RepairRun.RunState oldState = repairRun.get().getRunState();
+    final RepairRun.RunState oldState = repairRun.getRunState();
+
+    if (intesityStr.isPresent() && oldState == newState && newState == RepairRun.RunState.PAUSED) {
```
michaelsembwever: is `intesityStr.isPresent()` required in the conditional…?

src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
```diff
@@ -448,6 +472,14 @@ private Response abortRun(RepairRun repairRun, RepairUnit repairUnit, int segmen
     return Response.ok().entity(new RepairRunStatus(newRun, repairUnit, segmentsRepaired)).build();
   }
 
+  private Response editRun(
+          RepairRun repairRun, RepairUnit repairUnit, int segmentsRepaired, Double intensity
+  ) throws ReaperException {
```
michaelsembwever: unusual line wrapping of method declaration.

src/server/src/main/java/io/cassandrareaper/service/RepairManager.java
```diff
@@ -188,7 +188,8 @@ public RepairRun startRepairRun(RepairRun runToBeStarted) throws ReaperException
         return updatedRun;
       }
       case PAUSED: {
-        RepairRun updatedRun = runToBeStarted.with()
+        RepairRun updatedRun = runToBeStarted
+            .with()
```
michaelsembwever: superfluous changes to the patch. please remove.

michaelsembwever (changes requested):
a few small changes requested, and one question around appropriate and clear restful response codes and messages…

kipwoker:
@michaelsembwever 

> what's the first "init" commit about?

It's just a habit when I start new branch.

src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
kipwoker: It's just a DRY implementation.
There were a lot of places with `repairRun.get()` below `!repairRun.isPresent()` check. In my opinion it's clearly to read.
May you explain what's wrong?

src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
kipwoker: Yes, because you can use REST API. For example, make two PUT calls with PAUSED state and with no intensity. For this case response should be 301 Not Modified, right?
This condition means that "you can set intensity but it will be used in case when repair run was paused and you don't change state in this during this request".

src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
kipwoker: Linter throwed error about line size. Suggestions?

src/server/src/main/java/io/cassandrareaper/service/RepairManager.java
kipwoker: Why? I just made it readable and similar to NOT_STARTED case. Look at 180 line.

src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
```diff
@@ -379,22 +392,33 @@ public Response modifyRunState(
...
+    final RepairRun.RunState oldState = repairRun.getRunState();
+
+    if (intesityStr.isPresent() && oldState == newState && newState == RepairRun.RunState.PAUSED) {
+      final Response intensityValidationResponse = validateIntensity(intesityStr);
+      if (intensityValidationResponse != null) {
+        return intensityValidationResponse;
+      }
+
+      final double intensity = Double.parseDouble(intesityStr.get());
+      return editRun(repairRun, repairUnit, segmentsRepaired, intensity);
```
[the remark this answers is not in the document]
kipwoker: `intensity` parameter will be ignored in case when old state and new state is not PAUSED.
So for next two requests
`PUT state=RUNNING&intensity=null`
`PUT state=RUNNING&intensity=0.7`
response will be `Not Modified`
Just like intensity parameter doesn't exist.
For this case can be more aggresive logic like `Bad Request`. But I don't understand profit of this logic.

src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
kipwoker: Fixed.

src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
kipwoker: Fixed.

michaelsembwever:
> > what's the first "init" commit about?
> It's just a habit when I start new branch.

Can you get rid of it please :-)

src/server/src/main/java/io/cassandrareaper/service/RepairManager.java
michaelsembwever: it's not related to the PR, and the statement on this line isn't otherwise touched in the PR.
if you'd like to offer code style/formatting improvements that should be done as a separate PR.

src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
michaelsembwever: fair enough.
then the project's precedence on this style would be the following
```
  private Response editRun(
          RepairRun repairRun,
          RepairUnit repairUnit,
          int segmentsRepaired,
          Double intensity) throws ReaperException {
```

src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
michaelsembwever: it's not related to the PR, and the statement on this line isn't otherwise touched in the PR.
if you'd like to offer code style/formatting improvements that should be done as a separate PR.

src/server/src/test/java/io/cassandrareaper/resources/RepairRunResourceTest.java
```diff
@@ -364,7 +364,7 @@ public void testPauseNotRunningRun() throws InterruptedException, ReaperExceptio
   public void testPauseNotExistingRun() throws InterruptedException, ReaperException {
     RepairRunResource resource = new RepairRunResource(context);
     Response response = resource.modifyRunState(uriInfo, UUIDs.timeBased(),
-        Optional.of(RepairRun.RunState.PAUSED.toString()));
+        Optional.of(RepairRun.RunState.PAUSED.toString()), Optional.absent());
     assertEquals(Response.Status.NOT_FOUND.getStatusCode(), response.getStatus());
     assertEquals(0, context.storage.getRepairRunsWithState(RepairRun.RunState.RUNNING).size());
   }
```
michaelsembwever: it would be great with some tests also directly against `RepairManager.editRepairRun(..)` (in RepairManagerTest)

src/server/src/main/java/io/cassandrareaper/service/RepairManager.java
```diff
@@ -218,6 +219,19 @@ public RepairRun startRepairRun(RepairRun runToBeStarted) throws ReaperException
     }
   }
 
+  public RepairRun editRepairRun(RepairRun repairRun, Double intensity) throws ReaperException {
```
michaelsembwever: looking through `RepairManager` the method names are more explicit in their actionable names.
so i'm thinking here that `updateRepairRunIntensity(..)` would be a better name.
what do you think?

src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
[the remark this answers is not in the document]
michaelsembwever: > > what's the response to the (restful) user when newState != RepairRun.RunState.PAUSED a change in intensity was specified but not fulfilled but the repair run was otherwise modified?
>
> intensity parameter will be ignored in case when old state and new state is not PAUSED.
> So for next two requests
> `PUT state=RUNNING&intensity=null`
> `PUT state=RUNNING&intensity=0.7`
> response will be `Not Modified`
> Just like intensity parameter doesn't exist.
> For this case can be more aggressive logic like `Bad Request`. But I don't understand profit of this logic.

I don't think we need more return codes or logic, eg `Bad Request`.
It's just to review that we arn't going to be returning any response status codes that confuse users…

So the logic is (in order of precedence): 
 
 # if request does not specify new state, then BAD_REQUEST is returned,
 # if the repair run (or unit) is not found, then NOT_FOUND is returned,
 # if there's any other repair runs for the related repair unit and they are RUNNING, then return CONFLICT,
 # if state is unaltered and PAUSED, intensity specified but invalid then BAD_REQUEST is returned,
 # if state is unaltered and PAUSED, intensity specified and valid then OK is returned,
 # if state is unaltered, all other parameters are ignored, and NOT_MODIFIED is returned,
 # if state is altered, but an invalid transition, all other parameters are ignored, and METHOD_NOT_ALLOWED returned,
 # if state is altered, parameters are honoured (intensity possibly changed), and OK returned.

The issue I see here is that there's no catch of when:
 
 - (in step 5) intensity is specified and valid but matches existing value. NOT_MODIFIED should be returned instead of OK.
 - (after step 5) when a state transition is specified as well as a new intensity but that new intensity is ignored.

What do you think about creating a separate REST endpoint for handling updates to intensity, so to simplify the above logic?
I'm open to ideas…

src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
```diff
@@ -379,22 +393,33 @@ public Response modifyRunState(
...
+    final RepairRun.RunState oldState = repairRun.getRunState();
+
+    if (intensityStr.isPresent() && oldState == newState && newState == RepairRun.RunState.PAUSED) {
+      final Response intensityValidationResponse = validateIntensity(intensityStr);
+      if (intensityValidationResponse != null) {
+        return intensityValidationResponse;
+      }
+
+      final double intensity = Double.parseDouble(intensityStr.get());
+      return editRun(repairRun, repairUnit, segmentsRepaired, intensity);
```
michaelsembwever: > > what's the response to the (restful) user when newState != RepairRun.RunState.PAUSED a change in intensity was specified but not fulfilled but the repair run was otherwise modified?
>
> intensity parameter will be ignored in case when old state and new state is not PAUSED.
> So for next two requests
> `PUT state=RUNNING&intensity=null`
> `PUT state=RUNNING&intensity=0.7`
> response will be `Not Modified`
> Just like intensity parameter doesn't exist.
> For this case can be more aggressive logic like `Bad Request`. But I don't understand profit of this logic.

I don't think we need more return codes or logic, eg `Bad Request`.
It's just to review that we arn't going to be returning any response status codes that confuse users…

So the logic is (in order of precedence): 
 
 1. if request does not specify new state, then BAD_REQUEST is returned,
 1. if the repair run (or unit) is not found, then NOT_FOUND is returned,
 1. if there's any other repair runs for the related repair unit and they are RUNNING, then return CONFLICT,
 1. if state is unaltered and PAUSED, intensity specified but invalid then BAD_REQUEST is returned,
 1. if state is unaltered and PAUSED, intensity specified and valid then OK is returned,
 1. if state is unaltered, all other parameters are ignored, and NOT_MODIFIED is returned,
 1. if state is altered, but an invalid transition, then all other parameters are ignored, and METHOD_NOT_ALLOWED returned,
 1. if state is altered, the transition valid, then all other parameters are ignored, and OK returned.

Does this sound right?
The issue I see here is that there's no catch of when:
 
 - (in step 5) intensity is specified and valid but matches existing value. NOT_MODIFIED should be returned instead of OK.
 - (after step 5) when a state transition is specified as well as a new intensity but that new intensity is ignored.

What do you think about creating a separate REST endpoint for handling updates to intensity, so to simplify the above logic?
I'm open to ideas…

src/server/src/main/java/io/cassandrareaper/service/RepairManager.java
kipwoker: I thought that this method can be extended in future by other parameters (not only intensity). That's why I named it in common case. According to open-closed principle you need only extend this method with new parameter if you want to edit something else. Without any renaming.

src/server/src/main/java/io/cassandrareaper/service/RepairManager.java
kipwoker: Why I need to make another PR for this trifle? I don't understand benefit. Can you explain?

src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
kipwoker: I thought about another endpoint but current logic means
PUT repair_run/{id} -- edit repair run with {id}
There is nothing about edit only state.
That's why I wrote code in this method.
In future you perhaps will want to extend edit (PUT) method by another parameters. Think about it. How would you like to do it?

If you want my opinion I think that this method is messy. Better to rewrite all method transitions to state machine implementation. So there will be just one PUT method but it will reroute call through finite state machine graph.

src/server/src/test/java/io/cassandrareaper/resources/RepairRunResourceTest.java
kipwoker: Yes. You're right. Will do it.

src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
kipwoker: Okay.

src/server/src/main/java/io/cassandrareaper/service/RepairManager.java
kipwoker: I mean, I can split it into several commits, isn't enough?

kipwoker:
By the way, why RepairRunResourceTest.java contains a lot of `Thread.sleep` usages? It works without it.

src/server/src/main/java/io/cassandrareaper/service/RepairManager.java
michaelsembwever: let's do without premature design for extension for now. 

separate methods for separate actions seems to make more sense in this class for now.

michaelsembwever:
> By the way, why RepairRunResourceTest.java contains a lot of Thread.sleep usages? It works without it.

Good question. Have you looked through the commit history for clues? Have you tried testing locally on a fully loaded/saturated machine? (travis/circleci behave quite differently to laptops).
Anyway… this is off-topic. If you find anything interesting it warrants a new issue/PR.

src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
```diff
@@ -320,6 +311,26 @@ static Response checkRequestForAddRepair(
...
+        }
+      } catch (NumberFormatException ex) {
+        LOG.error(ex.getMessage(), ex);
+        return Response.status(Response.Status.BAD_REQUEST)
+            .entity("invalid value for query parameter \"intensity\": " + intensityStr.get())
+            .build();
+      }
+    }
+    return null;
+  }
```
michaelsembwever: this is trivial… but what about 
```
   private static double parseIntensity(Optional<String> intensityStr) throws ValidationException {
       Preconditions.checkArgument(intensityStr.isPresent());
       try {
         final Double intensity = Double.parseDouble(intensityStr.get());
         if (intensity <= 0.0 || intensity > 1.0) {
           throw new ValidationException("query parameter \"intensity\" must be in half closed range (0.0, 1.0]: " + intensityStr.get());
         }
       } catch (NumberFormatException ex) {
         throw new ValidationException("invalid value for query parameter \"intensity\": " + intensityStr.get());
       }
       return Double.parseDouble(intensityStr.get());
   }
```

then the lines below could be…
```
    if (intensityStr.isPresent()) {
       try {
           return updateRepairRunIntensity(repairRun.get(), repairUnit.get(), segmentsRepaired, parseIntensity(intensityStr));
        } catch (ValidationException ex) {
           return Response.status(Response.Status.BAD_REQUEST).entity(ex.getMessage()).build();
        }
     }
```
(and the previous and first call to `validateIntensity(..)` from `checkRequestForAddRepair(..)` would remain inside the `if (intensityStr.isPresent()) {..}` conditional)

src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
```diff
@@ -381,6 +393,23 @@ public Response modifyRunState(
...
+    }
+
+    if (isIntensityPresent) {
+      final Response intensityValidationResponse = validateIntensity(intensityStr);
+      if (intensityValidationResponse != null) {
+        return intensityValidationResponse;
+      }
+
+      final double intensity = Double.parseDouble(intensityStr.get());
+      return editRun(repairRun.get(), repairUnit.get(), segmentsRepaired, intensity);
```
michaelsembwever: Continued from https://github.com/thelastpickle/cassandra-reaper/pull/303#discussion_r157944766

> So there will be just one PUT method but it will reroute call through finite state machine graph.

That would not be idempotent, which is required of PUT requests. 

> I thought about another endpoint but current logic means…

The existing endpoint isn't perfect, but is difficult to remove.
So I'd suggest adding the two following endpoints:

 - `PUT repair_run/{id}/state/{state}`
 - `PUT repair_run/{id}/intensity/{intensity}`

And change the existing endpoint `PUT repair_run/{id}?state={state}` to return a MOVED_PERMANENT to `PUT repair_run/{id}/state/{state}`.

Then logic for each endpoint would be:

 - `PUT repair_run/{id}?state={state}`

 1. if request does not specify new state, then BAD_REQUEST is returned,
 1. if the repair run (or unit) is not found, then NOT_FOUND is returned,
 1. return MOVED_PERMANENTLY to `PUT repair_run/{id}/state/{state}`.

 - `PUT repair_run/{id}/state/{state}`

 1. if request does not specify new state, then BAD_REQUEST is returned,
 1. if the repair run (or unit) is not found, then NOT_FOUND is returned,
 1. if there's any other repair runs for the related repair unit and they are RUNNING, then return CONFLICT,
 1. if state is unaltered, all other parameters are ignored, and NOT_MODIFIED is returned,
 1. if state is altered, but an invalid transition, then all other parameters are ignored, and METHOD_NOT_ALLOWED returned,
 1. if state is altered, the transition valid, then all other parameters are ignored, and OK returned.

 - `PUT repair_run/{id}/intensity/{intensity}`

 1. if request does not specify new state, then BAD_REQUEST is returned,
 1. if the repair run (or unit) is not found, then NOT_FOUND is returned,
 1. if there's any other repair runs for the related repair unit and they are RUNNING, then return CONFLICT,
 1. if state is not PAUSED then return CONFLICT,
 1. if state is PAUSED, intensity specified but invalid, then BAD_REQUEST is returned,
 1. if state is PAUSED, intensity specified and valid, then OK is returned,

src/server/src/test/java/io/cassandrareaper/resources/RepairRunResourceTest.java
kipwoker: I did it. Check it out.

src/server/src/main/java/io/cassandrareaper/service/RepairManager.java
kipwoker: Okay

src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
kipwoker: It will be exception logic. I prefer to avoid that in case when I know what can be wrong. Exceptions for exceptional situations. It's better to implement DDD conceptions here.
Intensity — it's a value object. It's not double actually, because for this domain logic it's float value between (0.0, 1.0]. So better to separate domain logic and infrastructure part of code (like working with db layer and working with api layer). But there is a lot of work for refactoring.
1. First of all RepairRun object must have field intensity with it own type like `private Intensity intensity`.
2. Domain data type Intensity must contain logic about it rightness (like value boundaries (0.0, 1.0]).
3. Finally project must provide opportunity for correct object creation (through the factory or static factory method).

I will encapsulate this idea only on RepairRunresource level, because have no time for huge refactoring.

src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
kipwoker: I've made some refactoring. But it still looks like spaghetti code. You should look to Railway Oriented Programming for validation: https://fsharpforfunandprofit.com/rop/
There are examples for C#/F# but I hope you can understand concept.

kipwoker:
> Good question. Have you looked through the commit history for clues? Have you tried testing locally on a fully loaded/saturated machine? (travis/circleci behave quite differently to laptops).
Anyway… this is off-topic. If you find anything interesting it warrants a new issue/PR.

I wrote my own tests without Thread.sleep and it passed checks at your ci systems. I also read commit history but commit messages didn't clarify anything for me.

src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
michaelsembwever: > Exceptions for exceptional situations. 

I would agree with this over public methods and public APIs, but here we're working privately inside the class.

The approach I suggested was to minimise the patch size: less code (so long as it is separates concerns and is readable) means easier evolution and less bugs. The addition of Intensity and Result classes outside of the package unnecessarily expands the API surface. 

I'll take what you got that is good and code up what I was thinking…

src/server/src/test/java/io/cassandrareaper/resources/RepairRunResourceTest.java
michaelsembwever: sry, where is it?

src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
michaelsembwever: > I'll take what you got that is good and code up what I was thinking…

This [commit](https://github.com/thelastpickle/cassandra-reaper/commit/e3e772945f1b8676140a74a944fbb3fc1a33e964) is more along the lines of what I was thinking: https://github.com/thelastpickle/cassandra-reaper/commit/e3e772945f1b8676140a74a944fbb3fc1a33e964

It provides the feature in half the patch size.

src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
michaelsembwever: Take a look at the rewrite i did in https://github.com/thelastpickle/cassandra-reaper/commit/e3e772945f1b8676140a74a944fbb3fc1a33e964

If we wanted to separate the validation concerns the approach for this code would be via jaxrs validation:  https://samaxes.com/2014/04/jaxrs-beanvalidation-javaee7-wildfly/
(and that would be in a separate PR)

src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
kipwoker: > Result classes outside of the package unnecessarily expands the API surface.

What are you talking about? Result and Intensity encapsulated at cassandra-reaper-server. I used them for RepairRunResource only.

src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
kipwoker: Actually I tired of arguing about this PR for a week. I just want to let it go. If you think that it's okay, accept PR. I just need possibility to change intensity.

kipwoker: I read your changes and understand what you mean. Would you like to merge it to master?

src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
michaelsembwever: > What are you talking about? Result and Intensity encapsulated at cassandra-reaper-server. I used them for RepairRunResource only.

 By "package" here I meant the java package. Result and Intensity had been put into `io.cassandrareaper.core`, a package only for model beans.
The view beans are in a different package `io.cassandrareaper.resources.view` (which i don't agree with, they should be in the resources package).
And we have no validation (or input) beans yet. 

For more background info on this take a look at https://github.com/thelastpickle/cassandra-reaper/issues/163

src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
michaelsembwever: > I read your changes and understand what you mean. Would you like to merge it to master?

Ok. I will merge it. Thanks.

> Actually I tired of arguing about this PR for a week. I just want to let it go. 

Please don't be. Putting the extra effort into each contribution is always worth it, especially as it's often someone else that has to fix or build something on top of it later on.
There's been no argument here. [Technical detailed and dispassionate criticism](http://producingoss.com/en/managing-participants.html#praise-and-criticism) under peer review helps produce solid software. Such technical criticism is a form of flattery: if your contribution was no good you would not have receive any feedback at all. 

Again, thanks for the contribution and effort.

michaelsembwever:
ps @kipwoker ,  did you add tests in `RepairManagerTest`?
i'm a bit confused by the outcome of this thread: https://github.com/thelastpickle/cassandra-reaper/pull/303#discussion_r158226910

kipwoker:
@michaelsembwever in final version I wrote only `testModifyIntensity` at `RepairRunResourceTest`. I saw tests at `RepairManagerTest` and they full of mocks. I understand it. You need it to test difficult logic for another methods. But there is nothing to test at `updateRepairRunIntensity` method.

Actually we should avoid mocks in our tests. Take a look at good article which explains that https://medium.com/javascript-scene/mocking-is-a-code-smell-944a70c90a6a
Better to test such functionality through integration tests. But i didn't find place where I may implement this logic.

michaelsembwever:
> Actually we should avoid mocks in our tests. 

Completely agree on the comment regarding the use of mocking in unit tests.

Have added a test in `RepairManagerTest` (removing one mock/spy at least).

Otherwise your work @kipwoker has been merged under https://github.com/thelastpickle/cassandra-reaper/pull/308

PR #305, opened 2017-12-19
allow EC2 multiregion address translation for custom domains

vektorich:
We have a custom DNS setup allowing cross-cloud communication.

The problem is that having custom DNS prevents current implementation of AddressTranslator from working correctly.

I have made a patch to the code which solves the problem without breaking any existing functionality:

1) copied the code of com.datastax.driver.core.policies.EC2MultiRegionAddressTranslator from datastax driver to io.cassandrareaper.jmx.CustomEC2MultiRegionAddressTranslator and replaced the references to it.

2) added code to remove the custom domain name part (defined by new config parameter String addressTranslatorRemoveDomain )

Example config:
useAddressTranslator: true
addressTranslatorRemoveDomain: .custom.pub

I think this patch would be useful for any cloud users that have custom domains defined.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/ReaperApplication.java
+++ b/src/server/src/main/java/io/cassandrareaper/ReaperApplication.java
@@ -15,6 +15,7 @@
 package io.cassandrareaper;
 
 import io.cassandrareaper.ReaperApplicationConfiguration.JmxCredentials;
+import io.cassandrareaper.jmx.CustomEC2MultiRegionAddressTranslator;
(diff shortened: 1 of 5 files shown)


adejanovski:
Hi @vektorich,

thanks for sending that PR.
While we could live with some code duplicated from the Datastax Java Driver, we feel like this fix may better be pushed to the original project, and improve the existing address translator.

Did you try to contribute that change to the driver project ?
If ever they refused that patch we could consider merging it into Reaper.

Thanks !

michaelsembwever:
Closing, until we hear further feedback…

PR #307, opened 2018-01-01
NodeStatus: Handle hostnames in endpointsStatus output

tedoc2000:
Cassandra does a ipAddress.toString() when creating the output
for FailureDetectorMBean.getAllEndpointStates(), which can
make the ip address header lines look like:
hostname/<ipaddress>

It also does the same for the key in FailureDetectorMBean.getSimpleStates()

Changed code to not use regexes to split getAllEndpointStates() output but
instead to group them into strings by looking for lines that don't start
with two spaces and treating those as record "starts". Hostnames are stripped
from the header lines if present. getSimpleStates() keys are also sanitized of
hostnames.

Added a test with hostnames in the output from getAllEndpointStates() and
getSimpleStates()


tedoc2000:
Verified that cassandra always returns lines separated by "\n" and always indents with two spaces

src/server/src/main/java/io/cassandrareaper/resources/view/NodesStatus.java
```diff
@@ -85,11 +86,39 @@ private GossipInfo parseEndpointStatesString(
...
     Matcher matcher;
 
-    String[] strEndpoints = allEndpointStates.split("(?<![0-9a-zA-Z ])/");
+    // Split into endpointState record strings
+    String[] endpointLines = allEndpointStates.split("\n");
+    List<String> strEndpoints = Lists.newArrayList();
+    int idx = 0;
+    StringBuilder oldLine = null;
+    do {
+      String line = endpointLines[idx];
```
udaysagar2177: for-each loop

adejanovski:
Hi @tedoc2000,

sorry for the time it took to review you PR and thank you for submitting it.
I've stumbled upon that problem myself just 2 days ago and am actually happy someone already fixed it for me :D  

Tests are looking good, I'll merge.

Thanks again !

PR #310, opened 2018-01-11
Fix computing too many segments on high clusters with high number of vnodes

adejanovski:
Since we moved to a segment count per node, the number of globally generated segments
per cluster has been over computed on clusters with 256 vnodes per node.
As Reaper tries to generate equally sized segments, using max(16, number_of_vnodes) created
segment sizes that ended up splitting large vnodes into smaller ones, resulting in having way
more segments than vnodes. We now try to split nodes by 16 and create one segment per node
if there are more than 16.


adejanovski:
@michaelsembwever : let me know if you can review this in priority has it will speed up repairs by orders of magnitude in many cases, and there is no workaround on the current release.

Thanks !

michaelsembwever:
My favourite PRs are those that you can go to the test changes and understand everything from that! 👍

src/server/src/main/java/io/cassandrareaper/service/RepairRunService.java
```diff
@@ -177,18 +175,12 @@ public RepairRun registerRepairRun(
 
   static int computeGlobalSegmentCount(
       int segmentCountPerNode,
-      Map<List<String>, List<String>> rangeToEndpoint,
       Map<String, List<RingRange>> endpointToRange) {
 
     int nodeCount = Math.max(1, endpointToRange.keySet().size());
```
michaelsembwever: is `endpointToRange.keySet().size()` ever less than one?

_if it were me_ i would tighten things up here.
`rangeToEndpoint` comes from `buildEndpointToRangeMap(..)` which in turn comes from `jmxProxy.getRangeToEndpointMap(..)`.

If there are no endpoints for the range in question we've got ourselves a programmatic error (i think) and we should catch that accordingly.

So the method could be written like:
```
  static int computeGlobalSegmentCount(int segmentCountPerNode, Map<String, List<RingRange>> endpointToRange) {
    Preconditions.checkArgument(1 <= endpointToRange.keySet().size());

    return endpointToRange.keySet().size()
        * (segmentCountPerNode != 0 ? segmentCountPerNode : DEFAULT_SEGMENT_COUNT_PER_NODE);
  }
```
(i'd also put an assert check in buildEndpointToRangeMap, before the return statement)

michaelsembwever (approved):
looks good to me.
i'd tighten up the `computeGlobalSegmentCount(..)` method, but whether that's a related part of this PR is an opinion.

adejanovski:
I've reworked `computeGlobalSegmentCount()` according to your suggestions.
Thanks for the review !

PR #316, opened 2018-01-19
Add support for handling clusters with different JMX credentials

### LINKED ISSUE #143: JMX credentials per cluster
Currently only one set of JMX credentials can be specified in the reaper configuration, meaning that multiple clusters with different JMX credentials require multiple reaper installations.

This is somewhat related to #70. Would be great if one reaper instance could support multiple clusters.

adejanovski:
Fixes #143 

This PR allows to handle clusters with different credentials, which are stored in the reaper yaml file (storing it in the db would require some fancy encryption that I'm not sure we want to handle).

JMX credentials must be added by cluster name : 

```
jmxCredentials:
  clusterProduction1:
    username: user1
    password: password1
  clusterProduction2:
    username: user2
    password: password2
```

Adding a new cluster with such authentication requires to pass the name of the cluster with the ip/hostname : `10.0.1.10@clusterProduction1`
This way, Reaper can locate the right credentials to use as we don't know the cluster name as long as we didn't get it... through JMX :)

`jmxAuth` is still usable and will be chosen by default unless one specific cluster's credentials are defined in `jmxCredentials`.

Tests and docs have been updated appropriately.


src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
```diff
@@ -536,23 +541,29 @@ private void handlePotentialStuckRepairs(LazyInitializer<Set<String>> busyHosts,
...
-        NodeMetrics metrics = NodeMetrics.builder()
-            .withNode(node)
-            .withDatacenter(nodeDc)
-            .withCluster(nodeProxy.getClusterName())
-            .withPendingCompactions(nodeProxy.getPendingCompactions())
-            .withHasRepairRunning(nodeProxy.isRepairRunning())
-            .withActiveAnticompactions(0) // for future use
-            .build();
+      try (JmxProxy nodeProxy =
+          context.jmxConnectionFactory.connect(
```
michaelsembwever: trivial: put the operator (`=`) on the new line. (so the precedence becomes: operators or arguments start a wrapped line)

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
```diff
@@ -536,23 +541,29 @@ private void handlePotentialStuckRepairs(LazyInitializer<Set<String>> busyHosts,
...
 
         return Pair.of(node, Optional.of(metrics));
       } catch (RuntimeException | ReaperException e) {
-        LOG.debug("failed to query metrics for host {}, trying to get metrics from storage...", node, e);
+        LOG.debug(
+            "failed to query metrics for host {}, trying to get metrics from storage...", node, e);
 
-        if (DatacenterAvailability.ALL != context.config.getDatacenterAvailability() && !nodeDc.equals(localDc)) {
+        if (DatacenterAvailability.ALL != context.config.getDatacenterAvailability()
+            && !nodeDc.equals(localDc)) {
```
michaelsembwever: trivial: why are these lines getting changed?

src/server/src/test/resources/cassandra-reaper-cassandra-at.yaml
```diff
@@ -37,6 +37,14 @@ jmxPorts:
...
   127.0.0.7: 7700
   127.0.0.8: 7800
+  
+jmxCredentials:
+  "test cluster":
+    username: cassandra
+    password: cassandra
+  cluster2:
+    username: cassandra
+    password: cassandra2
```
michaelsembwever: do we still also have a cassandra yaml (and associated test runs) on a non-jmx-secured cluster?

src/server/src/main/java/io/cassandrareaper/ReaperApplicationConfiguration.java
```diff
@@ -14,6 +14,8 @@
 
 package io.cassandrareaper;
 
+import io.cassandrareaper.core.Cluster;
```
michaelsembwever: the config classes should not depend upon the model beans.
this relates to the work in https://github.com/thelastpickle/cassandra-reaper/issues/163
specifically the last comment https://github.com/thelastpickle/cassandra-reaper/issues/163#issuecomment-325483387 where it's visualised that this class is one of the untangled classes (and it would be nice to keep it that way).

i think this has to be solved by moving the logic introduced in `getJmxCredentialsForCluster(..)` back up a level.
This concern (lack of encapsulation) is also repeated in the concern that we've added a parameter to the _connect_ methods in `JmxConnectionFactory` (procedural vs object-oriented code).

I'm aware of the difficulty around the jmx connections (and credentials) now being dependant on both host _and cluster_. My first guess is, to keep the whole `Map<String, JmxCredentials> jmxCredentials` as a final instance field in `JmxConnectionFactory` since this is not call dynamic. And then replace  each `String host` parameter in the different _connect*_ methods with a `Node node` parameter, where Node is a pojo with cluster and host.

The default `connectionTimeout` could also be stored as a final instance field, allowing simpler _connect*_ methods to be added and used in most places. But that's outside the scope of this PR.

michaelsembwever (changes requested):

adejanovski:
Hi @michaelsembwever,

I pushed a new commit to observe your recommendations : https://github.com/thelastpickle/cassandra-reaper/pull/316/commits/468dff2bd6524be1d3ce5667bb35581a58eba434

I'll be checking how we can test this during integration tests.

adejanovski:
@michaelsembwever : I've changed the integration tests so that we use authentication in JMX all the time.

The PR should hopefully be good to be merged now.

michaelsembwever (approved):

michaelsembwever:
LGTM 👍

PR #317, opened 2018-01-22
missing segment startTime when trigger repair has nothing to repair

michaelsembwever:
Another fix for `IllegalStateException: startTime must be set if segment is RUNNING or DONE`

 this time due to `context.storage.updateRepairSegment()` not properly setting startTime after coordinator.triggerRepair() returns 0 and segment is marked as DONE.

The segment's startTime is not set until in `processTriggeredSegment(..)`
  But in `runRepair()` there's a number of exit (return) points possible before this method is called (and without startTime being set).

ref: https://github.com/thelastpickle/cassandra-reaper/issues/291#issuecomment-359148919


adejanovski (approved):
LGTM

PR #325, opened 2018-01-31
added 1st set of metrics suggested

debswade:


zznate:
Matches style and clean test run. Thanks for the patch @debswade !

PR #342, opened 2018-02-13
Switch to re-usable JMX connections instead of re-creating one each time it's needed

adejanovski:
Reaper is currently leaking threads even though the JmxProxy object implements AutoCloseable
and all connections "seem" to be closed appropriately.
The fact that a new JMX connection is created for each segment and each time Reaper needs one
is a waste of resource though and opens up the possibility of leaks.
This new implementation involves re-uses JMX connections and ties SegmentRunners with JmxProxy objects
to properly handle async repair status notifications.

Thread usage before the patch : 
capture d ecran 2018-02-13 a 20 21 06

And after the patch : 
capture d ecran 2018-02-13 a 20 20 47

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/jmx/JmxConnectionFactory.java
+++ b/src/server/src/main/java/io/cassandrareaper/jmx/JmxConnectionFactory.java
@@ -24,19 +24,23 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ConcurrentMap;
 
+import com.codahale.metrics.Gauge;
 import com.codahale.metrics.MetricRegistry;
 import com.datastax.driver.core.policies.EC2MultiRegionAddressTranslator;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Function;
 import com.google.common.base.Optional;
 import com.google.common.base.Preconditions;
+import com.google.common.collect.Maps;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class JmxConnectionFactory {
 
   private static final Logger LOG = LoggerFactory.getLogger(JmxConnectionFactory.class);
-
+  private static final ConcurrentMap<String, JmxProxy> JMX_CONNECTIONS = Maps.newConcurrentMap();
   private final MetricRegistry metricRegistry;
   private final HostConnectionCounters hostConnectionCounters;
   private Map<String, Integer> jmxPorts;
@@ -47,11 +51,19 @@ public class JmxConnectionFactory {
   public JmxConnectionFactory() {
     this.metricRegistry = new MetricRegistry();
     hostConnectionCounters = new HostConnectionCounters(metricRegistry);
+    registerConnectionsGauge();
   }
 
   public JmxConnectionFactory(MetricRegistry metricRegistry) {
     this.metricRegistry = metricRegistry;
     hostConnectionCounters = new HostConnectionCounters(metricRegistry);
+    registerConnectionsGauge();
+  }
+
+  private void registerConnectionsGauge() {
+    this.metricRegistry.register(
+        MetricRegistry.name(JmxConnectionFactory.class, "openJmxConnections"),
+        (Gauge<Integer>) () -> JMX_CONNECTIONS.size());
   }
 
   protected JmxProxy connect(Optional<RepairStatusHandler> handler, String host, int connectionTimeout)
@@ -61,19 +73,30 @@ public class JmxConnectionFactory {
       host = host + ":" + jmxPorts.get(host);
     }
 
-    String username = null;
-    String password = null;
-    if (jmxAuth != null) {
-      username = jmxAuth.getUsername();
-      password = jmxAuth.getPassword();
-    }
+    String username = (jmxAuth != null) ? jmxAuth.getUsername() : null;
+    String password = (jmxAuth != null) ? jmxAuth.getPassword() : null;
+
     try {
-      JmxProxy jmxProxy = JmxProxyImpl.connect(handler, host, username, password, addressTranslator, connectionTimeout);
-      hostConnectionCounters.incrementSuccessfulConnections(host);
-      return jmxProxy;
-    } catch (ReaperException | RuntimeException e) {
-      hostConnectionCounters.decrementSuccessfulConnections(host);
-      throw e;
+      JmxConnectionProvider provider =
+          new JmxConnectionProvider(
+              host, username, password, connectionTimeout, this.metricRegistry);
+      JMX_CONNECTIONS.computeIfAbsent(host, provider::apply);
+      JmxProxy proxy = JMX_CONNECTIONS.get(host);
+      if (!proxy.isConnectionAlive()) {
+        LOG.info("Adding new JMX Proxy for host {}", host);
+        JMX_CONNECTIONS.put(host, provider.apply(host)).close();
+        proxy.close();
+      }
+      return JMX_CONNECTIONS.get(host);
+    } catch (RuntimeException ex) {
+      // unpack any exception behind JmxConnectionProvider.apply(..)
+      if (ex.getCause() instanceof InterruptedException) {
+        throw (InterruptedException) ex.getCause();
+      }
+      if (ex.getCause() instanceof ReaperException) {
+        throw (ReaperException) ex.getCause();
+      }
+      throw ex;
     }
   }
 
@@ -132,5 +155,42 @@ public class JmxConnectionFactory {
 
   public final HostConnectionCounters getHostConnectionCounters() {
     return hostConnectionCounters;
+  }
+
+  private class JmxConnectionProvider implements Function<String, JmxProxy> {
+
+    private final String host;
+    private final String username;
+    private final String password;
+    private final int connectionTimeout;
+    private final MetricRegistry metricRegistry;
+
+    JmxConnectionProvider(
+        String host,
+        String username,
+        String password,
+        int connectionTimeout,
+        MetricRegistry metricRegistry) {
+      this.host = host;
+      this.username = username;
+      this.password = password;
+      this.connectionTimeout = connectionTimeout;
+      this.metricRegistry = metricRegistry;
+    }
+
+    @Override
+    public JmxProxy apply(String host) {
+      Preconditions.checkArgument(host.equals(this.host));
+      try {
+        JmxProxy proxy =
+            JmxProxyImpl.connect(
+                host, username, password, addressTranslator, connectionTimeout, metricRegistry);
+        hostConnectionCounters.incrementSuccessfulConnections(host);
+        return proxy;
+      } catch (ReaperException | InterruptedException ex) {
+        hostConnectionCounters.decrementSuccessfulConnections(host);
+        throw new RuntimeException(ex);
+      }
+    }
   }
 }
--- a/src/server/src/main/java/io/cassandrareaper/jmx/JmxProxy.java
+++ b/src/server/src/main/java/io/cassandrareaper/jmx/JmxProxy.java
@@ -23,6 +23,7 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+
 import javax.management.AttributeNotFoundException;
 import javax.management.MBeanException;
 import javax.management.NotificationListener;
@@ -32,7 +33,7 @@ import javax.validation.constraints.NotNull;
 import org.apache.cassandra.repair.RepairParallelism;
 
 
-public interface JmxProxy extends AutoCloseable, NotificationListener {
+public interface JmxProxy extends NotificationListener {
 
   Duration DEFAULT_JMX_CONNECTION_TIMEOUT = Duration.ofSeconds(5);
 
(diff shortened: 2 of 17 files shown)


michaelsembwever:
trivial: the commit message has some grammer mistakes, and odd line wrapping.

src/server/src/main/java/io/cassandrareaper/jmx/JmxProxy.java
```diff
@@ -111,9 +112,10 @@
   List<String> tokenRangeToEndpoint(String keyspace, RingRange tokenRange);
 
   /**
-   * Triggers a repair of range (beginToken, endToken] for given keyspace and column family. The repair is triggered by
-   * {@link org.apache.cassandra.service.StorageServiceMBean#forceRepairRangeAsync} For time being, we don't allow local
-   * nor snapshot repairs.
+   * Triggers a repair of range (beginToken, endToken] for given keyspace and column family. The
+   * repair is triggered by {@link
+   * org.apache.cassandra.service.StorageServiceMBean#forceRepairRangeAsync} For time being, we
```
michaelsembwever: can the link macro be wrapped in apidoc like that?

src/server/src/main/java/io/cassandrareaper/jmx/JmxProxyImpl.java
```diff
@@ -848,30 +890,19 @@ public boolean isConnectionAlive() {
...
-      } catch (InstanceNotFoundException | ListenerNotFoundException | IOException e) {
-        LOG.debug("failed on removing notification listener", e);
-      }
-    }
-    try {
-      jmxConnector.close();
-    } catch (IOException e) {
-      LOG.warn("failed closing a JMX connection", e);
-    }
+    // Not closing anymore, we re-use JMX connections now
```
michaelsembwever: i don't believe this is true.

JmxProxy is no longer an AutoClosable. That needs to be fixed. 
That also involves a handful more lines of code change (moving the resource declaration in try-resource statements onto the subsequent line inside the try-catch block).

But this close method still is required (with all the previous code) as is evident on this line https://github.com/thelastpickle/cassandra-reaper/pull/342/files#diff-739e34b5f2ca85ff72da4077afc1ea01R87

src/server/src/main/java/io/cassandrareaper/jmx/JmxProxyImpl.java
```diff
@@ -112,6 +111,7 @@
   private final String hostBeforeTranslation;
   private final JMXServiceURL jmxUrl;
   private final String clusterName;
+  private final HashMap<Integer, RepairStatusHandler> repairStatusHandlers = Maps.newHashMap();
```
Bj0rnen: Seems like you can also remove the old `private final Optional<RepairStatusHandler> repairStatusHandler`.

src/server/src/main/java/io/cassandrareaper/jmx/JmxProxyImpl.java
```diff
@@ -790,12 +821,15 @@ public void handleNotification(Notification notification, Object handback) {
     // we're interested in "repair"
     String type = notification.getType();
     LOG.debug(
-        "Received notification: {} with type {} and repairStatusHandler {}", notification, type, repairStatusHandler);
-    if (repairStatusHandler.isPresent() && ("repair").equals(type)) {
+        "Received notification: {} with type {} and repairStatusHandler {}",
+        notification,
+        type,
+        repairStatusHandlers);
```
Bj0rnen: Should it print the whole map? Maybe you don't get the command id, so you can't do anything else? What's the value of printing the handler anyway?

Bj0rnen: Ah I see you get `repairNo` in two different ways depending on API. Maybe log in those functions?

src/server/src/main/java/io/cassandrareaper/jmx/JmxConnectionFactory.java
```diff
@@ -68,9 +70,27 @@ protected JmxProxy connect(Optional<RepairStatusHandler> handler, String host, i
...
+      if (previousConnection == null) {
+        // The putIfAbsent succeeded and our JmxProxy object is the one in the map
+        LOG.info("Adding new JMX Proxy for host {}", host);
+
+      } else {
+        if (previousConnection.isConnectionAlive()) {
+          LOG.info("Reusing previously created JMX connection for host {}", host);
+          jmxProxy.close();
+        } else {
+          LOG.info("Recreating JMX connection for host {}", host);
```
michaelsembwever: the previousConnection should be closed here

```
previousConnection.close();
```

michaelsembwever (changes requested):
Good work.

A couple of small improvements are required i do believe.

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
```diff
@@ -794,16 +797,22 @@ private boolean handleJmxNotificationForCassandra22(
...
       case COMPLETE:
         // This gets called through the JMX proxy at the end
         // regardless of succeeded or failed sessions.
         LOG.debug(
-            "repair session finished for segment with id '{}' and repair number '{}'", segmentId, repairNumber);
+            "repair session finished for segment with id '{}' and repair number '{}'",
+            segmentId,
+            repairNumber);
         condition.signalAll();
+        jmxProxy.removeRepairStatusHandler(repairNumber);
```
Bj0rnen, line 813: Is this actually done in all code paths that are supposed to stop the SegmentRunner? Like failure cases. It seems to me like the status handler has to be remove or else `this` will never be garbage collected (since the status handler _is_ `this`). You've probably thought about this, I'm just asking for a double-check.

Bj0rnen (changes requested):

src/server/src/main/java/io/cassandrareaper/jmx/JmxProxyImpl.java
adejanovski: I've removed the extension of AutoClosable in JmxProxy and fixed all try-catch-resource blocks involving it.
`close()` is back and only used in the `JmxConnectionFactory` to close defunct JMX connections.

src/server/src/main/java/io/cassandrareaper/jmx/JmxConnectionFactory.java
adejanovski: I've reworked that part to make it safer and avoid creating a new JMX connection only to close it if another one already existed.
The new code involves a synchronized section to make sure we're not re-creating the connections several times and to simplify the code (contention is not an issue here as the synchronized block is only used if there's no existing connection for that host).

src/server/src/main/java/io/cassandrareaper/jmx/JmxProxyImpl.java
adejanovski: Right ! 

It's deleted now.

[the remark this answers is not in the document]
adejanovski: I moved it to a `ConcurrentMap` and now write to it with `putIfAbsent()`.
The rest of the instance fields look good to me in terms of concurrency.

src/server/src/main/java/io/cassandrareaper/jmx/JmxProxyImpl.java
adejanovski: Logging the full map doesn't make sense indeed.
I've followed your recommendation and logged in `processOldApiNotification` and `processNewApiNotification` using the repair number.

adejanovski:
@Bj0rnen @michaelsembwever : I've made the requested changes, hopefully.
Let me know if it's looking good for you now, and thanks for the reviews !

src/server/src/main/java/io/cassandrareaper/jmx/JmxProxyImpl.java
```diff
@@ -848,24 +896,23 @@ public boolean isConnectionAlive() {
...
+      LOG.debug("Couldn't get Connection Id", e);
+      return false;
     }
-    return false;
   }
 
-  /**
-   * Cleanly shut down by un-registering the listener and closing the JMX connection.
-   */
+  public void removeRepairStatusHandler(int commandId) {
```
michaelsembwever: deserves `@Override`

src/server/src/main/java/io/cassandrareaper/jmx/JmxConnectionFactory.java
```diff
@@ -24,19 +24,21 @@
...
 import com.google.common.base.Preconditions;
+import com.google.common.collect.Maps;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class JmxConnectionFactory {
 
   private static final Logger LOG = LoggerFactory.getLogger(JmxConnectionFactory.class);
-
+  private static final ConcurrentMap<String, JmxProxy> JMX_CONNECTIONS = Maps.newConcurrentMap();
```
michaelsembwever, line 43: i would put a metrics gauge on this.

src/server/src/main/java/io/cassandrareaper/jmx/JmxProxyImpl.java
```diff
@@ -107,14 +108,14 @@
   private final EndpointSnitchInfoMBean endpointSnitchMbean;
   private final Object ssProxy;
   private final Object fdProxy;
-  private final Optional<RepairStatusHandler> repairStatusHandler;
   private final String host;
   private final String hostBeforeTranslation;
   private final JMXServiceURL jmxUrl;
   private final String clusterName;
+  private final ConcurrentMap<Integer, RepairStatusHandler> repairStatusHandlers =
+      Maps.newConcurrentMap();
```
michaelsembwever, line 118: it would be good with a metrics gauge on this.

especially as i can see leaks happening: the call to `removeRepairStatusHandler(..)` is happening in a number of places outside of finally blocks.

src/server/src/main/java/io/cassandrareaper/jmx/JmxConnectionFactory.java
```diff
@@ -68,9 +70,36 @@ protected JmxProxy connect(Optional<RepairStatusHandler> handler, String host, i
...
+            JMX_CONNECTIONS.putIfAbsent(
+                host,
+                JmxProxyImpl.connect(
+                    host, username, password, addressTranslator, connectionTimeout));
+            hostConnectionCounters.incrementSuccessfulConnections(host);
+          }
+        }
+      }
+
+      return JMX_CONNECTIONS.get(host);
```
michaelsembwever: the whole `connect(..)` method could be written a lot simpler (concurrency) with:

```
  protected JmxProxy connect(Optional<RepairStatusHandler> handler, String host, int connectionTimeout)
      throws ReaperException, InterruptedException {
    // use configured jmx port for host if provided
    if (jmxPorts != null && jmxPorts.containsKey(host) && !host.contains(":")) {
      host = host + ":" + jmxPorts.get(host);
    }

    String username = (jmxAuth != null) ? jmxAuth.getUsername() : null;
    String password = (jmxAuth != null) ? jmxAuth.getPassword() : null;

    try {
      JmxConnectionProvider provider = new JmxConnectionProvider(host, username, password, connectionTimeout);
      JMX_CONNECTIONS.computeIfAbsent(host, provider);
      JmxProxy proxy = JMX_CONNECTIONS.get(host);
      if (!proxy.isConnectionAlive()) {
        JMX_CONNECTIONS.put(host, provider.apply(host));
        proxy.close();
      }
      return JMX_CONNECTIONS.get(host);
    } catch (RuntimeException ex) {
      // unpack any exception behind JmxConnectionProvider.apply(..)
      if (ex.getCause() instanceof InterruptedException) {
        throw (InterruptedException) ex.getCause();
      }
      if (ex.getCause() instanceof ReaperException) {
        throw (ReaperException) ex.getCause();
      }
      throw ex;
    }
  }
```

along with adding the inner class:
```
  private class JmxConnectionProvider implements Function<String, JmxProxy> {

    private final String host;
    private final String username;
    private final String password;
    private final int connectionTimeout;

    public JmxConnectionProvider(String host, String username, String password, int connectionTimeout) {
      this.host = host;
      this.username = username;
      this.password = password;
      this.connectionTimeout = connectionTimeout;
    }

    @Override
    public JmxProxy apply(String host) {
      Preconditions.checkArgument(host.equals(this.host));
      LOG.info("Adding new JMX Proxy for host {}", host);
      try {
        JmxProxy proxy = JmxProxyImpl.connect(host, username, password, addressTranslator, connectionTimeout);
        hostConnectionCounters.incrementSuccessfulConnections(host);
        return proxy;
      } catch (ReaperException | InterruptedException ex) {
        hostConnectionCounters.decrementSuccessfulConnections(host);
        throw new RuntimeException(ex);
      }
    }
  }
```

The synchronized block and double-lock isn't required, and is lacking as it doesn't cover lines 73-75.
To be completely concurrent safe i think a read-write lock would be best.
But I reckon the alternative provided in this comment is good enough by relying solely on what ConcurrentMap provides. 

And if `JmxProxyImpl.close()` was made idempotent I'd replace the line
```
JMX_CONNECTIONS.put(host, provider.apply(host));
```
with
```
JMX_CONNECTIONS.put(host, provider.apply(host)).close();
```

michaelsembwever:
Cool work @adejanovski . This is a significant and obviously really important improvement to Reaper.

PR #350, opened 2018-02-21
updating metrics for repairProgress,segmentsDone,segmentsTotal and ad…

kingm5:
PR that will make segmentsDone,segmentsTotal,millisSinceLastRepair work and added the keyspace to the metric. This really helps .. as you can now track progress on a keyspace repair separately and the last time it was repaired. Regarding the fix to segmentsDone .. it now will update as the segments are getting processed. Regarding the fix to segmentsTotal .. it now will write the total before the repair starts. This way.. you know how many segments are going to be processed and you can track them again in realtime. Also, I changed the types to integers as it did not make sense to me have as a float. I may be wrong here, but currently I do not believe we can have 5.3 segments.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
+++ b/src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
@@ -57,6 +57,7 @@ final class RepairRunner implements Runnable {
   private final AtomicReferenceArray<UUID> currentlyRunningSegments;
   private final List<RingRange> parallelRanges;
   private final String metricNameForMillisSinceLastRepair;
+  private final String metricNameForMillisSinceLastRepair2;
   private float repairProgress;
   private float segmentsDone;
   private float segmentsTotal;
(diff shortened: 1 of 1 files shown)


src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
```diff
@@ -98,13 +98,27 @@
...
+                                                     repairRunId);
     context.metricRegistry.register(metricNameForRepairProgress, (Gauge<Float>) ()  -> repairProgress);
-    String metricNameForSegmentsDone = metricName("segmentsDone",  repairUnitClusterName, repairRunId);
-    context.metricRegistry.register(metricNameForSegmentsDone, (Gauge<Float>) ()  -> segmentsDone);
-    String metricNameForSegmentsTotal = metricName("segmentsTotal",  repairUnitClusterName, repairRunId);
-    context.metricRegistry.register(metricNameForSegmentsTotal, (Gauge<Float>) ()  -> segmentsTotal);
-    metricNameForMillisSinceLastRepair = metricName("millisSinceLastRepair", repairUnitClusterName, repairRunId);
+    metricNameForMillisSinceLastRepair = metricName("millisSinceLastRepair",
+                                                     repairUnitClusterName,
+                                                     repairUnitKeyspaceName,
```
adejanovski: I'm afraid that we may be changing the name of a metric that people might already be monitoring. Would you mind duplicating the metric so that both the new and the old one will be there ?

src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
```diff
@@ -58,8 +58,8 @@
   private final List<RingRange> parallelRanges;
   private final String metricNameForMillisSinceLastRepair;
   private float repairProgress;
-  private float segmentsDone;
-  private float segmentsTotal;
+  private int   segmentsDone;
```
adejanovski: Indeed, no need for a float here :)

src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
```diff
@@ -499,10 +516,10 @@ void killAndCleanupRunner() {
     Thread.currentThread().interrupt();
   }
 
-  private String metricName(String metric, String clusterName, UUID repairRunId) {
+  private String metricName(String metric, String clusterName, String keyspaceName, UUID repairRunId) {
```
adejanovski: That method would then require to have a sibling with no keyspace.
But I do understand the value of adding the keyspace here, I'm clearly not arguing with it. It's just a matter of not breaking people's automation that could rely on existing metrics.

adejanovski (changes requested):
That's a great improvement over the existing metrics. Let's just make sure the previous set of metric still exists and we'll see in the future if it makes sense to fully remove them.
We'll be able to merge once this is done.
Thanks !

src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
kingm5: Thanks Alex... and yes I totally understand your review. I will update the code and let you know when it is complete.

kingm5:
Below are the old metrics taken with this PR.

cassandrareaper.service.RepairRunner.repairProgress,env=aos3,id=82143bc018ed11e89b3933dba62da889
cassandrareaper.service.RepairRunner.segmentsDone,env=aos3,id=82143bc018ed11e89b3933dba62da889
cassandrareaper.service.RepairRunner.segmentsTotal,env=aos3,id=82143bc018ed11e89b3933dba62da889
cassandrareaper.service.RepairRunner.millisSinceLastRepair,env=aos3,id=e79310b0191611e8aa521b9e0a9630b0

adejanovski (approved):

adejanovski:
LGTM, merged.

PR #351, opened 2018-02-22
Remove unnecessary postpones and randomize next segment in postgres

adejanovski:
Postpone is not necessary when a segment is denied to run by `canRepair()`.
I've added randomization when picking the next segment with the postgres backend to avoid cycling on the same one all the time, which is already what we did with Cassandra.

I know for a fact that there are some exceptions we do not catch which can lead to a segment runner ending up stuck in `SEGMENT_RUNNERS`. I've added a catch for `RuntimeException` so that we don't miss them and clean up the map correctly.
I've also changed the two step check/add to `SEGMENT_RUNNER` that was done with `.contains()` then `.put()` with a single `.putIfAbsent()` operation.

This way, we will only increment the fail count for segments that actually fail during repair and we will prevent the creation of loads of tombstones in the Cassandra backend.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
+++ b/src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
@@ -223,12 +223,22 @@ final class SegmentRunner implements RepairStatusHandler, Runnable {
         throw new ReaperException("SegmentRunner already exists for segment with ID: " + segmentId);
       }
       SEGMENT_RUNNERS.put(segmentId, this);
+
       String keyspace = repairUnit.getKeyspaceName();
       boolean fullRepair = !repairUnit.getIncrementalRepair();
 
       LazyInitializer<Set<String>> busyHosts = new BusyHostsInitializer(coordinator);
       if (!canRepair(segment, keyspace, coordinator, busyHosts)) {
-        postponeCurrentSegment();
+        LOG.info(
+            "Cannot run segment {} for repair {} at the moment. Will try again later",
+            segmentId,
+            segment.getRunId());
+        SEGMENT_RUNNERS.remove(segment.getId());
+        try {
+          Thread.sleep(SLEEP_TIME_AFTER_POSTPONE_IN_MS);
+        } catch (InterruptedException e) {
+          LOG.debug("Interrupted while sleeping after a segment was postponed... weird stuff...");
+        }
         return false;
       }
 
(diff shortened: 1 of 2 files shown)


michaelsembwever:
@adejanovski should be two separate commits, (and ideally two PRs)

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
```diff
@@ -1059,11 +1070,15 @@ private int countRunningReapers() {
...
       Set<String> busyHosts = Sets.newHashSet();
-      ongoingRepairs.forEach((ongoingRepair) -> {
-        busyHosts.addAll(coordinator.tokenRangeToEndpoint(ongoingRepair.keyspaceName, ongoingRepair.tokenRange));
-      });
+      ongoingRepairs.forEach(
+          (ongoingRepair) -> {
+            busyHosts.addAll(
+                coordinator.tokenRangeToEndpoint(
+                    ongoingRepair.keyspaceName, ongoingRepair.tokenRange));
+          });
```
michaelsembwever: trivial: 
 did anything actually changed in the `initialize()` method?
 can you take it out of the PR please. it's not a great precedence to encourage.

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
```diff
@@ -218,17 +218,27 @@ private boolean runRepair() {
...
+        LOG.info(
+            "Cannot run segment {} for repair {} at the moment. Will try again later",
+            segmentId,
+            segment.getRunId());
+        SEGMENT_RUNNERS.remove(segment.getId());
+        try {
+          Thread.sleep(SLEEP_TIME_AFTER_POSTPONE_IN_MS);
+        } catch (InterruptedException e) {
+          LOG.debug("Interrupted while sleeping after a segment was postponed... weird stuff...");
+        }
```
michaelsembwever, line 241: yeah, you're right about this approach. hadn't realised no version of a postpone action was required at all. good stuff!

michaelsembwever:
happy to merge if it's two commits.

adejanovski:
@michaelsembwever : I've reduced the scope to just not postponing and shuffling segments in the postgres backend.
I'll send another PR for the SEGMENT_RUNNERS issues.

michaelsembwever (approved):

michaelsembwever (approved):

Issue #355, opened 2018-03-05
Integration tests for spreaper


michaelsembwever:
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=13951915)

The spreaper python script has no unit/integration tests.

A number of releases have already occurred where key functionality has been broken.

Even a simple integration test would be hugely valuable, for example…

* ccm start
* start reaper server
* add cluster
* add schedule
* add repair
* pause repair
* change intensity
* resume repair

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/REAP-175) by [Unito](https://www.unito.io)
┆Issue Number: REAP-175

rzvoncek:
Just got bitten by spreaper not working properly. I'm bumping this issue to prevent that happening in the future.

PR #356, opened 2018-03-05
Fix `spreaper resume-run`, `spreaper pause-run`, `spreaper abort-run`

michaelsembwever:
Fix `spreaper resume-run`, `spreaper pause-run`, `spreaper abort-run`.

All were broken since: 6f71b67 – Implemented PUT method for change intensity
 as the existing endpoint was changed to return a redirect to a new url.
 The spreaper python script was handling the redirect, but maintaining the PUT http method over onto the new url.

The fix was to get spreaper to use the new endpoints, avoiding the redirect.
The commit also
 - fixes the modifyRunIntensity endpoint (that was introduced in 6f71b67) so that it can be used when the repair run is in NOT_STARTED state, as well as PAUSED state, and
 - correct some of the apidoc in RepairRunResource.

### DIFF
--- a/src/packaging/bin/spreaper
+++ b/src/packaging/bin/spreaper
@@ -518,7 +518,7 @@ class ReaperCLI(object):
         printq("# Repair run with id={0} created".format(repair_run.get('id')))
         if not args.dont_start_repair:
             printq("# Starting repair run with id={0}".format(repair_run.get('id')))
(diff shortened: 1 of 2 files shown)


michaelsembwever:
Because of this regression i've created https://github.com/thelastpickle/cassandra-reaper/issues/355 in hope this can be prevented from happening again.

adejanovski (approved):

PR #362, opened 2018-03-08
Add support for snapshots

adejanovski:


src/server/src/main/java/io/cassandrareaper/jmx/JmxConnectionFactory.java
```diff
@@ -118,7 +118,7 @@ protected JmxProxy connect(
     }
   }
 
-  public final JmxProxy connect(Node node, int connectionTimeout)
+  public JmxProxy connect(Node node, int connectionTimeout)
```
michaelsembwever, line 121: why unfinalise the method?

(why not mock the method instead? as is the new trend in this project's tests…)

src/server/src/main/java/io/cassandrareaper/service/RepairRunService.java
```diff
@@ -189,9 +189,9 @@ public RepairRun registerRepairRun(
   static int computeGlobalSegmentCount(
       int segmentCountPerNode,
       Map<String, List<RingRange>> endpointToRange) {
-    Preconditions.checkArgument(1 <= endpointToRange.keySet().size());
+    //Preconditions.checkArgument(1 <= endpointToRange.keySet().size());
```
michaelsembwever: why is this now ok?

src/server/src/main/java/io/cassandrareaper/service/SnapshotManager.java
```diff
@@ -0,0 +1,330 @@
...
+  public static final String SNAPSHOT_PREFIX = "reaper";
+  private static final Logger LOG = LoggerFactory.getLogger(SnapshotManager.class);
+  private final AppContext context;
+  private final DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd-HH-mm-ss");
+  private final ExecutorService executor = Executors.newFixedThreadPool(5);
+  private final Cache<String, Snapshot> cache =
+      CacheBuilder.newBuilder()
+          .maximumSize(1000) // Taille Max
+          .expireAfterWrite(1, TimeUnit.HOURS) // TTL
+          .build();
```
michaelsembwever: the only benefit i can see of the cache is that it avoids a call to `context.storage.getSnapshot(..)`.

What's the measured optimisation that this provides? 

Line 173 can be replaced with `.parallelStream()` and the calls to `context.storage.getSnapshot(..)` will execute in parallel. I would think this is more effective (and better on heap) than the cache approach.

michaelsembwever:
Looks really solid overall. Threw at you a few small questions.

src/server/src/main/java/io/cassandrareaper/service/SnapshotManager.java
adejanovski: As I think having the cache will matter in large clusters, I've moved the cache to a `weakValues()` one as discussed together.

src/server/src/main/resources/db/cassandra/011_snapshots.cql
```diff
@@ -0,0 +1,20 @@
+--
+-- Upgrade to handle snapshots metadata
+
+
+DROP TABLE IF EXISTS snapshot;
```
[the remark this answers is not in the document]
adejanovski: Right, that's not needed/wanted.

src/server/src/main/java/io/cassandrareaper/service/RepairRunService.java
adejanovski: It was just a leftover from an attempt to run integration tests that didn't go through.
I've reverted this change.

src/server/src/main/java/io/cassandrareaper/jmx/JmxConnectionFactory.java
adejanovski: That's actually how many of the tests are built and overriding those methods is necessary.
Due to how the JmxConnectionFactory is injected in the context, I couldn't use a mock instead (as far as I can remember).

PR #373, opened 2018-03-21
Implement token range coalescing

adejanovski:
Token range coalescing enables to group vnodes that share the exact same replicas into a single segment.
Running subrange repair for multiple ranges at once is possible since 2.2 and has been optimized starting in 3.0 to be executed as a single session whenever the ranges all share the same replicas.

This will remove most of the overhead vnodes create on repair jobs and bring us back to something similar to clusters that use a single token per node : 
while currently Reaper must create at least one segment per vnode in the cluster, that PR will remove that limitation and allow to create for example 10 segments per node on a 6 nodes cluster with 256 vnodes per node.


src/server/src/main/java/io/cassandrareaper/service/RepairRunService.java
```diff
@@ -164,16 +163,27 @@ public RepairRun registerRepairRun(
...
+              sg.generateSegments(
+                  globalSegmentCount,
+                  tokens,
+                  repairUnit.getIncrementalRepair(),
+                  replicasToRange,
+                  cassandraVersion),
               repairUnit,
               endpointToRange);
+
+      //segments.stream().map(segment -> Segment.builder().withTokenRanges(Arrays.asList(segment)).withReplicas())
```
michaelsembwever: can be removed?

src/server/src/main/java/io/cassandrareaper/service/SegmentGenerator.java
```diff
@@ -81,76 +87,186 @@ static boolean greaterThanOrEqual(BigInteger big0, BigInteger big1) {
...
-    BigInteger total = BigInteger.ZERO;
-    for (RingRange segment : repairSegments) {
-      BigInteger size = segment.span(rangeSize);
-      total = total.add(size);
+    return repairSegments;
+  }
+
+  @VisibleForTesting
+  List<Segment> coalesceTokenRanges(
+      BigInteger targetSegmentSize, Map<List<String>, List<RingRange>> replicasToRange) {
```
michaelsembwever: trivial: one parameter per line please.

src/server/src/main/java/io/cassandrareaper/service/SegmentGenerator.java
```diff
@@ -81,76 +87,186 @@ static boolean greaterThanOrEqual(BigInteger big0, BigInteger big1) {
...
+    List<RingRange> tokenRangesForCurrentSegment = Lists.newArrayList();
+    BigInteger tokenCount = BigInteger.ZERO;
+
+    for (Entry<List<String>, List<RingRange>> tokenRangesByReplica : replicasToRange.entrySet()) {
+      LOG.info("Coalescing segments for nodes {}", tokenRangesByReplica.getKey());
+      if (!tokenRangesForCurrentSegment.isEmpty()) {
+        coalescedRepairSegments.add(
+            Segment.builder().withTokenRanges(tokenRangesForCurrentSegment).build());
+        tokenRangesForCurrentSegment = Lists.newArrayList();
+      }
```
michaelsembwever: i would expect lines 201-205 to go at the end of the for-loop (ie line 223), so that you could remove lines 227-230.

src/server/src/main/java/io/cassandrareaper/service/SegmentGenerator.java
```diff
@@ -81,76 +87,186 @@ static boolean greaterThanOrEqual(BigInteger big0, BigInteger big1) {
...
 
   protected boolean inRange(BigInteger token) {
     return !(lowerThan(token, rangeMin) || greaterThan(token, rangeMax));
   }
+
+  private boolean supportsSegmentCoalescing(String cassandraVersion) {
+    return !cassandraVersion.startsWith("1.")
+        && !cassandraVersion.startsWith("2.0")
+        && !cassandraVersion.startsWith("2.1");
+  }
```
michaelsembwever: why is this only Cassandra-2.2+?

and given there's no reason for 256 vnodes in Cassandra-3+ is coalescing an important feature?

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
```diff
@@ -390,18 +398,23 @@ public RepairRun addRepairRun(Builder repairRun, Collection<RepairSegment.Builde
...
+                  segment.getState().ordinal(),
+                  segment.getFailCount(),
+                  objectMapper.writeValueAsString(segment.getTokenRange().getTokenRanges())));
+        } catch (JsonProcessingException e) {
+          throw new ReaperException(e);
+        }
       }
 
-      if (100 <= repairRunBatch.size()) {
+      if (100 <= repairRunBatch.size() || segment.getTokenRange().getTokenRanges().size() > 1) {
```
michaelsembwever: why does a segment with more than one token range warrant a new batch?

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
```diff
@@ -700,14 +713,27 @@ private static boolean segmentIsWithinRange(RepairSegment segment, RingRange ran
 
   private static RepairSegment createRepairSegmentFromRow(Row segmentRow) {
 
-    RepairSegment.Builder builder = RepairSegment.builder(
+    List<RingRange> tokenRanges =
+        JsonParseUtils.parseRingRangeList(
+            Optional.fromNullable(segmentRow.getString("token_ranges")));
+
+    io.cassandrareaper.core.Segment.Builder segmentBuilder = Segment.builder();
```
michaelsembwever: does not need to be fully qualified?

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
```diff
@@ -390,18 +398,23 @@ public RepairRun addRepairRun(Builder repairRun, Collection<RepairSegment.Builde
...
+                  segment.getRunId(),
+                  segment.getId(),
+                  segment.getRepairUnitId(),
+                  segment.getStartToken(),
+                  segment.getEndToken(),
+                  segment.getState().ordinal(),
+                  segment.getFailCount(),
+                  objectMapper.writeValueAsString(segment.getTokenRange().getTokenRanges())));
+        } catch (JsonProcessingException e) {
+          throw new ReaperException(e);
```
michaelsembwever: i don't think a JsonProcessingException here is a valid cause to throw a ReaperException.
this can only be an IllegalStateException, since the json was put there by us in the first place.

(that means not changing the method signature to throw a ReaperException)

src/server/src/test/java/io/cassandrareaper/resources/RepairRunResourceTest.java
```diff
@@ -119,8 +118,8 @@ public void setUp() throws Exception {
     when(proxy.getEndpointToHostId()).thenReturn(NODES_MAP);
     when(proxy.getTokens()).thenReturn(TOKENS);
     when(proxy.isConnectionAlive()).thenReturn(Boolean.TRUE);
-    when(proxy.tokenRangeToEndpoint(anyString(), any(RingRange.class))).thenReturn(
-        Collections.singletonList(""));
+    when(proxy.tokenRangeToEndpoint(anyString(), any(io.cassandrareaper.core.Segment.class)))
```
michaelsembwever: does not need to be fully qualified

src/server/src/main/java/io/cassandrareaper/storage/MemoryStorage.java
```diff
@@ -289,7 +289,8 @@ public boolean updateRepairSegment(RepairSegment newRepairSegment) {
   public Optional<RepairSegment> getNextFreeSegmentInRange(UUID runId, Optional<RingRange> range) {
     if (range.isPresent()) {
       for (RepairSegment segment : repairSegmentsByRunId.get(runId).values()) {
-        if (segment.getState() == RepairSegment.State.NOT_STARTED && range.get().encloses(segment.getTokenRange())) {
+        if (segment.getState() == RepairSegment.State.NOT_STARTED
+            && range.get().encloses(segment.getTokenRange().getTokenRanges().get(0))) {
```
michaelsembwever: why only the segment's first token range?

michaelsembwever:
this is a big patch, and if affects so many tests it hard to say what it breaks and does not…
(the consequences of getting coalescing wrong in production are serious)

furthermore, ultimately, it only affects Cassandra-2.2 users. As come Cassandra-3.0 users should start to migrate to lower vnode numbers.

src/server/src/main/java/io/cassandrareaper/service/RepairRunService.java
```diff
@@ -208,10 +216,11 @@ static int computeGlobalSegmentCount(
           .stream()
           .filter(
               segment -> {
+                RingRange firstRange = segment.getTokenRanges().get(0);
```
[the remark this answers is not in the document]
adejanovski: Each segment has a "master" range that allows to maintain compatibility with the current design where one segment equals one token range.

src/server/src/main/java/io/cassandrareaper/service/SegmentGenerator.java
adejanovski: The method to trigger subrange repairs evolved in 2.2 to accept multiple token ranges. Before that, it would only accept a single range.

Most clusters are running with 256 vnodes, and it is still the official recommendation. Changing the number of vnodes is a complex operation as it requires to create a new DC and decommission the old one, or re-create the cluster from scratch so I do not expect this to change in a near future. 
Clusters will be upgraded from 2.x to 3.x/4.x with the same number of vnodes, and I'm not sure how many people are aware of the new token allocation algorithm.

So I strongly believe that token coalescing will remove the overhead of vnodes on repair for most of Reaper users for at least a few years.

If you have a large number of vnodes, there is an overhead for each as segments are processed through a scheduled loop in Reaper. The default loop interval was moved from 30s to 10s lately, which in a 6 nodes cluster using 256 vnodes makes each repair last 4 hours approximately, at best (1536 segments * 10s). Such clusters could easily be repaired using 20 segments per node instead (to prevent overstreaming) which would bring down the minimum repair time to 20 minutes (and even less if density is low enough to use a smaller number of segments per node).

So yes, I think it's a pretty big deal :)

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
adejanovski: Batches were failing consistently on large clusters because of the token ranges set. I decided to process segments that have several token ranges as the only query in the batch to prevent such failures.

src/server/src/main/java/io/cassandrareaper/storage/MemoryStorage.java
adejanovski: That allows the code to be compatible with the new way of storing segments.
Since all token ranges in a segment share the same replicas, working on the master/first one is safe.
Reaper uses the notion of parallel ranges, to separate ranges that can be dealt with concurrently, and if it is safe to run the master range, it is safe to run all other ranges in the segment.

adejanovski:
I strongly disagree on it affecting only 2.2 users.
3.0+ clusters with more than 256 vnodes are still the majority IMHO and I don't expect this to change anytime soon as migration to a lower number of vnodes is an expensive operation that requires deep knowledge.

I placed a precondition in the SegmentGenerator code to ensure that coalescing didn't forget any token range so that repairs are still fully processed. Tests are all passing (unit and integration) so I'm not sure I get your concern here. That would mean we don't trust our tests.

src/server/src/main/java/io/cassandrareaper/service/RepairRunService.java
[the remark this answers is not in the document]
michaelsembwever: could `Segment` have a method that says that intuitively? like `segment.getPrimaryTokenRange()`?

src/server/src/main/java/io/cassandrareaper/service/SegmentGenerator.java
michaelsembwever: in the `supportsSegmentCoalescing(..)` method can we add a check on a system variable as well please, so that this feature can be disabled if need be (in existing installations).

something like…
```
private static final boolean COALESCING_ENABLED = !Boolean.getBoolean(SegmentGenerator.class.getName() + ".coalescing.disabled");

…

  private static boolean supportsSegmentCoalescing(String cassandraVersion) {
    return COALESCING_ENABLED 
        && !(cassandraVersion.startsWith("1.") || cassandraVersion.startsWith("2.0") || cassandraVersion.startsWith("2.1"));
  }
```

src/server/src/main/java/io/cassandrareaper/storage/JsonParseUtils.java
```diff
@@ -0,0 +1,63 @@
...
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.base.Optional;
+import com.google.common.collect.Lists;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public final class JsonParseUtils {
```
michaelsembwever, line 31: do we really need to go through jackson?
i would think the serdes of the list of token ranges to/from a string would be simple enough logic to manually do…?

src/server/src/main/java/io/cassandrareaper/service/SegmentGenerator.java
```diff
@@ -81,76 +87,186 @@ static boolean greaterThanOrEqual(BigInteger big0, BigInteger big1) {
...
+      // verify that the whole range is repaired
+      BigInteger total = BigInteger.ZERO;
+      for (Segment segment : repairSegments) {
+        for (RingRange range : segment.getTokenRanges()) {
+          BigInteger size = range.span(rangeSize);
+          total = total.add(size);
+        }
       }
+      if (!total.equals(rangeSize) && !incrementalRepair) {
+        throw new ReaperException("Not entire ring would get repaired");
```
michaelsembwever: is this really a checked exception? how is the caller expected to deal with it and resume/repeat? how could this happen? maybe it's a IllegalStateException or an AssertionError?

src/server/src/main/java/io/cassandrareaper/service/SegmentGenerator.java
```diff
@@ -81,76 +87,186 @@ static boolean greaterThanOrEqual(BigInteger big0, BigInteger big1) {
...
+            .map(List::size)
+            .reduce((first, second) -> first + second)
+            .orElse(0);
+
+    LOG.debug("Coalesced ranges/Total ranges : {}/{}", coalescedRanges);
+
+    LOG.debug("Total number of ranges : {}", totalRanges);
+
+    // Check that we haven't left any token range outside of the resulting segments
+    Preconditions.checkArgument(
```
michaelsembwever: or `.checkState(..)` ?? (idk. who's at fault here?), or `assert …` ?

also lines 232-239 should be moved into a separate method, as they're only for the precondition call.

for example this would make more sense when/if as an assert…
``` 
assert hasCoalescedAllTokens(coalescedRepairSegments, replicasToRange) 
    : "Number of coalesced tokens doesn't match with the total number of tokens";
```

src/server/src/main/java/io/cassandrareaper/service/RepairRunService.java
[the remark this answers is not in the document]
adejanovski: Yes, I actually have such a method : `segment.getBaseRange()`
I didn't have it on the first iteration of changes and this is a leftover.

I've replaced `segment.getTokenRanges().get(0)` with `segment.getBaseRange()` now.

src/server/src/main/java/io/cassandrareaper/service/SegmentGenerator.java
adejanovski: right, good catch !

src/server/src/main/java/io/cassandrareaper/service/SegmentGenerator.java
adejanovski: I'm not sure it's super necessary, as disabling it is already possible by setting a number of segments that's at least the number of vnodes.

Sure you want this ?

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
adejanovski: It does, otherwise it conflicts with another Builder class.

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
adejanovski: ok, changed to IllegalStateException

src/server/src/test/java/io/cassandrareaper/resources/RepairRunResourceTest.java
adejanovski: ok

src/server/src/main/java/io/cassandrareaper/storage/MemoryStorage.java
adejanovski: same as above, and replaced with `.getBaseRange()`.

src/server/src/main/java/io/cassandrareaper/storage/JsonParseUtils.java
adejanovski: I prefer keeping a JSON representation and it has a low enough overhead in our case.

src/server/src/main/java/io/cassandrareaper/service/SegmentGenerator.java
adejanovski: This is out of the scope of this PR, as this code was already there before.

src/server/src/main/java/io/cassandrareaper/service/SegmentGenerator.java
michaelsembwever: yes please. we learnt this lesson from the Preconditions in segment state.

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
michaelsembwever: can you change the line to 
```
Segment.Builder segmentBuilder = Segment.builder();
```

src/server/src/main/java/io/cassandrareaper/service/SegmentGenerator.java
```diff
@@ -81,76 +87,186 @@ static boolean greaterThanOrEqual(BigInteger big0, BigInteger big1) {
...
+
+    // Check that we haven't left any token range outside of the resulting segments
+    Preconditions.checkState(
+        allTokensHaveBeenCoalesced(coalescedRepairSegments, replicasToRange),
+        "Number of coalesced tokens doesn't match with the total number of tokens");
+
+    return coalescedRepairSegments;
+  }
+
+  private boolean allTokensHaveBeenCoalesced(
```
michaelsembwever: can (and should) be a `private static` method.

src/server/src/main/java/io/cassandrareaper/service/SegmentGenerator.java
michaelsembwever: addressed here: https://github.com/thelastpickle/cassandra-reaper/pull/373/files#diff-282828adcfe6dcc488e490223b8f0a4bR231

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
michaelsembwever: addressed here: https://github.com/thelastpickle/cassandra-reaper/pull/373/files#diff-395e506278d41989792e2eb5585367c4R421

adejanovski:
Hey @michaelsembwever, I made all requested changes.
If you have no more I suggest we merge this PR as it's been opened for a while now and I'm eager to get it to master.

Thanks !

src/server/src/main/java/io/cassandrareaper/service/SegmentGenerator.java
```diff
@@ -29,6 +35,8 @@
 final class SegmentGenerator {
 
   private static final Logger LOG = LoggerFactory.getLogger(SegmentGenerator.class);
+  private static final boolean COALESCING_DISABLED =
+      Boolean.getBoolean("disable.tokenrange.coalescing");
```
michaelsembwever: nit-pick: the operator should carry. (otherwise it gives a false impression that arguments/parameters are being carried.) this applies generally to your patches.

michaelsembwever: nit-pick: i'd favour prefixing system variables with an appropriate application label so to provide some context/taxonomy.
for example ` SegmentGenerator.class.getName() + ".coalescing.disabled"` 

(system variables can come from anywhere, just not the application, so it nice to isolate them appropriately, especially a flag that's not to be generally announced.)

michaelsembwever (approved):
a couple of small nit-picking comments added, but otherwise looks good.

PR #380, opened 2018-03-25
Fix remote jmx access when DatacenterAvailability is not ALL

kousin197804:
getRemoteNodeMetrics gets metrics of remote node via JMX even the DatacenterAvailability is not ALL.
When JMX on remote nodes is not accessible, users will get many error logs like below

`ERROR  [2018-03-22 06:36:50,689] [pool-8-thread-7] i.c.j.JmxConnectionFactory - Failed creating a new JMX connection to 100.73.69.88 
java.lang.RuntimeException: io.cassandrareaper.ReaperException: Failure when establishing JMX connection to 100.73.69.88:7199
	at io.cassandrareaper.jmx.JmxConnectionFactory$JmxConnectionProvider.apply(JmxConnectionFactory.java:201)
	at java.util.concurrent.ConcurrentHashMap.computeIfAbsent(ConcurrentHashMap.java:1660)
	at io.cassandrareaper.jmx.JmxConnectionFactory.connect(JmxConnectionFactory.java:91)
	at io.cassandrareaper.jmx.JmxConnectionFactory.connect(JmxConnectionFactory.java:113)
	at io.cassandrareaper.service.SegmentRunner.lambda$getNodeMetrics$3(SegmentRunner.java:579)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)`

When DatacenterAvailability is not ALL (EACH or LOCAL), we should assume that JMX on remotes nodes are not accessible. Therefore I changed the logic in getRemoteNodeMetrics to get metrics via storage when DatacenterAvailability is not ALL.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
+++ b/src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
@@ -582,36 +582,32 @@ final class SegmentRunner implements RepairStatusHandler, Runnable {
   Callable<Pair<String, Optional<NodeMetrics>>> getNodeMetrics(String node, String localDc, String nodeDc) {
 
     return () -> {
-      try {
-        JmxProxy nodeProxy =
-            context.jmxConnectionFactory.connect(
-                Node.builder().withClusterName(clusterName).withHostname(node).build(),
-                context.config.getJmxConnectionTimeoutInSeconds());
+      LOG.debug("getMetricsForHost {} / {} / {}", node, localDc, nodeDc);
 
-        NodeMetrics metrics =
-            NodeMetrics.builder()
-                .withNode(node)
-                .withDatacenter(nodeDc)
-                .withCluster(nodeProxy.getClusterName())
-                .withPendingCompactions(nodeProxy.getPendingCompactions())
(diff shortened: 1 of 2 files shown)


adejanovski (changes requested):
Hi @kousin197804, 

it would indeed trigger less exceptions this way, and the local Reaper instance could still provide the metrics itself through the metrics request system if it can access the node.

Your code currently does not match our style requirements and I cannot compile it (nor Travis it seems as your PR didn't trigger any job).

Could you fix the errors reported by Maven and push your code again ?
Would it be possible as well to squash both commits into a single one ?

Thanks !

kousin197804:
Hi @adejanovski 

I have fixed the style and squashed the commits. The Travis build has passed.

Thanks!

adejanovski (approved):
LGTM. 

Thanks !

PR #383, opened 2018-03-27
Gracefully degrade on 2.0 and prior for snapshot listing.

adejanovski:
Snapshot listing through JMX was introduced in 2.1 and prior versions can only take snapshots.
This commit prevents from having lots of stack traces in the logs when using the snapshot features in C* <= 2.0 and display a proper message in the UI.


michaelsembwever (dismissed):
needs to show green CI tests

adejanovski:
@michaelsembwever : all good now

src/server/src/main/java/io/cassandrareaper/jmx/JmxProxy.java
```diff
@@ -138,7 +138,7 @@ int triggerRepair(
 
   void removeRepairStatusHandler(int commandId);
 
-  List<Snapshot> listSnapshots();
+  List<Snapshot> listSnapshots() throws ReaperException;
```
michaelsembwever: method never throws `ReaperException`

src/server/src/main/java/io/cassandrareaper/jmx/JmxProxyImpl.java
```diff
@@ -1091,8 +1091,16 @@ private void registerConnectionsGauge() {
   }
 
   @Override
-  public List<Snapshot> listSnapshots() {
+  public List<Snapshot> listSnapshots() throws ReaperException {
```
michaelsembwever: method never throws `ReaperException`

michaelsembwever (changes requested):
looks good other than `ReaperException` never actually being thrown.

.circleci/config.yml
```diff
@@ -82,7 +90,6 @@ jobs:
               ccm start -v
               ccm status
               ccm checklogerror
-              ccm node1 nodetool status
               MAVEN_OPTS="-Xmx1g" mvn clean install
               # TODO: parallelise the following
               mvn surefire:test -Dtest=ReaperIT
```
michaelsembwever: what do these changes have to do with this PR?

if it's general CircleCI stability improvements (eg i can see circleci is failing a bit recently) shouldn't it go in as a separate commit and PR?

src/server/src/main/java/io/cassandrareaper/jmx/JmxProxyImpl.java
adejanovski: `versionCompare()` can throw a ReaperException, which is why it's required to either catch or throw it.

adejanovski:
I've made the CircleCI fix part of another PR that was merged to master and I rebased this one on top of it.

src/server/src/main/java/io/cassandrareaper/jmx/JmxProxyImpl.java
michaelsembwever: my bad.

that's confusing when `UnsupportedOperationException` is the exception noticeably being thrown.

And I don't think `versionCompare(..)` should be throwing a `ReaperException`. what has it to do with Reaper? (it even says it's code taken from stackoverflow.) My expectation was that a method like `versionCompare(..)` would only be throwing a IllegalArgumentException.

And why arn't we using `com.datastax.driver.core.VersionNumber` instead?


maybe these questions are out-of-scope for this PR, as this is code that was previously there?

still i would add `UnsupportedOperationException`, as it plays an important role in the api (regardless of whether it's unchecked).
and in a separate commit remove `throws ReaperException` and replace `versionCompare(..)` with that from the cassandra driver.

adejanovski:
@michaelsembwever : I made the changes and the CircleCI tests will pass once #389 will be merged and I'll rebase this branch on master.

src/server/src/main/java/io/cassandrareaper/jmx/JmxProxyImpl.java
```diff
@@ -945,61 +943,20 @@ public void close() {
...
-        // e.g. "1.2.3" = "1.2.3" or "1.2.3" < "1.2.3.4"
-        return Integer.signum(parts1.length - parts2.length);
-      }
-    } catch (RuntimeException ex) {
-      LOG.error("failed comparing strings for versions: '{}' '{}'", str1, str2);
-      throw new ReaperException(ex);
-    }
+  static Integer versionCompare(String str1, String str2) {
+    VersionNumber version1 = com.datastax.driver.core.VersionNumber.parse(str1);
+    VersionNumber version2 = com.datastax.driver.core.VersionNumber.parse(str2);
```
michaelsembwever: trivial: not too sure why the fully qualified classes here when it's already imported.

michaelsembwever (approved):
gold: thanks for putting the VersionNumber change in. hope you agree it feels good to remove so much code!

src/server/src/main/java/io/cassandrareaper/jmx/JmxProxyImpl.java
adejanovski: My bad, I'll fix this before merging.
And yes, feels great to remove the boilerplate :)

Issue #385, opened 2018-03-29
Convert cluster names for Graphite ingestion


joaquincasares:
If a cluster name has a `.` in it, like `tlp-monitoring-stack-2.2`, Graphite will split the metric endpoint into two:

* `tlp-monitoring-stack-2`
* `2`

We need to sterilize `.`'s into `_`'s scores or `-`'s.

Same thing for hostnames that are IP addresses. `.`'s should not create new Graphite endpoints.

michaelsembwever:
Addressed in pull request https://github.com/thelastpickle/cassandra-reaper/pull/386

PR #386, opened 2018-03-29
Sanitise cluster and host names going into metric names

michaelsembwever:
ref: Convert cluster names for Graphite ingestion #385 – https://github.com/thelastpickle/cassandra-reaper/issues/385

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/jmx/JmxProxyImpl.java
+++ b/src/server/src/main/java/io/cassandrareaper/jmx/JmxProxyImpl.java
@@ -1075,18 +1075,25 @@ final class JmxProxyImpl implements JmxProxy {
 
   private void registerConnectionsGauge() {
     try {
-      if (!this.metricRegistry
+      if (!metricRegistry
           .getGauges()
           .containsKey(
               MetricRegistry.name(
-                  JmxProxyImpl.class, this.clusterName, this.host, "repairStatusHandlers"))) {
-        this.metricRegistry.register(
+                  JmxProxyImpl.class,
+                  clusterName.replace('.', '-'),
+                  host.replace('.', '-'),
+                  "repairStatusHandlers"))) {
+
+        metricRegistry.register(
             MetricRegistry.name(
-                JmxProxyImpl.class, this.clusterName, this.host, "repairStatusHandlers"),
-            (Gauge<Integer>) () -> this.repairStatusHandlers.size());
+                JmxProxyImpl.class,
+                clusterName.replace('.', '-'),
+                host.replace('.', '-'),
+                "repairStatusHandlers"),
+            (Gauge<Integer>) () -> repairStatusHandlers.size());
       }
     } catch (IllegalArgumentException e) {
-      LOG.warn("Cannot create connection gauge for node {}", this.host, e);
+      LOG.warn("Cannot create connection gauge for node {}", host, e);
     }
   }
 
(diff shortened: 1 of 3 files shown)


src/server/src/main/java/io/cassandrareaper/jmx/JmxProxyImpl.java
```diff
@@ -1075,18 +1075,25 @@ public String getColumnFamily() {
...
     try {
-      if (!this.metricRegistry
+      if (!metricRegistry
           .getGauges()
           .containsKey(
               MetricRegistry.name(
-                  JmxProxyImpl.class, this.clusterName, this.host, "repairStatusHandlers"))) {
-        this.metricRegistry.register(
+                  JmxProxyImpl.class,
+                  clusterName.replaceAll("[^A-Za-z0-9]", ""),
```
adejanovski: You'll remove any dash or underscore that could exist in the cluster name. This could make it harder for peeps to match their cluster names automatically.

src/server/src/main/java/io/cassandrareaper/jmx/JmxProxyImpl.java
```diff
@@ -1075,18 +1075,25 @@ public String getColumnFamily() {
...
+                  host.replace('.', '-'),
+                  "repairStatusHandlers"))) {
+
+        metricRegistry.register(
             MetricRegistry.name(
-                JmxProxyImpl.class, this.clusterName, this.host, "repairStatusHandlers"),
-            (Gauge<Integer>) () -> this.repairStatusHandlers.size());
+                JmxProxyImpl.class,
+                clusterName.replaceAll("[^A-Za-z0-9]", ""),
```
adejanovski: Same here. 
FYI, Reaper already sanitizes cluster names by default.

src/server/src/main/java/io/cassandrareaper/service/Heart.java
```diff
@@ -115,12 +115,16 @@ private void updateRequestedNodeMetrics() {
                       .forEach(req -> {
 
                         LOG.info("Got metric request for node {} in {}", req.getNode(), req.getCluster());
-                        try (Timer.Context t1 = timer(context, req.getCluster(), req.getNode())) {
+                        try (Timer.Context t1 = timer(
+                            context,
+                            req.getCluster().replaceAll("[^A-Za-z0-9]", ""),
```
adejanovski: Same here :)

adejanovski (dismissed):
I think we shouldn't touch the cluster names in metrics as Reaper already sanitizes them, and that we should use this PR to use the repair unit id instead of the repair run id in the metrics that track repair progress/time since last execution.

src/server/src/main/java/io/cassandrareaper/jmx/JmxProxyImpl.java
michaelsembwever: it was @joaquincasares  that reported that cluster names with `.` could get into the metric names.
so reaper doesn't sanitise them enough…

maybe it should just be `clusterName.replace('.', '-')` instead of `clusterName.replaceAll("[^A-Za-z0-9]", "")`

adejanovski: yes it would be better IMHO

joaquincasares:
Using the latest image of `thelastpickle/cassandra-reaper:latest-cassandra-storage-support-2_0`, I still see at least one issue where the name of the cluster has a `.`:

screen shot 2018-04-03 at 7 57 57 pm

michaelsembwever:
> Using the latest image of thelastpickle/cassandra-reaper:latest-cassandra-storage-support-2_0, I still see at least one issue where the name of the cluster has a .:

Has been addressed in https://github.com/thelastpickle/cassandra-reaper/commit/d7689fa7860f65064a13283a14a948991e6cb1b6

PR #391, opened 2018-03-30
Update the list of nodes for a cluster at the start of each repair run.

adejanovski:
This commit also allows to update the node list easily with the addCluster operation.
If a cluster is added but already exists, its node list will be updated with the current list of live nodes.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
+++ b/src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
@@ -184,10 +184,26 @@ public final class ClusterResource {
     }
     Optional<Cluster> existingCluster = context.storage.getCluster(newCluster.getName());
     if (existingCluster.isPresent()) {
-      LOG.info("cluster already stored with this name: {}", existingCluster);
-      return Response.status(403)
-          .entity(String.format("cluster \"%s\" already exists", existingCluster.get().getName()))
-          .build();
+      LOG.info(
+          "Cluster already stored with this name: {}. Trying to updating the node list.",
+          existingCluster.get().getName());
+      Cluster updatedCluster = updateClusterSeeds(existingCluster.get(), seedHost.get());
+      if (updatedCluster.getSeedHosts().equals(existingCluster.get().getSeedHosts())) {
+        LOG.info("Node list of cluster {} is already up to date.", existingCluster.get().getName());
+        return Response.notModified()
+            .entity(
+                String.format(
+                    "Topology hasn't changed in cluster %s. No update was performed.",
+                    existingCluster.get().getName()))
+            .build();
+      } else {
+        LOG.info(
+            "Node list of cluster {} has been updated with the current topology.",
+            existingCluster.get().getName());
+        return Response.ok()
+            .entity(String.format("Updated cluster %s node list", existingCluster.get().getName()))
+            .build();
+      }
     } else {
       LOG.info("creating new cluster based on given seed host: {}", newCluster.getName());
       context.storage.addCluster(newCluster);
@@ -264,6 +280,7 @@ public final class ClusterResource {
     return new Cluster(clusterName.get(), partitioner.get(), seedHostsFinal);
   }
 
+  @Deprecated
   @PUT
   @Path("/{cluster_name}")
   public Response modifyClusterSeed(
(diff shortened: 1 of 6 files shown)


src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
```diff
@@ -279,45 +296,69 @@ public Response modifyClusterSeed(
...
+   * Updates the list of nodes of a cluster based on the current topology.
+   *
+   * @param cluster the Cluster object we intend to update
+   * @param seedHosts a list of hosts to connect to in the cluster
+   * @return the updated cluster object with a refreshed seed list
+   * @throws ReaperException Any runtime exception that could be triggered
+   */
+  private Cluster updateClusterSeeds(Cluster cluster, Optional<String> seedHosts)
+      throws ReaperException {
+    Set<String> newSeeds = parseSeedHosts(seedHosts.get());
```
Bj0rnen: This seems to assume that seedHosts isn't absent (and will fail if it is). Why are we taking an Optional? For stronger type safety, the responsibility of unwrapping should be in the caller of this function, both of which already do an `.isPresent()` check.

Bj0rnen (changes requested):

src/server/src/test/java/io/cassandrareaper/acceptance/BasicSteps.java
```diff
@@ -292,22 +292,27 @@ public void an_add_cluster_request_is_made_to_reaper() throws Throwable {
...
+                            Response.Status.NOT_MODIFIED.getStatusCode()))
+                    .contains(response.getStatus());
+
+                if (Response.Status.CREATED.getStatusCode() == response.getStatus()) {
+                  responseData.append(response.readEntity(String.class));
+                  Map<String, Object> cluster =
+                      SimpleReaperClient.parseClusterStatusJSON(responseData.toString());
+                  TestContext.TEST_CLUSTER = (String) cluster.get("name");
+                }
+              });
```
michaelsembwever, line 315: what changed here?

adejanovski: The FORBIDDEN status code was traded for a NOT_MODIFIED one as subsequent attempts to create a cluster will update its seed node list (and if it's the same as before, it will return NOT_MODIFIED).

src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
adejanovski: That's right, I'll fix this.

Bj0rnen (approved):

src/server/src/test/java/io/cassandrareaper/acceptance/BasicSteps.java
michaelsembwever: `NOT_MODIFIED` isn't mean to be used in this situation. It is only for responding to conditional-GETs. 

This has been addressed in https://github.com/thelastpickle/cassandra-reaper/pull/401

src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
```diff
@@ -184,10 +184,26 @@ public Response addCluster(
...
-      return Response.status(403)
-          .entity(String.format("cluster \"%s\" already exists", existingCluster.get().getName()))
-          .build();
+      LOG.info(
+          "Cluster already stored with this name: {}. Trying to updating the node list.",
+          existingCluster.get().getName());
+      Cluster updatedCluster = updateClusterSeeds(existingCluster.get(), seedHost.get());
+      if (updatedCluster.getSeedHosts().equals(existingCluster.get().getSeedHosts())) {
+        LOG.info("Node list of cluster {} is already up to date.", existingCluster.get().getName());
+        return Response.notModified()
```
michaelsembwever, line 193: `NOT_MODIFIED` isn't mean to be used in this situation. It is only for responding to conditional-GETs.

This has been addressed in #401

src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
```diff
@@ -184,10 +184,26 @@ public Response addCluster(
...
+                    "Topology hasn't changed in cluster %s. No update was performed.",
+                    existingCluster.get().getName()))
+            .build();
+      } else {
+        LOG.info(
+            "Node list of cluster {} has been updated with the current topology.",
+            existingCluster.get().getName());
+        return Response.ok()
+            .entity(String.format("Updated cluster %s node list", existingCluster.get().getName()))
+            .build();
```
michaelsembwever, line 205: it's odd behaviour that this rest endpoint can return 200, but in one situation return a string message, and in another situation return the ClusterStatus json…

This has been addressed in #401

src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
```diff
@@ -264,6 +280,7 @@ private Cluster createClusterWithSeedHost(String seedHostInput) throws ReaperExc
     return new Cluster(clusterName.get(), partitioner.get(), seedHostsFinal);
   }
 
+  @Deprecated
```
michaelsembwever, line 283: `PUT` and `POST` endpoints are not the same thing, so this shouldn't be deprecated.

The PUT endpoint offers a number of valid differences from enforcing a safety check that the seed must belong to the named cluster, to the presumption of idempotence.

This has been addressed in #401

michaelsembwever:
I added a number of dawdle duckling comments here that I completely missed when first reviewing this, my bad.

Issue #369, opened 2018-03-19
Add Stream Tracking


rzvoncek:
This issue tracks the work related to adding the capability of tracking Cassandra streams to Reaper.

A draft of this functionality has been proposed in #392. Based on the review of @adejanovski and @michaelsembwever, here are the needed improvements:
* The newly added POJOs must truly be POJOs:
  * `FileProgress` should be make private.
  * Any method that is not a getter/setter should go to a factory class:
    * Most notably `Stream::newStream` and `Stream::streamId`.
    * But also `<add|update|finalize>Stream()` in `StreamSession`.
    * Fields and classes should be made final wherever possible.
    * Guava precondition checks should be placed where appropriate, notably for `StreamSession` fields.
* Streams from all C* versions should be tracked.
  * In the MVP, streams are tracked only in 2.2.9 and later.
* Persistent stream notification listeners.
  * Subscription to stream messages should got to `JmxFactory`. The current placement in `StreamManager` doesn't re-establish the listener when reaper is restarted.
* Make stream tracking configurable.
  * It should be possible to configure things like number of threads handling stream notifications and cache retention via Reaper's YAML.
* `StreamBuilder` should be package protected.
* The `@Path` in cluster resource may not need the `/cluster/` part.
* Use `@Alpha` for anything that's not tested
* Handle lost notifications.
  * In addition to listening to notifications, Reaper should sometimes also pull stream info to ensure it has correct stream information.

Once the above is done, this should happen too:
* The stream info should be stored in a distributed storage so that a multi-instance Reaper deployments offer consistent streaming info.

rzvoncek:
Done in #465.

PR #392, opened 2018-04-03
Reaper #369 Backend for tracking streams

rzvoncek:
Still work in progress 

Sample payload: https://gist.github.com/rzvoncek/fcf0ec4f570af3cfb9794afc846ee496

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/core/Stream.java
+++ b/src/server/src/main/java/io/cassandrareaper/core/Stream.java
@@ -1,0 +1,289 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.cassandrareaper.core;
+
+import java.util.Map;
+
+
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+import com.google.common.collect.Maps;
+import org.apache.cassandra.streaming.ProgressInfo;
+import org.apache.cassandra.streaming.SessionInfo;
+
+/**
+ * This class describes the state of a file exchange between two Cassandra nodes.
+ */
+public final class Stream {
+
+  /**
+   * Node on the _local_ side of the stream. The JMX notification came from this node.
+   */
+  private String host;
+
+  /**
+   * Node on the _remote_ side of the stream.
+   */
+  private String peer;
+
+  /**
+   * Total number of bytes to transfer from host to peer.
+   */
+  private long sizeToSend;
+
+  /**
+   * Total number of bytes to transfer from peer to host.
+   */
+  private long sizeToReceive;
+
+  /**
+   * Per-file progress from host to peer.
+   */
+  private Map<String, FileProgress> progressSent;
+
+  /**
+   * Per-file progress from peer to host.
+   */
+  private Map<String, FileProgress> progressReceived;
+
+  /**
+   * Timestamp of the last JMX notification used to update this stream.
+   */
+  private long lastUpdated;
+
+  /**
+   * Indicates if the stream has completed.
+   */
+  private boolean completed;
+
+  /**
+   * Indicates if the stream has completed successfully.
+   */
+  private boolean success;
+
+  /**
+   * Number of bytes transferred from host to peer. Needed only to make the JSON automagic see this field.
+   */
+  private long sizeSent;
+
+  /**
+   * Number of bytes transferred from peer to host. Needed only to make the JSON automagic see this field.
+   */
+  private long sizeReceived;
+
+  private Stream(Builder builder) {
+    this.host = builder.host;
+    this.peer = builder.peer;
+    this.sizeToSend = builder.sizeToSend;
+    this.sizeToReceive = builder.sizeToReceive;
+    this.progressSent = builder.progressSent;
+    this.progressReceived = builder.progressReceived;
+    this.lastUpdated = builder.lastUpdated;
+    this.completed = builder.completed;
+    this.success = builder.success;
+  }
+
+  public String getHost() {
+    return host;
+  }
+
+  public String getPeer() {
+    return peer;
+  }
+
+  public long getSizeToSend() {
+    return sizeToSend;
+  }
+
+  public long getSizeToReceive() {
+    return sizeToReceive;
+  }
+
+  /**
+   * Computes number of bytes currently sent from host to peer.
+   * @return sum of bytes sent across all files in this stream.
+   */
+  public long getSizeSent() {
+    return this.progressSent.values().stream()
+        .map(fileProgress -> fileProgress.currentBytes)
+        .mapToLong(Long::longValue).sum();
+  }
+
+  /**
+   * Computes the number of bytes currently received by host from peer.
+   * @return sum of bytes received across all files in this stream.
+   */
+  public long getSizeReceived() {
+    return this.progressReceived.values().stream()
+        .map(fileProgress -> fileProgress.currentBytes)
+        .mapToLong(Long::longValue).sum();
+  }
+
+  public Map<String, FileProgress> getProgressReceived() {
+    return progressReceived;
+  }
+
+  public Map<String, FileProgress> getProgressSent() {
+    return progressSent;
+  }
+
+  public long getLastUpdated() {
+    return lastUpdated;
+  }
+
+  public boolean getCompleted() {
+    return completed;
+  }
+
+  public boolean getSuccess() {
+    return success;
+  }
+
+  public void setSizeSent(ProgressInfo progressInfo) {
+    progressSent.compute(progressInfo.fileName, (fileName, oldProgress) -> toFileProgress(progressInfo));
+  }
+
+  public void setSizeReceived(ProgressInfo progressInfo) {
+    progressReceived.compute(progressInfo.fileName, (fileName, oldProgress) -> toFileProgress(progressInfo));
+  }
+
+  public void setLastUpdated(long lastUpdated) {
+    this.lastUpdated = lastUpdated;
+  }
+
+  public void setCompleted(boolean completed) {
+    this.completed = completed;
+  }
+
+  public void setSuccess(boolean success) {
+    this.success = success;
+  }
+
+  public String streamId() {
+    return streamId(this.host, this.peer);
+  }
+
+  public static String streamId(String host, String peer) {
+    return String.format("%s->%s", host, peer.replaceAll("/", ""));
+  }
+
+  public String toString() {
+    return String.format(
+        "Stream(host=%s, peer=%s, toSend=%d, sent=%d, toReceive=%d, received=%d, updated=%d, completed=%b",
+        host, peer, sizeToSend, getSizeSent(), sizeToReceive, getSizeReceived(), lastUpdated, completed);
+  }
+
+  public static Stream.Builder builder() {
+    return new Stream.Builder();
+  }
+
+  @JsonPOJOBuilder(buildMethodName = "build", withPrefix = "with")
+  public static final class Builder {
+    private String host;
+    private String peer;
+    private long sizeToSend;
+    private long sizeToReceive;
+    private Map<String, FileProgress> progressSent;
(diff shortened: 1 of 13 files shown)


src/server/src/main/java/io/cassandrareaper/core/Stream.java
```diff
@@ -0,0 +1,289 @@
...
+      return this;
+    }
+
+    public Stream build() {
+      return new Stream(this);
+    }
+
+  }
+
+  public static class FileProgress {
```
adejanovski, line 256: If the class is public I think it should live in its own file instead so that Stream.java only contains the Stream POJO.

src/server/src/main/java/io/cassandrareaper/core/Stream.java
```diff
@@ -0,0 +1,289 @@
...
+      this.totalBytes = totalBytes;
+      this.currentBytes = currentBytes;
+    }
+  }
+
+  private static FileProgress toFileProgress(ProgressInfo pi) {
+    return new FileProgress(pi.totalBytes, pi.currentBytes);
+  }
+
+  public static Stream newStream(String host, SessionInfo sessionInfo, long timeStamp) {
```
adejanovski, line 270: `newStream` should not be part of this class file as it's a factory method to create Stream objects. The Stream POJO should only contain getters/setters and the builder class.
You could create a StreamFactory class that would contain this method (and the others we could need in the future).

src/server/src/main/java/io/cassandrareaper/core/StreamSession.java
```diff
@@ -0,0 +1,184 @@
...
+   *
+   * <p>If the notification comes with old timestamps, then no update happens.
+   *
+   * @param host originating the notification
+   * @param peer the second node participating in the stream
+   * @param success of the streaming session between the two nodes
+   * @param timeStamp of the JMX notification
+   * @return itself with stream finalized
+   */
+  public StreamSession finalizeStream(String host, String peer, Boolean success, long timeStamp) {
```
adejanovski, line 139: `addStream()`, `updateStream()`  and `finalizeStream()` should be taken out of this class to comply with the design of POJO classes. Maybe those should be part of the StreamManager class instead, reworked by taking a StreamSession object as additional argument.

src/server/src/main/java/io/cassandrareaper/jmx/JmxProxyImpl.java
```diff
@@ -852,6 +871,11 @@ public void handleNotification(Notification notification, Object handback) {
     if (("progress").equals(type)) {
       processNewApiNotification(notification);
     }
+
+    if (type.startsWith("org.apache.cassandra.streaming")) {
```
adejanovski, line 875: Is this true from 1.2 to 3.11.2 ? (just asking, I don't have the answer :) )

src/server/src/main/java/io/cassandrareaper/service/StreamManager.java
```diff
@@ -0,0 +1,204 @@
...
+
+/**
+ * This class implements the functionality needed for receiving,
+ * handling and serving streaming-related notifications.
+ */
+public final class StreamManager implements StreamStatusHandler {
+
+  private static final Logger LOG = LoggerFactory.getLogger(StreamManager.class);
+  private final AppContext context;
+  private final ExecutorService executor = Executors.newFixedThreadPool(5);
```
adejanovski, line 54: I know I've done things this way in the SnapshotManager, but I'm considering the option of making the thread pool size configurable in the yaml file.
It could be necessary to speed things up in large clusters.

What do you think ?

src/server/src/main/java/io/cassandrareaper/service/StreamManager.java
```diff
@@ -0,0 +1,204 @@
...
+          .map(node -> connectJmxTask(node))
+          .collect(Collectors.toList());
+
+      List<JmxProxy> nodeProxies = Lists.newArrayList();
+      List<Future<JmxProxy>> futures = executor.invokeAll(connectionTasks);
+      for (Future<JmxProxy> future : futures) {
+        nodeProxies.add(future.get());
+      }
+
+      nodeProxies.forEach(nodeJmxProxy -> nodeJmxProxy.addStreamStatusHandler(this));
```
adejanovski, line 86: How do you handle the case where a connection is lost to a specific node ? Will the handler be recreated when we create a new connection to this node ?

adejanovski: I guess the best way to do this would be to subscribe in the JmxConnectionFactory instead.

src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
```diff
@@ -192,6 +192,8 @@ public Response addCluster(
       LOG.info("creating new cluster based on given seed host: {}", newCluster.getName());
       context.storage.addCluster(newCluster);
 
+      context.streamManager.addListeners(newCluster);
```
adejanovski, line 195: This should not be necessary if the listeners are added in the JmxConnectionFactory, which would cover all cases of JMX connection.

adejanovski (changes requested):
I have some minor concerns that should be easy to fix. 
I think though that we should have some double checking mechanism to handle lost notifications. If Reaper missed one (for whatever reason), we should verify on the node what's the current status of each supposedly running stream. This way we'll use the subscription to avoid scanning the whole cluster, but would query specific nodes for updated status.

rzvoncek:
Thanks @adejanovski for the comments. I'll work on addressing the points one by one.

src/server/src/main/java/io/cassandrareaper/core/Stream.java
rzvoncek: @adejanovski would i be acceptable to make the class private? It is pretty small and mean to hold two longs.

src/server/src/main/java/io/cassandrareaper/service/StreamManager.java
rzvoncek: I agree with making this configurable. While at it, we can also allow configuring other things, such as size and retention times of the cache holding the stream info.

src/server/src/main/java/io/cassandrareaper/core/Stream.java
```diff
@@ -0,0 +1,289 @@
...
+    this.success = success;
+  }
+
+  public String streamId() {
+    return streamId(this.host, this.peer);
+  }
+
+  public static String streamId(String host, String peer) {
+    return String.format("%s->%s", host, peer.replaceAll("/", ""));
+  }
```
michaelsembwever, line 179: should go to the `StreamFactory` class @adejanovski suggests

src/server/src/main/java/io/cassandrareaper/core/Stream.java
```diff
@@ -0,0 +1,289 @@
...
+
+/**
+ * This class describes the state of a file exchange between two Cassandra nodes.
+ */
+public final class Stream {
+
+  /**
+   * Node on the _local_ side of the stream. The JMX notification came from this node.
+   */
+  private String host;
```
michaelsembwever, line 33: a number of these fields can be made final

src/server/src/main/java/io/cassandrareaper/core/StreamSession.java
```diff
@@ -0,0 +1,184 @@
...
+import org.apache.cassandra.streaming.SessionInfo;
+
+
+/**
+ * This class represents a Cassandra streaming session.
+ *
+ * <p>Streaming session is identified by an UUID. A streaming session includes a collection of bi-directional data
+ * exchanges between two nodes.
+ */
+public class StreamSession {
```
michaelsembwever, line 31: can be `final`?

src/server/src/main/java/io/cassandrareaper/core/StreamSession.java
```diff
@@ -0,0 +1,184 @@
...
+   */
+  private String planId;
+
+  /**
+   * The streams belonging to this session.
+   *
+   * <p>Stored in a map keyed by {@link Stream#streamId()} to allow fast lookup of streams needed
+   * to handle stream updates.
+   */
+  private Map<String, Stream> streams;
```
michaelsembwever, line 44: can be `final`

src/server/src/main/java/io/cassandrareaper/core/StreamSession.java
```diff
@@ -0,0 +1,184 @@
...
+      return this;
+    }
+
+    public Builder withNoStreams() {
+      this.streams = Maps.newConcurrentMap();
+      return this;
+    }
+
+    public StreamSession build() {
+      return new StreamSession(this);
```
michaelsembwever, line 172: Both `planId` and `streams` could be null.
I don't think you'd want that. And if so, guava precondition checks should be in place here.

src/server/src/main/java/io/cassandrareaper/core/Stream.java
```diff
@@ -0,0 +1,289 @@
...
+
+  public String toString() {
+    return String.format(
+        "Stream(host=%s, peer=%s, toSend=%d, sent=%d, toReceive=%d, received=%d, updated=%d, completed=%b",
+        host, peer, sizeToSend, getSizeSent(), sizeToReceive, getSizeReceived(), lastUpdated, completed);
+  }
+
+  public static Stream.Builder builder() {
+    return new Stream.Builder();
+  }
```
michaelsembwever, line 189: presuming `StreamFactory` was in the same package, this method and the following Builder class should be package-protected.

src/server/src/main/java/io/cassandrareaper/resources/StreamResource.java
```diff
@@ -0,0 +1,77 @@
...
+          .listStreams(Node.builder().withClusterName(clusterName).withHostname(host).build());
+      return Response.ok().entity(streams).build();
+    } catch (ReaperException e) {
+      LOG.error(e.getMessage(), e);
+      return Response.status(500).entity(e.getMessage()).build();
+    }
+  }
+
+  @GET
+  @Path("/cluster/{clusterName}")
```
michaelsembwever, line 65: why not just `@Path("/{clusterName}")`

src/server/src/main/java/io/cassandrareaper/storage/StreamCache.java
```diff
@@ -0,0 +1,146 @@
...
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
+import com.google.common.collect.Maps;
+import org.apache.cassandra.streaming.ProgressInfo;
+import org.apache.cassandra.streaming.SessionInfo;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class StreamCache {
```
michaelsembwever, line 38: can be `final`?

src/server/src/main/java/io/cassandrareaper/service/StreamManager.java
michaelsembwever: i would think one thread per registered cluster would be enough?

src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
michaelsembwever: yes. and this only works if a cluster is registered first and in the same process. The reaper process can be restarted frequently and the stream manager listener will never be re-connected.

src/server/src/main/java/io/cassandrareaper/service/StreamManager.java
```diff
@@ -0,0 +1,204 @@
...
+/**
+ * This class implements the functionality needed for receiving,
+ * handling and serving streaming-related notifications.
+ */
+public final class StreamManager implements StreamStatusHandler {
+
+  private static final Logger LOG = LoggerFactory.getLogger(StreamManager.class);
+  private final AppContext context;
+  private final ExecutorService executor = Executors.newFixedThreadPool(5);
+  private final StreamCache streamSessions = new StreamCache();
```
michaelsembwever, line 55: so if i understand this correct…

the streaming overview presented by reaper is solely in-memory, and limited to the past 24 hours.
this means that different reaper instances can present different views of streaming plans.

it's great that an incremental solution is being rolled out, but still i think this deserves some design docs, and to be clearly marked `@Alpha` of its new classes.

src/server/src/main/java/io/cassandrareaper/service/StreamManager.java
michaelsembwever: this design needs to be concurrent-safe.
but it is not: 
 - pojos should be immutable (progress components separated into separate immutable pojos), 
 - only concurrent collections used (can be enforced by private instantiation)

a number of places could do with guava precondition checks, eg checking against nulls as so forth, eg ensuring concurrency state guarantees.

rzvoncek:
Closing in favour of #465.

Issue #398, opened 2018-04-10
"given schedule_trigger_time is in the past"


mrwulf:
When creating a new schedule the validation seems to be against a different time zone. Initially I thought this was an issue with the reaper container (we're running reaper in kubernetes) having the wrong timezone, but that doesn't seem to have fixed the validation. It would be nice if (when creating a new schedule) the start time was initialized to the current system time and/or if the start time is in the past the schedule just immediately goes into effect.

adejanovski:
All times are using UTC in the server while the frontend uses no specific timezone, which creates the (crappy) behavior you're observing.
The simplest way to deal with this is indeed to allow schedules that start in the past and let Reaper trigger a new run straight away.
I've always been very frustrated with this as well :D

amorton:
grooming: @adejanovski has code  on his laptop that will send the timezone from the UI to the server so we can sort out the correct UTC time in the back end. Will also immediately start a scheduled repair that starts in the past

PR #399, opened 2018-04-11
Fix duration issues in the UI

### LINKED ISSUE #397: cassandra storage: start_time and end_time appear to be wrong
Hi,

since I've migrated to the cassandra storage I noticed that the ETA reported in the UI is always wrong, it's always "a few seconds ago".

Looking at the JSON returned by `/repair_run` the field `estimated_time_of_arrival` is null even on RUNNING repairs. Here's a copy of the JSON of one repair:

    {
      "blacklisted_tables": [],
      "cause": "scheduled run (schedule id 34a06a50-39a7-11e8-996d-7d32f3609d07)",
      "cluster_name": "cluster",
      "column_families": [],
      "creation_time": "2018-04-08T14:37:20Z",
      "datacenters": [],
      "duration": "0 seconds",
      "end_time": "2018-04-08T14:37:20Z",
      "estimated_time_of_arrival": null,
      "id": "58178b00-3b3a-11e8-996d-7d32f3609d07",
      "incremental_repair": false,
      "intensity": 0.30000001192092896,
      "keyspace_name": "myks",
      "last_event": "Postponed due to affected hosts already doing repairs",
      "nodes": [],
      "owner": "auto-scheduling",
      "pause_time": "2018-04-10T15:56:22Z",
      "repair_parallelism": "PARALLEL",
      "segments_repaired": 958,
      "start_time": "2018-04-08T14:37:20Z",
      "state": "RUNNING",
      "total_segments": 4096
    }

I added some log and it seems the ETA is set to null because start_time and end_time are both not null _and_ end_time is after start_time, so it looks like reaper thinks the repair is done ?

These are the values with the milliseconds:

```start time: 2018-04-08T14:37:20.735Z, end time: 2018-04-08T14:37:20.963Z```

adejanovski:
fixes #397 

Timezone shenanigans between the server and the UI can lead to inconsistent display of ETA or duration.
The repair run status is now shipping with a currentTime value that is generated on the server side to match the startTime/endTime/pauseTime timezones and allow consistent diffs to be made.


mrwulf:
Any chance this will fix #398 too?

adejanovski:
@mrwulf : actually not, as the problem in #398 is that the UI sends times in your local Timezone, but the server uses UTC. I'll see what I can do for that.

PR #400, opened 2018-04-11
Try to show IPv6 endpoints in UI correctly

dormidon:
Hi!
We've tried to use the reaper for Cassandra IPv6 nodes and encountered a problem while viewing these nodes in UI. It looks like this:

While for IPv4 nodes endpoint name field is viewed correctly.

The problem rises from impossibility to parse `endpoint` field of `EndpointState` class from IPv6 address. 
The PR suggests a way to resolve it.


adejanovski:
LGTM, thanks !

dormidon:
Thanks a lot for merge!

PR #401, opened 2018-04-15
RESTful ClusterResource 

michaelsembwever:
Improve REST API to ClusterResource mutations: 

 - idempotence on POST /cluster?seedHost=xxx
    (created 201 if created, ok 200 if node list updated, or no-content 204 if not modified),
 - undeprecate PUT /cluster/<cluster-name>?seedHost=xxx
    it is not the same rest endpoint as the POST,
 - DELETE /cluster/<cluster-name> returns accepted 202,
 - update spreaper to follow to such location headers whenever specifiec,
 - simplify code in ClusterResource by:
   -- delegating PUT/POST endpoints to the new private method `addOrUpdateCluster(..)`,
   -- adding the private method `connectAny(..)`,
   -- using the safer `uriInfo.getBaseUriBuilder()` builder to generate endpoint locations,
 - in unit tests:
   -- consistently use jetty's `HttpStatus` when checking response codes, for readability,
   -- consistently use `org.junit.Assert.assert*(..)` methods,
   -- use correct uri base,
   -- add separate tests for PUT versus POST variants,
 - in integration tests:
   -- manually redirect to GET location, first asserting PUT/POST contains no response body,
   -- update rest endpoint to update repair run's state, so to avoid the 303 redirect on the legacy deprecated endpoint

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
+++ b/src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
@@ -39,6 +39,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.stream.Collectors;
 
+import javax.annotation.Nullable;
 import javax.ws.rs.DELETE;
 import javax.ws.rs.GET;
 import javax.ws.rs.POST;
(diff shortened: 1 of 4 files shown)


michaelsembwever:
work in progress fixing the integration tests…

michaelsembwever:
> work in progress fixing the integration tests…

fixed.

src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
```diff
@@ -159,161 +145,117 @@ public Response getClusterTables(@PathParam("cluster_name") String clusterName)
...
+
+        LOG.info(msg);
+        return Response.status(Response.Status.BAD_REQUEST).entity(msg).build();
+      }
+      Optional<Cluster> existingCluster = context.storage.getCluster(cluster.getName());
+      URI location = uriInfo.getBaseUriBuilder().path("cluster").path(cluster.getName()).build();
+      if (existingCluster.isPresent()) {
+        LOG.debug("Attempting updating nodelist for cluster {}", existingCluster.get().getName());
+
+        cluster = updateClusterSeeds(existingCluster.get(), seedHost.get());
```
ossarga: Perhaps add a comment here saying something like.

> In the case where the cluster is already managed by reaper and no change is made on this call, return 204. Otherwise, return 200 to say something was updated.

src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
```diff
@@ -159,161 +145,117 @@ public Response getClusterTables(@PathParam("cluster_name") String clusterName)
...
+        if (cluster.getSeedHosts().equals(existingCluster.get().getSeedHosts())) {
+          LOG.debug("Nodelist of cluster {} is already up to date.", existingCluster.get().getName());
+          return Response.noContent().location(location).build();
+        } else {
+          LOG.info("Nodelist of cluster {} updated", existingCluster.get().getName());
+          return Response.ok().location(location).build();
+        }
+      } else {
+        LOG.info("creating new cluster based on given seed host: {}", cluster.getName());
```
ossarga: Nit pick: Check to see if all the capitalisation in the log messages is consistent.

src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
```diff
@@ -322,88 +264,59 @@ public Response modifyClusterSeed(
...
    */
   @DELETE
   @Path("/{cluster_name}")
   public Response deleteCluster(
-      @PathParam("cluster_name") String clusterName)
-      throws ReaperException {
+      @PathParam("cluster_name") String clusterName) {
 
-    LOG.info("delete cluster called with clusterName: {}", clusterName);
+    LOG.info("delete cluster {}", clusterName);
```
ossarga, line 303: Nit pick: Perhaps change this to say something like:

> "Deleting cluster {}"

ossarga (approved):
Looks good to me.

src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
michaelsembwever: the verb isn't appropriate as that isn't yet confirmed.

PR #404, opened 2018-04-18
RESTful RepairScheduleResource

michaelsembwever:
- on POST: detect existing schedules and return no-content 204 if matching, otherwise conflict 409,
 - on PUT (to modify state): return no-content 204 instead of not-modified 304, as the latter is for conditional-GETs,
 - DELETE /cluster/ returns accepted 202 (instead of ok 200), and returns conflict 409 instead of forbidden 403,
 - expose RepairScheduleService.conflictingRepairSchedule(..) so we're explicitly handling that response rather than catching a ReaperException,
 - improve RepairScheduleService.isConflictingSchedules(..) to thoroughly check for overlapping tables between repairs,
 - simplify RepairUnitService by
  -- using RepairUnit.Builder for parameters
  -- externalise the check against cassandra-2.0 and incremental repairs out to RepairRunResource.checkRequestForAddRepair(..)
 - simplify code by using primitive boolean for `incrementalRepair` parameters/variables,
 - simplify IStorage.getRepairUnit(..) to use RepairUnit.Builder for parameters,
 - in integration tests ensure DELETEs don't contain a response body, and PUTs and POSTs are bounced to their GET location before validating expected response body

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/core/RepairUnit.java
+++ b/src/server/src/main/java/io/cassandrareaper/core/RepairUnit.java
@@ -14,6 +14,7 @@
 
 package io.cassandrareaper.core;
 
+import java.util.Objects;
 import java.util.Set;
 import java.util.UUID;
 
@@ -23,7 +24,7 @@ public final class RepairUnit {
   private final String clusterName;
   private final String keyspaceName;
   private final Set<String> columnFamilies;
-  private final Boolean incrementalRepair;
+  private final boolean incrementalRepair;
   private final Set<String> nodes;
   private final Set<String> datacenters;
   private final Set<String> blacklistedTables;
@@ -55,7 +56,7 @@ public final class RepairUnit {
     return columnFamilies;
   }
 
-  public Boolean getIncrementalRepair() {
+  public boolean getIncrementalRepair() {
     return incrementalRepair;
   }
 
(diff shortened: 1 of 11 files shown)


src/server/src/main/java/io/cassandrareaper/service/RepairScheduleService.java
```diff
@@ -77,52 +107,41 @@ public RepairSchedule storeNewRepairSchedule(
             segmentCountPerNode);
 
     scheduleBuilder.owner(owner);
+    return context.storage.addRepairSchedule(scheduleBuilder);
+  }
 
-    Collection<RepairSchedule> repairSchedules = context.storage
-        .getRepairSchedulesForClusterAndKeyspace(repairUnit.getClusterName(), repairUnit.getKeyspaceName());
+  private boolean isConflictingSchedules(Cluster cluster, RepairUnit newUnit, RepairUnit existingUnit) {
```
ossarga: Might be worth adding a comment outlining the steps to check for overlapping tables in a schedule.

src/server/src/main/java/io/cassandrareaper/core/RepairUnit.java
```diff
@@ -115,5 +116,42 @@ private Builder(RepairUnit original) {
     public RepairUnit build(UUID id) {
       return new RepairUnit(this, id);
     }
+
+    @Override
+    public int hashCode() {
```
ossarga: Need a comment here about why 7 and 59 are significant. It may even be worth assigning them to a `static final int` variable.

PR #409, opened 2018-04-28
RESTful RepairRunResource & SnapshotResource

michaelsembwever:
This pull request completes the work on improving the REST API to Reaper, from https://github.com/thelastpickle/cassandra-reaper/pull/404 and https://github.com/thelastpickle/cassandra-reaper/pull/401 , bringing the whole REST API to a more cohesive and consistent place.

 - on POST/PUT: detect existing schedules and return no-content 204 if matching, otherwise conflict 409,
 - DELETE  returns accepted 202 (instead of ok 200), return conflict 409 instead of 403 forbidden,
 - DELETEs on snapshots may return 404 not_found if already visibly deleted, otherwise 202 accepted,
 - JmxProxy's clearSnapshot() methods no longer throw ReaperException, and the unused clearAllSnapshots() methods removed,
 - the AssertionError thrown by StorageService.clearSnapshot(..) when the snapshot can't be found is now logged as an error without throwing the ReaperException,
 - clean code applied a bit eagerly to SnapshotResource and SnapshotManager,

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/jmx/JmxProxyImpl.java
+++ b/src/server/src/main/java/io/cassandrareaper/jmx/JmxProxyImpl.java
@@ -979,37 +979,28 @@ final class JmxProxyImpl implements JmxProxy {
   }
 
   @Override
-  public void clearSnapshot(String repairId, String keyspaceName) throws ReaperException {
-    if (repairId == null || ("").equals(repairId)) {
+  public void clearSnapshot(String repairId, String keyspaceName) {
+    if (null == repairId || repairId.isEmpty()) {
       // Passing in null or empty string will clear all snapshots on the hos
       throw new IllegalArgumentException("repairId cannot be null or empty string");
     }
     try {
       ((StorageServiceMBean) ssProxy).clearSnapshot(repairId, keyspaceName);
-    } catch (IOException e) {
-      throw new ReaperException(e);
+    } catch (AssertionError | IOException e) {
+      LOG.error("failed to clear snapshot " + repairId + " in keyspace " + keyspaceName, e);
     }
   }
 
   @Override
-  public void clearSnapshot(String snapshotName) throws ReaperException {
-    if (snapshotName == null || ("").equals(snapshotName)) {
+  public void clearSnapshot(String snapshotName) {
+    if (null == snapshotName || snapshotName.isEmpty()) {
       // Passing in null or empty string will clear all snapshots on the hos
       throw new IllegalArgumentException("snapshotName cannot be null or empty string");
     }
     try {
       ((StorageServiceMBean) ssProxy).clearSnapshot(snapshotName);
-    } catch (IOException e) {
-      throw new ReaperException(e);
-    }
-  }
-
-  @Override
-  public void clearAllSnapshots() throws ReaperException {
-    try {
-      ((StorageServiceMBean) ssProxy).clearSnapshot("");
-    } catch (IOException e) {
-      throw new ReaperException(e);
+    } catch (AssertionError | IOException e) {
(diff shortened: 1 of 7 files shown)


michaelsembwever:
>  - clean code applied a bit eagerly to SnapshotResource and SnapshotManager,

I've separated this out into a separate commit so to make the PR easier to review.
the first commit contains the actual code changes required.
the second commit contains some code cleansing changes.

src/server/src/main/java/io/cassandrareaper/jmx/JmxProxyImpl.java
```diff
@@ -979,37 +979,28 @@ static Integer versionCompare(String str1, String str2) {
...
+    if (null == repairId || repairId.isEmpty()) {
       // Passing in null or empty string will clear all snapshots on the hos
       throw new IllegalArgumentException("repairId cannot be null or empty string");
     }
     try {
       ((StorageServiceMBean) ssProxy).clearSnapshot(repairId, keyspaceName);
-    } catch (IOException e) {
-      throw new ReaperException(e);
+    } catch (AssertionError | IOException e) {
+      LOG.error("failed to clear snapshot " + repairId + " in keyspace " + keyspaceName, e);
```
adejanovski, line 990: Why are we swallowing the exception ? 
The UI will never know that the call failed and it will be required to dig in the logs to see what happened.

src/server/src/main/java/io/cassandrareaper/jmx/JmxProxyImpl.java
```diff
@@ -979,37 +979,28 @@ static Integer versionCompare(String str1, String str2) {
...
-  }
-
-  @Override
-  public void clearAllSnapshots() throws ReaperException {
-    try {
-      ((StorageServiceMBean) ssProxy).clearSnapshot("");
-    } catch (IOException e) {
-      throw new ReaperException(e);
+    } catch (AssertionError | IOException e) {
+      LOG.error("failed to clear snapshot " + snapshotName, e);
```
adejanovski, line 1003: Same here about exception not being thrown.

adejanovski (changes requested):
I think clear snapshot exceptions should be rethrown so that the UI can report them.

Also I don't see updates to spreaper nor the UI, and changing the REST API should have a direct impact on both. 
Did you run extensive testing on both to make sure nothing was broken there ?

src/server/src/main/java/io/cassandrareaper/jmx/JmxProxyImpl.java
michaelsembwever: See the description in the commit message:
> the AssertionError thrown by StorageService.clearSnapshot(..) when the snapshot can't be found is now logged as an error without throwing the ReaperException,

Clearing the snapshot only fails (here) if the snapshot has since been deleted.
The UI does not need to provide an error message, the user requested the snapshot to be deleted and the snapshot has been deleted.
That's the rationale for not throwing the exception: it's an error that can be logged and kept internal, of no concern to the user.

michaelsembwever:
> Also I don't see updates to spreaper nor the UI, and changing the REST API should have a direct impact on both.

No impacts on spreaper or UI. (Which helps illustrates the correctness of this work.)

src/server/src/main/java/io/cassandrareaper/jmx/JmxProxyImpl.java
michaelsembwever: same as https://github.com/thelastpickle/cassandra-reaper/pull/409/files#r187540930

adejanovski (approved):

PR #412, opened 2018-05-01
Avoid blocking the jmx broadcasting thread, offload work to a separate thread immediately

michaelsembwever:
As is documented here: https://docs.oracle.com/cd/E28280_01/web.1111/e13728/notifications.htm#JMXCU214


adejanovski (approved):
LGTM

PR #415, opened 2018-05-12
Always build the npm and bower UI components

michaelsembwever:
Consists of three commits:

**1. Always build the npm and bower UI components**  
 - Remove the `build-ui` profile
 - Do not store generated artifacts in version control (remove `src/server/src/main/resources/assets`)
 - Do not build npm and bower a second time with the maven-source-plugin:attach-sources goal (by not forking it)
 - Update docs to remove reference of `-Pbuild-ui`
 - Update travis and circleci scripts to install the required npm and bower tools

**2. Remove the unused `src/docs/public/` directory**
 - generated docs go to (and are served from) the `docs/` directory, the removed directory does not appear to be used

**3. Regenerate (and publish) the website with above changes.**

PR #417, opened 2018-05-14
[src/docs] Update Bootstrap and Fontawesome to use latest CDN  over HTTPS.

terrcin:
I use the EFF [HTTPS Everywhere browser plugin](https://www.eff.org/https-everywhere) which is forcing bootstrap and fontawesome to load over HTTPS instead of HTTP and breaking the public website for me. The layout and image sizes are all distorted making the site unusable.

This is happening because a different file is loading over HTTPS (rather than HTTP) which doesn't match the `integrity` hash in the script and link tag and so the browser blocks it being loaded.

To resolve this I have updated src/docs to explicitly use HTTPS as recommended by [bootstrapcdn](https://www.bootstrapcdn.com/legacy/bootstrap/).

Note: this PR is based off https://github.com/thelastpickle/cassandra-reaper/pull/415 to prevent conflicts.

### DIFF
--- a/src/docs/themes/hugo-universal-theme/layouts/partials/head.html
+++ b/src/docs/themes/hugo-universal-theme/layouts/partials/head.html
@@ -9,13 +9,13 @@
   {{ if .Keywords }}
   <meta name="keywords" content="{{ delimit .Keywords ", " }}">
   {{ else if .Site.Params.DefaultKeywords }}
-  <meta name="keywords" content="{{ delimit .Site.Params.DefaultKeywords ", " }}">	
+  <meta name="keywords" content="{{ delimit .Site.Params.DefaultKeywords ", " }}">
   {{ end }}
 
   {{ if .Description }}
   <meta name="description" content="{{ .Description }}">
   {{ else if .Site.Params.DefaultDescription }}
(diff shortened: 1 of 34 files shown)


terrcin:
FYI: I'm using a much newer version of Hugo 0.26 => 40.3 and compiling the site seems to have generated files that were not previously there.

src/docs/themes/hugo-universal-theme/layouts/partials/head.html
```diff
@@ -9,22 +9,22 @@
...
   {{ else if .Site.Params.DefaultKeywords }}
-  <meta name="keywords" content="{{ delimit .Site.Params.DefaultKeywords ", " }}">	
+  <meta name="keywords" content="{{ delimit .Site.Params.DefaultKeywords ", " }}">
   {{ end }}
 
   {{ if .Description }}
   <meta name="description" content="{{ .Description }}">
   {{ else if .Site.Params.DefaultDescription }}
-  <meta name="description" content="{{ .Site.Params.DefaultDescription }}">	
+  <meta name="description" content="{{ .Site.Params.DefaultDescription }}">
```
terrcin, line 18: My editor auto removed trailing whitespace.

PR #419, opened 2018-05-15
Update REST API and Docker vars docs

adejanovski:


michaelsembwever:
Repeating the suggestion to that given to @terrcin that the docs regeneration should be a separate commit (for history sake and for making the review simpler to review).

This should also probably be based on top of @terrcin's https://github.com/thelastpickle/cassandra-reaper/pull/419

PR #424, opened 2018-05-18
CircleCI – parametrise the build, using yaml references

michaelsembwever:
- workflow and parameters of the different supported versions
 - comment out Cassandra-4.0 backend testing until that is supported

Example of the workflow that tests different versions of Cassandra:
screen shot 2018-05-18 at 15 28 22


adejanovski (approved):
Looks great !

PR #425, opened 2018-05-25
Preconfigured clusters

benbromhead:
Allow reaper to setup clusters that are predefined in the yaml file.

If we already know the cluster we want to manage repairs on, we can preconfigure in the yaml file and don't have to resort to scripts that poll/wait for reaper to come up before configuring via API. 

Note there are two style checks failing. One appears unrelated to this change, one appears to be silly... LMK if these need to be addressed.

### DIFF
--- a/src/docs/content/docs/configuration/reaper_specific.md
+++ b/src/docs/content/docs/configuration/reaper_specific.md
@@ -392,4 +392,15 @@ accessControl:
     iniConfigs: ["file:/path/to/shiro.ini"]
 ```
 
+### `preconfiguredClusters`
 
+Type: *Object*
+
+Optional mapping of Cluster names to contact point that reaper will automatically add as clusters being repaired. Useful with autoScheduling.
+
+```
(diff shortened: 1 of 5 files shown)


michaelsembwever:
> …moving away from an immutable deployment philosophy we are trying to stick with.

We do a wrapped docker image that has a entrypoint.sh which does the poll and configure. Isn't that immutable, if not pretty damn close?

> Would it be cleaner to remove the explicit per cluster config in the yaml file and just have a boolean flag to manage repairs on the Reaper Cassandra storage backend?

Yes, I'd take that @benbromhead 
@adejanovski what are your thoughts on it?

adejanovski:
Hi @benbromhead ,

I'd be in favor of implementing boolean flag technique as well.

Thanks !

adejanovski:
This will be done automatically in sidecar mode (currently available in master).
Closing this PR for now, feel free to reopen if needed.

PR #427, opened 2018-05-28
Add automated purge functions

adejanovski:
This commit creates 2 new config elements to control the depth of the repair run history as old runs get read often, especially through the UI.
Runs can be purged based on a time based depth (purge everything that's more than 30 days old for example) and/or based on a number of runs to keep per repair unit (keep the last 10 runs for each repair unit, for example).
The purge manager is scheduled to run every hour (the value is hardcoded).

### DIFF
--- a/src/packaging/resource/cassandra-reaper-cassandra-ssl.yaml
+++ b/src/packaging/resource/cassandra-reaper-cassandra-ssl.yaml
@@ -14,6 +14,8 @@ repairManagerSchedulingIntervalSeconds: 10
 activateQueryLogger: false
 jmxConnectionTimeoutInSeconds: 5
 useAddressTranslator: false
+# purgeRecordsAfterInDays: 30
+# numberOfRunsToKeepPerUnit: 10
 
 # datacenterAvailability has three possible values: ALL | LOCAL | EACH
 # the correct value to use depends on whether jmx ports to C* nodes in remote datacenters are accessible
--- a/src/packaging/resource/cassandra-reaper-cassandra.yaml
+++ b/src/packaging/resource/cassandra-reaper-cassandra.yaml
@@ -15,6 +15,8 @@ repairManagerSchedulingIntervalSeconds: 10
 activateQueryLogger: false
 jmxConnectionTimeoutInSeconds: 5
 useAddressTranslator: false
+# purgeRecordsAfterInDays: 30
+# numberOfRunsToKeepPerUnit: 10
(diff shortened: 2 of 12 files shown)


michaelsembwever:
It seems a shame that we're not instead using Cassandra's TTL at write-time.
After all, this would be very simple to implement: just pick up `ReaperApplicationConfiguration.purgeRecordsAfterInDays` and apply it in `CassandraStorage.addRepairRun(..)`

Which raises the following questions:
 - to which backends do we know this is a problem with?
 - how and where do we know it's a problem? and 
 - to what extent does it impact the different backends?

If the identified problem is this:
> …as old runs get read often, especially through the UI

then why is the UI (which limits the results usually to 10) having to read all the repair runs? For example in the Cassandra backend the repair run IDs are time-based UUIDs, so it is possible to paginate/limit the queries more effectively.

Maybe the purge timer and deletes should only operate on postgresql. Maybe we don't need it at all?

adejanovski:
I disagree and think we need to have a common way of handling purge between backends. 
On top of it, TTL will create more tombstones than partition deletes and leave us transiently with partial rows that will fail some preconditions we have on the segments.
And we couldn't achieve the "keep the last 10 runs for each repair unit" feature which is my favorite over a time based purge.

We know it's a problem because Reaper has to cycle through all repair runs on a regular basis, and when there are lots of repairs/segments it leads to a lot of pressure on the cluster.
Then you can only delete a cluster when it has no repair run in history, which makes it challenging since we don't offer a bulk delete feature for repair runs.

It'll be an easy win for users who want to automate the cleanup of the database.

adejanovski:
@michaelsembwever, I think it is due diligence to provide automated purge functions and cleanup our database.
Otherwise it relies on the user to delete history through the UI (which is utterly painful) or by accessing the database directly (dangerous).
Since we commit to support both Postgres and Cassandra, I think we must try to use the same mechanism as much as possible to ease up maintenance.
The purge will also limit the amount of data that is read from the database. While Postgres handles this gracefully using indexes, we do not have the same ability to filter rows server side with Cassandra. This leads to reading a lot of data on a regular basis just to filter it client side.
While I reckon there's been some improvements lately and more should be done, it's not a trivial task with C*, and the purge will make Reaper lighter for clusters used as backend.

The size on disk of the Reaper db is really not a problem as you've already calculated in the past, and LCS does a good job at dealing with overwrites.

PR #431, opened 2018-05-31
Avoid the NPE in Heart.beat when shutting down

michaelsembwever:
Often when running the integration tests, during shutdown, a `NullPointerException` is thrown from the inner forkJoin in the `updateRequestedNodeMetrics()` method.

This fix allows the RepairManager to be closed, which in turns closes the Heart instance before the shutdown otherwise initiates.


adejanovski (approved):
LGTM

PR #433, opened 2018-06-01
Add repair threads support

adejanovski:
Since Cassandra 2.2, repairs are multithreaded in order to process several token ranges concurrently and speed up the process.
While it would have made no sense to support this before, token range coalescing made it necessary to maximize performance on Cassandra 2.2.
No more than four threads are authorized by Cassandra. The number of repair threads can be set differently for each repair run/schedule.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
+++ b/src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
@@ -104,20 +104,23 @@ public final class RepairRunResource {
       @QueryParam("incrementalRepair") Optional<String> incrementalRepairStr,
       @QueryParam("nodes") Optional<String> nodesToRepairParam,
       @QueryParam("datacenters") Optional<String> datacentersToRepairParam,
-      @QueryParam("blacklistedTables") Optional<String> blacklistedTableNamesParam) {
+      @QueryParam("blacklistedTables") Optional<String> blacklistedTableNamesParam,
+      @QueryParam("repairThreadCount") Optional<Integer> repairThreadCountParam) {
 
     try {
-      final Response possibleFailedResponse = RepairRunResource.checkRequestForAddRepair(
-          context,
-          clusterName,
-          keyspace,
-          owner,
-          segmentCountPerNode,
-          repairParallelism,
-          intensityStr,
-          incrementalRepairStr,
-          nodesToRepairParam,
-          datacentersToRepairParam);
+      final Response possibleFailedResponse
+          = RepairRunResource.checkRequestForAddRepair(
+              context,
+              clusterName,
+              keyspace,
+              owner,
+              segmentCountPerNode,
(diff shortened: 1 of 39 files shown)


src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
```diff
@@ -104,20 +104,23 @@ public Response addRepairRun(
...
-          clusterName,
-          keyspace,
-          owner,
-          segmentCountPerNode,
-          repairParallelism,
-          intensityStr,
-          incrementalRepairStr,
-          nodesToRepairParam,
-          datacentersToRepairParam);
+      final Response possibleFailedResponse =
```
michaelsembwever: nit-pick: the operator shall carry over to the new line. 

ie
```
     = RepairRunResource.checkRequestForAddRepair(
              context,
              clusterName,
              keyspace,
              owner,
              segmentCountPerNode,
              repairParallelism,
              intensityStr,
              incrementalRepairStr,
              nodesToRepairParam,
              datacentersToRepairParam,
              repairThreadCountParam);
```
it's an axiom to make the distinction between assignment, method calls, and arguments, clearer.

Doing this you'd also reduce the patch size by 50+ lines.

src/server/src/main/resources/db/cassandra/013_multithreaded_repair.cql
```diff
@@ -0,0 +1,4 @@
+--
+-- Upgrade to allow segments to contain several token ranges
+
+ALTER TABLE repair_unit_v1 ADD repair_thread_count int;
```
michaelsembwever, line 4: Even on Cassandra-2.2 (CQL-3.3.1) the `tinyint` data type was available.
As I don't think we're going to say thread count 128+ maybe tinyint is the appropriate type to use here.

michaelsembwever:
Solid 💪 

Apart from the suggestion on replacing `int` with `tinyint`, it's a +1 from me.
(Standardising how we wrap lines would also be fantastic too :)

michaelsembwever (approved):

src/server/src/main/resources/db/cassandra/013_multithreaded_repair.cql
adejanovski: The database is already very small and the repair unit table is one with just a few rows in it, so I don't feel a strong need to use `tinyint` even if the value is planned to be very small (limited to 4 actually).

michaelsembwever: Sure, but given the effort to make that change why not be more accurate? It's not just about the disk space used, but the intuitiveness that the correct type implies in its schema.

adejanovski:
I'll skip switching to `tinyint` as the driver requests a codec for `tinyint`<--> `Integer` which is not worth adding, sorry.

Merging.

Issue #435, opened 2018-06-05
Optimize repair runs scans with the Cassandra backend


adejanovski:
Let's list all places where we perform repair run scans in Reaper and see what's the best way to optimize the reads.

The plan is to avoid performing full scans when we can avoid them.

amorton:
grooming: needs further definition of "optimise"

michaelsembwever:
From https://github.com/thelastpickle/cassandra-reaper/pull/427#issuecomment-393712932

enrich the `repair_run_by_cluster` table so that initial fetch of the repair run IDs can be filtered better… (it is after all a bit crazy that we have no notion of distinguishing between current and archived repair runs in the C* backend). For example:
```
CREATE TABLE IF NOT EXISTS repair_run_by_cluster (
  cluster_name text,
  repair_run_state text,
  id           timeuuid,
  PRIMARY KEY(cluster_name, repair_run_state, id)
)
```
(and something similar for `repair_run_by_unit`).

It's not that I'm against the idea of the purge feature, but I am conscious about the practicality of it and the (small) cognitive load it does add to the user and the developer. And above all else would like to see the discussion around this stuff land 'on paper'.

michaelsembwever:
Closing, as #437 is waiting on review now.

PR #437, opened 2018-06-06
Implement limit in `CassandraStorage.getRepairRunsForCluster(cluster, limit)`

michaelsembwever:
As this was previously unimplemented it lead to large scans that performed poorly, and impacted the ability of a Cassandra backend to store repair run history.
This came up in PR#427 where a purge feature that deleted repair run history was deemed to be necessary, and again in issue #435 which presumed the fix would have to be schema changes.
Further investigation reveal it was simply the matter that the IStorage contract was not being fully implemented.

The implementation takes a small step further by also adding a limit to `IStorage.getRepairRunsForCluster(cluster)`
 and narrowing the return type of `IStorage.getRepairRunIdsForCluster(cluster)` from Collection to SortedSet, as this collection now contains time-based UUIDs which can (and should) be returned sorted in reverse chronological order.

References:
 - Optimize repair runs scans with the Cassandra backend :: https://github.com/thelastpickle/cassandra-reaper/issues/435
 - Add automated purge functions :: https://github.com/thelastpickle/cassandra-reaper/pull/427

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
+++ b/src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
@@ -312,7 +312,7 @@ public final class ClusterResource {
           .entity("cluster \"" + clusterName + "\" cannot be deleted, as it has repair schedules")
           .build();
     }
-    if (!context.storage.getRepairRunsForCluster(clusterName).isEmpty()) {
+    if (!context.storage.getRepairRunsForCluster(clusterName, Optional.absent()).isEmpty()) {
       return Response.status(Response.Status.CONFLICT)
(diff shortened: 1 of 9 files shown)


michaelsembwever:
A few extra asserts added to `RepairRunResourceTest`

(Ideally it would be nice to run `RepairRunResourceTest.testAddRepairRun()` against each of the storage backend implementations.)

adejanovski (approved):
We could get into some unexpected situations (from the user point of view) when some repair started earlier end up later than others, as we're sorting on the start time.
But I agree that there's no simple way to achieve a sort by end time in Cassandra easily.

We need to followup with another PR where the UI will send the number of requested runs in the REST calls.

PR #441, opened 2018-06-09
Diagnostic Cassandra events

michaelsembwever:
See [documentation page](https://github.com/thelastpickle/cassandra-reaper/blob/mck/spodkowinski/diag_events_jmx/src/docs/content/docs/usage/cassandra-diagnostics.md) for a more thorough info on set up and usage.

Reference 
 - https://issues.apache.org/jira/browse/CASSANDRA-12944 – Diagnostics Events [Cassandra Server]


src/server/src/main/java/io/cassandrareaper/ReaperApplication.java
```diff
@@ -237,6 +235,22 @@ public void run(ReaperApplicationConfiguration config, Environment environment)
...
+
+    // Enable cross-origin requests for using external GUI applications.
+    if (config.isEnableCrossOrigin() || System.getProperty("enableCrossOrigin") != null) {
+      final FilterRegistration.Dynamic cors
+          = environment.servlets().addFilter("crossOriginRequests", CrossOriginFilter.class);
+      cors.setInitParameter("allowedOrigins", "*");
+      cors.setInitParameter("allowedHeaders", "X-Requested-With,Content-Type,Accept,Origin");
+      cors.setInitParameter("allowedMethods", "OPTIONS,GET,PUT,POST,DELETE,HEAD");
+      cors.addMappingForUrlPatterns(EnumSet.allOf(DispatcherType.class), true, "/*");
+    }
```
michaelsembwever: i need to move the cors stuff (lines 244-252) back to where it came from (it's not part of the patch)

src/server/src/main/java/io/cassandrareaper/resources/DiagEventSubscriptionResource.java
```diff
@@ -0,0 +1,128 @@
...
+
+    LOG.debug("deleting subscription with id: {}", id);
+    DiagEventSubscription subscription = diagEventService.getEventSubscription(id);
+    if (null == subscription) { //@fixme change to catch
+      return Response.status(Response.Status.NOT_FOUND).build();
+    }
+
+    assert id.equals(subscription.getId());
+    if (diagEventService.deleteEventSubscription(id)) {
+      return Response.ok().entity(subscription).build();
```
michaelsembwever: rest commands shouldn't return bodies/entities, they should redirect to GET the resource instead. same applies to the post above. i will fix.

(i thought we have a test that enforced this…?)

src/server/src/main/java/io/cassandrareaper/storage/MemoryStorage.java
```diff
@@ -512,4 +514,49 @@ public Snapshot getSnapshot(String clusterName, String snapshotName) {
...
+  }
+
+  @Override
+  public DiagEventSubscription getEventSubscription(UUID id) {
+    return subscriptionsById.get(id);
+  }
+
+  @Override
+  public DiagEventSubscription addEventSubscription(DiagEventSubscription subscription) {
+    if (subscription.getId() == null) {
```
spodkowinski: Needs to be checked using `isPresent()` instead of `null`

src/server/src/main/java/io/cassandrareaper/ReaperApplication.java
spodkowinski: Not sure if that was intentional or not. But I don't mind if you would.

src/server/src/main/java/io/cassandrareaper/resources/DiagEventSubscriptionResource.java
michaelsembwever: and `observable.js` won't (or can't) follow a redirect to the resource?

michaelsembwever:
Addressed two of the comments, and rebased off master.

michaelsembwever:
Docs page added at https://github.com/thelastpickle/cassandra-reaper/blob/mck/spodkowinski/diag_events_jmx/src/docs/content/docs/usage/cassandra-diagnostics.md (@adejanovski )

michaelsembwever:
@spodkowinski i can no longer get the node multiselect working on the event subscription section. any chance you could help me? i've tried a few things so far without much luck…

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
```diff
@@ -1565,6 +1579,60 @@ private static boolean withinRange(RepairSegment segment, Optional<RingRange> ra
...
+        row.getList("events", String.class),
+        row.getBool("export_sse"),
+        row.getString("export_file_logger"),
+        row.getString("export_http_endpoint"));
+  }
+
+  @Override
+  public DiagEventSubscription addEventSubscription(DiagEventSubscription subscription) {
+    session.execute(saveDiagnosticEventPrepStmt.bind(
+        subscription.getId(),
```
adejanovski: `subscription.getId()` is returning an `Optional<UUID>` while the query is expecting a `UUID`.
I've tried to add `.get()` but the id was null when creating new subscriptions. It ended up working after replacing this line with : 

```
        subscription.getId().orElse(UUID.randomUUID()),
```

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
```diff
@@ -1565,6 +1579,60 @@ private static boolean withinRange(RepairSegment segment, Optional<RingRange> ra
...
+        subscription.getExportFileLogger(),
+        subscription.getExportHttpEndpoint()));
+
+    return subscription;
+  }
+
+  @Override
+  public boolean deleteEventSubscription(UUID id) {
+    session.executeAsync(deleteDiagnosticEventPrepStmt.bind(id));
+    return false;
```
adejanovski: This makes every REST call fail with a 500 error when trying to delete existing subscriptions.

adejanovski: Since we need to refresh the UI to get an updated view of the subscriptions, I think we need to use a synchronous query instead, at `LOCAL_QUORUM`.

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
michaelsembwever: Thanks for spotting this.  The transformation was there already in `MemoryStorage`. I pushed it up to `DiagEventSubscriptionService`, and in both storages put int a precondition check.

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
michaelsembwever: Going through the UI does not need strict consistency. (You could just be on a page that lists a subscription that has been deleted elsewhere.)
It doesn't really need to be sync either, durability of the delete is not critical. 

I presume the REST call was failing only because of the `return false`, thanks for fixing that.

michaelsembwever:
Subscribing and live streaming events over the SSE is working again.

I will rebase off master tomorrow… EDIT: has been done.

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
```diff
@@ -1575,6 +1589,62 @@ private static boolean withinRange(RepairSegment segment, Optional<RingRange> ra
...
+        subscription.getExportSse(),
+        subscription.getExportFileLogger(),
+        subscription.getExportHttpEndpoint()));
+
+    return subscription;
+  }
+
+  @Override
+  public boolean deleteEventSubscription(UUID id) {
+    session.executeAsync(deleteDiagnosticEventPrepStmt.bind(id));
```
adejanovski, line 1647: We seem to have lost our conversation around this being async.
I think there's nothing worse than clicking a button that says "Delete" and nothing happens in the UI (and then I have to refresh so that it eventually disappears).
Using a synchronous query with `LOCAL_QUORUM` seems to me like a very reasonable tradeoff as this won't get called at fast pace. The performance gain with running it async is not worth having a UI that doesn't instantly reflects the change (I'm telling this because I've seen that inconsistent behavior in the UI when deleting subscriptions).

Are you ok with making a synchronous call here?

src/ui/app/jsx/event-screen.jsx
```diff
@@ -0,0 +1,201 @@
...
+    if(this.state.activeSubscription) {
+      this.props.unlistenSubscriptionSubject.onNext(this.state.activeSubscription);
+    }
+  },
+
+  render: function() {
+
+    var btnLiveView = <button type="button" className="btn btn-sm btn-block btn-default" disabled="disabled">No active subscriptions for live view</button>
+    if (this.state.activeSubscription) {
+      btnLiveView = <button type="button" className="btn btn-sm btn-block btn-success" onClick={this._onCancelView}>Viewing events for {this.state.activeSubscription.description}</button>
```
adejanovski, line 121: I really don't understand this choice (not meaning to be rude here ;) ).
Nothing indicates that pressing the button will stop the flow of events, and using a green button doesn't show that you're going to stop something.
I really vote in favor of an orange button and a text that says "Press to stop..."

adejanovski (approved):
I'm ok with merging the PR as is it for introducing the feature as a beta.
I think the feature is really made for Cassandra developers and not standard Cassandra operators.
Improvements to make it easier to use : 
- add pretty formatting of the json blobs with the events details (schema changes for example are hard to read)
- add some docs or write a blog post with concrete examples of useful cases where this feature can help

michaelsembwever:
(adding to @adejanovski's list above)

The following tasks will have to be done in subsequent PRs…
- [x] add pretty formatting of the json blobs with the events details (schema changes for example are hard to read),
- [ ] add some docs or write a blog post with concrete examples of useful cases where this feature can help,
- [ ] improve UI around subscribing and unsubscribing live events. the green bar is confusing… (see @alex's input above),
- [ ] take another look at the react code, making any necessary small changes to bring it back to what the standard style and patterns in the codebase are,
- [x] fix rest endpoints to use `location` response header where appropriate (commands don't have entity bodies),
- [ ] hardcode the nodes field to the localhost, in the subscription form, when Reaper is running in sidecar mode,

src/ui/app/jsx/event-screen.jsx
michaelsembwever: UI/UX is tricky.

I never interpreted it as a button. And never read its text as an action, but rather as state.
To toggle it to disable it was intuitive enough. And displayed as a green indicated to me that it was "active". If it were orange on the other hand I might think that there was something wrong and it wasn't properly activated.

This is a beta feature at the moment, and I'm happy to let it in as the contributor designed it, and to come back on it when broader feedback is received on it.
@spodkowinski what are your thoughts on this?

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
michaelsembwever: See https://github.com/thelastpickle/cassandra-reaper/pull/441#discussion_r314040277

The UI should instantly reflect the change regardless, as it should simply delete the html row (rather than refresh the page, or fetch the whole subscription list again and rebuild the table).

Also, the REST endpoint for the DELETE command returns `202 ACCEPTED`, so the client knows the delete is "being processed".

I gave it a test and was unable to reproduce any UI inconsistencies. (Actually i could, the whole list disappears, but leaving that for another day.)

src/server/src/main/java/io/cassandrareaper/resources/DiagEventSubscriptionResource.java
michaelsembwever: This has been fixed/improved. This DELETE rest endpoint now returns `202 ACCEPTED`, without an entity body, like other delete commands in Reaper.

spodkowinski:
Thanks both of you for rebasing and following up on bug fixes. Great to see this merged 🎉

src/ui/app/jsx/event-screen.jsx
spodkowinski: Red elements should point users to issues that need attention. Having the event flow active is the normal state, so we're green. It's also perfectly safe to click the green button. Having the flow paused may be temporary, but should be highlighted as such to get attention, so it's red. You're clicking to get back to normal again (green).

But this is all just my personal, totally unscientific view on this and I'm not considering myself an UX expert.

PR #442, opened 2018-06-11
Improve start time selection for schedules

### LINKED ISSUE #241: Add a "Submit Immediately" button to the schedules page
Alternatively, and probably more work, the "New Repair" form could be copied to the schedule page and a button added to each scheduled repair to pre-fill a new repair submission. For me, just creating a new repair with the same parameters as the scheduled job would be sufficient.

### LINKED ISSUE #398: "given schedule_trigger_time is in the past"

adejanovski:
Fixes #398 
Fixes #241 

This commit eases using the UI to create repair schedules by adjusting the time from the client timezone to UTC (as the server uses UTC).
It also allows activation times in the past, which will trigger a repair immediately.
A "run now" button was added to the UI in order to start a schedule right away by updating the next activation time.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/resources/RepairScheduleResource.java
+++ b/src/server/src/main/java/io/cassandrareaper/resources/RepairScheduleResource.java
@@ -123,10 +123,10 @@ public final class RepairScheduleResource {
       DateTime nextActivation;
       try {
         nextActivation = getNextActivationTime(scheduleTriggerTime);
-        if (nextActivation.isBeforeNow()) {
+        if (nextActivation.isBefore(DateTime.now().minusMinutes(15))) {
           return Response.status(Response.Status.BAD_REQUEST)
-              .entity("given schedule_trigger_time is in the past: "
-                  + RepairRunStatus.dateTimeToIso8601(nextActivation))
+              .entity("given schedule_trigger_time is too far in the past: "
+                      + RepairRunStatus.dateTimeToIso8601(nextActivation))
               .build();
         }
       } catch (IllegalArgumentException ex) {
@@ -384,6 +384,30 @@ public final class RepairScheduleResource {
   }
 
   /**
+   * Force start a repair from a schedule.
+   *
+   * @return detailed information about a repair schedule.
+   */
+  @POST
+  @Path("/start/{id}")
+  public Response startRepairSchedule(@PathParam("id") UUID repairScheduleId) {
+    LOG.debug("start repair_schedule called with: id = {}", repairScheduleId);
+    Optional<RepairSchedule> repairSchedule = context.storage.getRepairSchedule(repairScheduleId);
+    if (repairSchedule.isPresent()) {
+      RepairSchedule newSchedule =
+          new RepairSchedule.Builder(repairSchedule.get())
+              .nextActivation(DateTime.now())
+              .build(repairScheduleId);
+      context.storage.updateRepairSchedule(newSchedule);
+      return Response.ok().entity(getRepairScheduleStatus(newSchedule)).build();
+    } else {
+      return Response.status(404)
+          .entity("repair schedule with id " + repairScheduleId + " doesn't exist")
+          .build();
+    }
+  }
+
+  /**
    * @param clusterName The cluster_name for which the repair schedule belongs to.
    * @return all know repair schedules for a cluster.
    */
@@ -491,7 +515,8 @@ public final class RepairScheduleResource {
         .build();
   }
 
-  private DateTime getNextActivationTime(Optional<String> scheduleTriggerTime) throws IllegalArgumentException {
+  private DateTime getNextActivationTime(Optional<String> scheduleTriggerTime)
+      throws IllegalArgumentException {
     DateTime nextActivation;
     if (scheduleTriggerTime.isPresent()) {
       nextActivation = DateTime.parse(scheduleTriggerTime.get());
@@ -504,5 +529,4 @@ public final class RepairScheduleResource {
     }
     return nextActivation;
   }
-
 }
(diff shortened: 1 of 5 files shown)


michaelsembwever:
This introduces two features. Why wasn't it submitted as two separate pull requests? (At minimum two separate commits)

michaelsembwever:
>This commit eases using the UI to create repair schedules by adjusting the time from the client timezone to UTC (as the server uses UTC).

Was it considered to have clients send timestamps in UTC (opposed to having the server translate timestamps back to UTC).
The clients here are limited: `spreaper` and the UI.

The way `RepairScheduleResource.adjustTimeDifferenceOnNextActivation(..)` works with a delta in milliseconds makes me a bit uneasy. It's a bit of a hack in lei of proper timezone handling (or enforcing).

For example if it's possible to enforce clients to do timestamps in UTC it would be simpler approach because it
 - avoids adding the `clientTime` parameter everywhere,
 - avoids adding the `adjustTimeDifferenceOnNextActivation(..)` method,
 - makes more sense as `spreaper` already documents that the `--schedule-trigger-time` is expected in UTC format,
 - not have to figure out if the trigger time argument was specified by the client in its local TZ or intentionally in UTC (as documented).

Maybe this stuff was considered?

src/server/src/main/java/io/cassandrareaper/resources/RepairScheduleResource.java
```diff
@@ -520,4 +543,28 @@ private DateTime getNextActivationTime(Optional<String> scheduleTriggerTime) thr
...
+   */
+  @VisibleForTesting
+  public static DateTime adjustTimeDifferenceOnNextActivation(
+      DateTime nextActivation, Optional<String> clientTime) {
+    DateTime adjustedNextActivation = nextActivation;
+
+    if (clientTime.isPresent()) {
+      DateTime clientLocalTime = DateTime.parse(clientTime.get());
+      Long timeDifference = DateTime.now().getMillis() - clientLocalTime.getMillis();
+      adjustedNextActivation = adjustedNextActivation.plusMillis(timeDifference.intValue());
```
michaelsembwever: Isn't this making the presumption that the client is sending in the nextActiviation dateTime in its local timezone format?

The API requests that the `scheduleTriggerTime` argument be in UTC, ref https://github.com/thelastpickle/cassandra-reaper/blob/master/src/packaging/bin/spreaper#L236

Now if the client follows the API it breaks.

src/server/src/main/java/io/cassandrareaper/resources/RepairScheduleResource.java
```diff
@@ -506,10 +526,13 @@ public Response deleteRepairSchedule(
...
+  private DateTime getNextActivationTime(
+      Optional<String> scheduleTriggerTime, Optional<String> clientTime)
+      throws IllegalArgumentException {
     DateTime nextActivation;
     if (scheduleTriggerTime.isPresent()) {
       nextActivation = DateTime.parse(scheduleTriggerTime.get());
+      nextActivation = adjustTimeDifferenceOnNextActivation(nextActivation, clientTime);
       LOG.info("first schedule activation will be: {}", RepairRunStatus.dateTimeToIso8601(nextActivation));
     } else {
       nextActivation = DateTime.now().plusDays(1).withTimeAtStartOfDay();
```
michaelsembwever: i think this line deserves a comment, referencing that it's forced to UTC from https://github.com/thelastpickle/cassandra-reaper/blob/master/src/server/src/main/java/io/cassandrareaper/ReaperApplication.java#L142


Although that line in ReaperApplication really needs to be replaced with 
```
    TimeZone.setDefault(TimeZone.getTimeZone("UTC"));
    DateTimeZone.setDefault(DateTimeZone.UTC);```

**Review by michaelsembwever — CHANGES_REQUESTED** [review]

File: src/server/src/main/java/io/cassandrareaper/resources/RepairScheduleResource.java (hunk 2 above)
Comment (michaelsembwever) [reply src/server/src/main/java/io/cassandrareaper/resources/RepairScheduleResource.java]: I would also look at improving that line of code as well
```
nextActivation = DateTime.now().plusDays(1).withTimeAtStartOfDay();
```

What we want is midnight local time, not midnight UTC. (Ideally midnight to the datacenter/region the schedule is for.)
To make that possible we want need to save the original timezone in `ReaperApplication` before forcing it to UTC. The original saved timezone could then be made available via `AppContext`.

**Comment by adejanovski** [comment]
I've squashed and pushed my latest changes. Should be all good now @michaelsembwever .

**Review by michaelsembwever — DISMISSED** [review]

File: src/server/src/main/java/io/cassandrareaper/resources/RepairScheduleResource.java [hunk 3]
```diff
@@ -383,6 +377,30 @@ public Response getRepairSchedule(
...
+  @Path("/start/{id}")
+  public Response startRepairSchedule(@PathParam("id") UUID repairScheduleId) {
+    LOG.debug("start repair_schedule called with: id = {}", repairScheduleId);
+    Optional<RepairSchedule> repairSchedule = context.storage.getRepairSchedule(repairScheduleId);
+    if (repairSchedule.isPresent()) {
+      RepairSchedule newSchedule =
+          new RepairSchedule.Builder(repairSchedule.get())
+              .nextActivation(DateTime.now())
+              .build(repairScheduleId);
+      context.storage.updateRepairSchedule(newSchedule);
```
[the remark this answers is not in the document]
adejanovski: As discussed together, the schedule can already shift if Reaper misses the launch window or if the repair lasts for longer than necessary to match the activation time.
I agree we may want to come up with something more predictable in the future though.

PR #446, opened 2018-06-18
display node metrics

adejanovski:
Adds all the necessary bits to expose REST endpoints to retrieve tpstats, dropped messages and client requests metrics for a specific node and display them in the UI.

reaper-metrics1
reaper-metrics2
reaper-metrics3

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/service/MetricsGrabber.java
+++ b/src/server/src/main/java/io/cassandrareaper/service/MetricsGrabber.java
@@ -1,0 +1,222 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.cassandrareaper.service;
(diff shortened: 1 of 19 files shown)


src/server/src/main/java/io/cassandrareaper/service/MetricsGrabber.java
```diff
@@ -0,0 +1,222 @@
...
+      // We have several metric types that we need to process separately
+      // We'll group on MetricsHistogram::getType in order to generate one histogram per type
+      Map<String, List<JmxStat>> metrics =
+          pool.getValue().stream().collect(Collectors.groupingBy(JmxStat::getName));
+
+      for (Entry<String, List<JmxStat>> metric : metrics.entrySet()) {
+        MetricsHistogram.Builder builder =
+            MetricsHistogram.builder().withName(pool.getKey()).withType(metric.getKey());
+        for (JmxStat stat : metric.getValue()) {
+          if (stat.getAttribute().equals("Count")) {
```
michaelsembwever, line 158: trivial: safer to put the string literal first. eg `"string_literal".equal(…)`

 (it also makes it a lot easier to read, bc alignment)

michaelsembwever (changes requested):
Pretty solid piece of work here. Going to make a lot of users happy!

One small requested change:
 - there's a lot of specific `JMException` exceptions being listed in method signatures and caught in MetricsGrabber. I reckon the list should just be replaced with `JMException`.

It also needs a rebase, but otherwise it's a +1 from me.

michaelsembwever:
> there's a lot of specific JMException exceptions being listed in method signatures and caught in MetricsGrabber. I reckon the list should just be replaced with JMException.

@adejanovski i think you missed that this was also in reference to the JmxProxy method signatures. I've opened a PR in https://github.com/thelastpickle/cassandra-reaper/pull/507

PR #449, opened 2018-06-19
Handle out of order repair notifications

adejanovski:
In the case repair sessions end in a very short time (within a few 10s of milliseconds) we can get out of order notifications from Cassandra :
Instead of `START -> SUCCESS -> COMPLETE`, we get `START -> COMPLETE -> SUCCESS`
This commit handles this case by not signaling the wait condition of the segment unless we have both a `SUCCESS` or an `ERROR` notification plus a `COMPLETE` one.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
+++ b/src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
@@ -149,6 +149,11 @@ final class RepairRunner implements Runnable {
       throw new ReaperException(msg);
     }
 
+    LOG.debug(
+        "Possible parallel repairs : {}",
+        Math.min(
+            ranges.size() / ranges.values().iterator().next().size(),
+            Math.max(1, hostsInRing.keySet().size() / ranges.values().iterator().next().size())));
     return Math.min(
         ranges.size() / ranges.values().iterator().next().size(),
         Math.max(1, hostsInRing.keySet().size() / ranges.values().iterator().next().size()));
@@ -176,6 +181,8 @@ final class RepairRunner implements Runnable {
         new RingRange(
             segments.get((parallelRepairs - 1) * segments.size() / parallelRepairs).getStart(),
             segments.get(0).getStart()));
+
+    LOG.debug("Parallel ranges : {}", parallelRanges);
 
     return parallelRanges;
   }
--- a/src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
+++ b/src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
@@ -96,6 +96,8 @@ final class SegmentRunner implements RepairStatusHandler, Runnable {
   private int commandId;
   private final AtomicBoolean segmentFailed;
   private final UUID leaderElectionId;
+  private final AtomicBoolean successOrFailedNotified = new AtomicBoolean(false);
+  private final AtomicBoolean completeNotified = new AtomicBoolean(false);
 
 
   SegmentRunner(
@@ -808,6 +810,14 @@ final class SegmentRunner implements RepairStatusHandler, Runnable {
                     .withEndTime(DateTime.now())
                     .withId(segmentId)
                     .build());
+
+            successOrFailedNotified.set(true);
+            // Since we can get out of order notifications,
+            // we need to exit if we already got the COMPLETE notification.
+            if (completeNotified.get()) {
+              condition.signalAll();
+              jmxProxy.removeRepairStatusHandler(repairNumber);
+            }
             break;
           }
         } catch (AssertionError er) {
@@ -823,17 +833,29 @@ final class SegmentRunner implements RepairStatusHandler, Runnable {
             segmentId,
             repairNumber);
         failOutsideSynchronizedBlock = true;
+        successOrFailedNotified.set(true);
+        // Since we can get out of order notifications,
+        // we need to exit if we already got the COMPLETE notification.
+        if (completeNotified.get()) {
+          condition.signalAll();
+          jmxProxy.removeRepairStatusHandler(repairNumber);
+        }
         break;
 
       case COMPLETE:
         // This gets called through the JMX proxy at the end
         // regardless of succeeded or failed sessions.
+        // Since we can get out of order notifications,
+        // we won't exit unless we already got a SUCCESS or ERROR notification.
         LOG.debug(
             "repair session finished for segment with id '{}' and repair number '{}'",
             segmentId,
             repairNumber);
-        condition.signalAll();
-        jmxProxy.removeRepairStatusHandler(repairNumber);
+        completeNotified.set(true);
+        if (successOrFailedNotified.get()) {
+          condition.signalAll();
+          jmxProxy.removeRepairStatusHandler(repairNumber);
+        }
         break;
       default:
         LOG.debug(
@@ -894,6 +916,14 @@ final class SegmentRunner implements RepairStatusHandler, Runnable {
                     .withId(segmentId)
                     .build());
 
+            // Since we can get out of order notifications,
+            // we need to exit if we already got the COMPLETE notification.
+            successOrFailedNotified.set(true);
+            if (completeNotified.get()) {
+              condition.signalAll();
+              jmxProxy.removeRepairStatusHandler(repairNumber);
+            }
+
             break;
           }
         } catch (AssertionError er) {
@@ -903,19 +933,33 @@ final class SegmentRunner implements RepairStatusHandler, Runnable {
         break;
 
       case SESSION_FAILED:
-        LOG.warn("repair session failed for segment with id '{}' and repair number '{}'", segmentId, repairNumber);
+        LOG.warn(
+            "repair session failed for segment with id '{}' and repair number '{}'",
+            segmentId,
+            repairNumber);
         failOutsideSynchronizedBlock = true;
+        // Since we can get out of order notifications,
+        // we need to exit if we already got the COMPLETE notification.
+        successOrFailedNotified.set(true);
+        if (completeNotified.get()) {
+          condition.signalAll();
+          jmxProxy.removeRepairStatusHandler(repairNumber);
+        }
         break;
 
       case FINISHED:
         // This gets called through the JMX proxy at the end
         // regardless of succeeded or failed sessions.
+        // Since we can get out of order notifications,
+        // we won't exit unless we already got a SUCCESS or ERROR notification.
         LOG.debug(
             "repair session finished for segment with id '{}' and repair number '{}'",
             segmentId,
             repairNumber);
-        condition.signalAll();
-        jmxProxy.removeRepairStatusHandler(repairNumber);
+        if (successOrFailedNotified.get()) {
+          condition.signalAll();
+          jmxProxy.removeRepairStatusHandler(repairNumber);
+        }
         break;
       default:
         LOG.debug(
--- a/src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
+++ b/src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
@@ -791,6 +791,11 @@ final class SegmentRunner implements RepairStatusHandler, Runnable {
         break;
 
       case SUCCESS:
+
+        Preconditions.checkState(
+            !successOrFailedNotified.get(),
+            "illegal multiple 'SUCCESS' and 'FAILURE', %s:%s", repairRunner.getRepairRunId(), segmentId);
+
         try {
           if (segmentFailed.get()) {
             LOG.debug(
@@ -828,6 +833,11 @@ final class SegmentRunner implements RepairStatusHandler, Runnable {
 
       case ERROR:
       case ABORT:
+
+        Preconditions.checkState(
+            !successOrFailedNotified.get(),
+            "illegal multiple 'SUCCESS' and 'FAILURE', %s:%s", repairRunner.getRepairRunId(), segmentId);
+
         LOG.warn(
             "repair session failed for segment with id '{}' and repair number '{}'",
             segmentId,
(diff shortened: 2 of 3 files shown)


michaelsembwever:
Wanna rebase this off master @adejanovski ?

src/server/src/test/java/io/cassandrareaper/service/SegmentRunnerTest.java
```diff
@@ -475,6 +475,145 @@ protected JmxProxy connect(
     assertEquals(2, storage.getRepairSegment(runId, segmentId).get().getFailCount());
   }
 
+  @Test
+  public void outOfOrderTest() throws InterruptedException, ReaperException, ExecutionException {
```
michaelsembwever: i have verified that this test fails without the patch applied. 👍

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
```diff
@@ -124,6 +127,9 @@
     this.repairUnit = repairUnit;
     this.repairRunner = repairRunner;
     this.segmentFailed = new AtomicBoolean(false);
+    this.gotCompleteNotification = new AtomicBoolean(false);
+    this.gotFailedNotification = new AtomicBoolean(false);
+    this.gotSuccessNotification = new AtomicBoolean(false);
```
michaelsembwever: can we use the following shorter names for these plesae?
{{completeNotified}}, {{failedNotified}}, {{successNotified}}

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
```diff
@@ -96,6 +96,9 @@
   private int commandId;
   private final AtomicBoolean segmentFailed;
   private final UUID leaderElectionId;
+  private final AtomicBoolean gotFailedNotification;
+  private final AtomicBoolean gotSuccessNotification;
+  private final AtomicBoolean gotCompleteNotification;
```
michaelsembwever: can we use just two AtomicBooleans, and use the following shorter names for these please?
`completeNotified`, `failedOrSuccessNotified`

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
michaelsembwever: can we remove these 3 lines please, they are not needed (the fields can be initialised on lines 99-101)

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
```diff
@@ -808,6 +814,14 @@ private boolean handleJmxNotificationForCassandra22(
                     .withEndTime(DateTime.now())
                     .withId(segmentId)
                     .build());
+
+            gotSuccessNotification.set(Boolean.TRUE);
```
michaelsembwever: autobox?
ie `completeNotified.set(true);`

src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
```diff
@@ -149,6 +149,11 @@ static int getPossibleParallelRepairsCount(
       throw new ReaperException(msg);
     }
 
+    LOG.debug(
+        "Possible parallel repairs : {}",
+        Math.min(
+            ranges.size() / ranges.values().iterator().next().size(),
+            Math.max(1, hostsInRing.keySet().size() / ranges.values().iterator().next().size())));
```
michaelsembwever, line 156: maybe wrap this debug statement inside `if (LOG.isDebugEnabled()) {…}`

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
```diff
@@ -894,6 +920,14 @@ private boolean handleJmxNotificationForCassandra21(
                     .withId(segmentId)
                     .build());
 
+            // Since we can get out of order notifications,
+            // we need to exit if we already got the COMPLETE notification.
+            gotSuccessNotification.set(Boolean.TRUE);
+            if (gotCompleteNotification.get()) {
```
michaelsembwever: this will never be true

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
```diff
@@ -903,19 +937,33 @@ private boolean handleJmxNotificationForCassandra21(
...
-        LOG.warn("repair session failed for segment with id '{}' and repair number '{}'", segmentId, repairNumber);
+        LOG.warn(
+            "repair session failed for segment with id '{}' and repair number '{}'",
+            segmentId,
+            repairNumber);
         failOutsideSynchronizedBlock = true;
+        // Since we can get out of order notifications,
+        // we need to exit if we already got the COMPLETE notification.
+        gotFailedNotification.set(Boolean.TRUE);
+        if (gotCompleteNotification.get()) {
```
michaelsembwever: this will never be true

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
```diff
@@ -903,19 +937,33 @@ private boolean handleJmxNotificationForCassandra21(
...
 
       case SESSION_FAILED:
-        LOG.warn("repair session failed for segment with id '{}' and repair number '{}'", segmentId, repairNumber);
+        LOG.warn(
+            "repair session failed for segment with id '{}' and repair number '{}'",
+            segmentId,
+            repairNumber);
         failOutsideSynchronizedBlock = true;
+        // Since we can get out of order notifications,
+        // we need to exit if we already got the COMPLETE notification.
```
michaelsembwever, line 967: can we add an `assert` here that success hasn't already been marked true.

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
```diff
@@ -894,6 +920,14 @@ private boolean handleJmxNotificationForCassandra21(
                     .withId(segmentId)
                     .build());
 
+            // Since we can get out of order notifications,
+            // we need to exit if we already got the COMPLETE notification.
```
michaelsembwever, line 940: can we add an `assert` here that failed hasn't already been marked true.

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
```diff
@@ -823,17 +837,29 @@ private boolean handleJmxNotificationForCassandra22(
             segmentId,
             repairNumber);
         failOutsideSynchronizedBlock = true;
```
michaelsembwever, line 845: can we add an `assert` here that success hasn't already been marked true.

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
```diff
@@ -808,6 +814,14 @@ private boolean handleJmxNotificationForCassandra22(
                     .withEndTime(DateTime.now())
                     .withId(segmentId)
                     .build());
+
```
michaelsembwever, line 818: can we add an `assert` here that failed hasn't already been marked true.

michaelsembwever (changes requested):

src/server/src/test/java/io/cassandrareaper/service/SegmentRunnerTest.java
michaelsembwever: can we get a variations of this test for 
 - Cassandra-2.1 and Cassandra-2.2
 - all different sequences of notifications (out of order `failed` isn't tested)

michaelsembwever (changes requested):

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
adejanovski: Just a habit I've taken... I'll switch back to `true`.

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
adejanovski: ok, but I was just being consistent with the way we deal with `segmentFailed` ;)

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
adejanovski: sure thing

src/server/src/test/java/io/cassandrareaper/service/SegmentRunnerTest.java
adejanovski: cool

src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
adejanovski: Not really concerned with performance here, I'd rather avoid adding lines and conditional blocks if you don't mind.

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
adejanovski: This is precisely what we're trying to deal with. We get cases where `COMPLETE` gets received before `SUCCESS` does.
What makes you think it can never be true ?

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
adejanovski: same here

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
adejanovski: `failedNotified` or `segmentFailed` ?

In any case, what's the point ? There would be a big fat bug in Cassandra if we could get a `SESSION_FAILED` notif and then a `SESSION_SUCCESS`. 
It shouldn't be up to Reaper to assert here, as it's not anymore a matter of race conditions (which Reaper can/should handle).

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
adejanovski: Same as before, I don't see the point.

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
adejanovski: again

adejanovski:
I'll rebase, address some of your comments and add a test for out of order failed notif (and duplicate the tests for 2.1/2.2).

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
michaelsembwever: that's the point of assertions, to assert something we presume is true.
it also serves as documentation (another contributor to reaper may not know that about Cassandra)
(it also good practice to be defensive against external systems)
(it is a use-case where i'd use `assert ` instead of `Preconditions.check…()`)

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
```diff
@@ -96,6 +96,9 @@
   private int commandId;
   private final AtomicBoolean segmentFailed;
   private final UUID leaderElectionId;
+  private final AtomicBoolean failedNotified = new AtomicBoolean(false);
+  private final AtomicBoolean successNotified = new AtomicBoolean(false);
```
michaelsembwever: thoughts about replacing these two above with just `failedOrSuccessNotified` ?

michaelsembwever (dismissed):
Still keen to see those assertions in place.

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
adejanovski: I made this change.

adejanovski:
Still don't see the value of adding more assertions here as we're not checking something that is generated by Reaper's behavior. If we go down that path we'll end up asserting everything that shouldn't be happening.
Let's move on please ?

michaelsembwever:
> Still don't see the value of adding more assertions here as we're not checking something that is generated by Reaper's behaviour.

I see the value, and it's only four lines of code being requested.

Apart from existing reasons given, the code is introducing an adhoc state-machine with the use of multiple `AtomicBoolean`, and i feel it makes sense that we validate the transitions. 
Doing so also documents the state-machine, as has been said already.

> If we go down that path we'll end up asserting everything that shouldn't be happening.

Adding assertions is a good thing, it adds both defensiveness and documentation, both of which are valuable in open-source code where anyone can (and is encouraged to) contribute.
So yes, if the code is ambiguous (or can permit invalid state in any way) then an assertion/precondition should be used.

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
```diff
@@ -791,6 +791,11 @@ private boolean handleJmxNotificationForCassandra22(
         break;
 
       case SUCCESS:
+
+        Preconditions.checkState(
+            !successOrFailedNotified.get(),
+            "multiple notifications of either 'SUCCESS' or 'FAILURE' are not valid");
+
```
michaelsembwever: I would like to update (force push) this commit so that these lines contain the repairId and segmentId (so it's not so cryptic to debug in the event it does happen).
(The proposal is in https://github.com/thelastpickle/cassandra-reaper/commit/1cd25771dda94ffc7dfc908d215a4bd6b5df701d )

@adejanovski am i free to force push that commit to this branch?

adejanovski: @michaelsembwever yes feel free to force push.

michaelsembwever: thanks. done.

michaelsembwever (approved):

PR #461, opened 2018-06-29
Fixes #311.  Allow webui to be aware of a server defined applicationContextPath setting

### LINKED ISSUE #311: base url not retained
Hi, just installed the tool recently to try it out and I have had a few issues so far:
The service runs behind reverse proxies and I had to manually patch the jar/deps.js because the /ping and /cluster endpoint did not take the full path from the url into account:
let's say I redirect my service to : 
http://myproxyurl/cassandra/reaper

I would expect the endpoints to take that full url into account for the ajax requests like:
http://myproxyurl/cassandra/reaper/ping

problem is that it gets redirected to:
http://myproxyurl/ping

I would not want to add all possible urls into the reverse proxies configuration as I have multiple services running on them but would prefer the deps.js to take that into account...

denniskline:
This commit expands URL_PREFIX to accommodate someone defining an applicationContextPath in the config.yaml.  

Previously, webui would make all server requests to "/".  This approach works great for containers or other single-use ip/cname but falls short when there are multiple apps running on the same server fronted by something like apache using a reverse proxy.


adejanovski:
Hi @denniskline,

thanks for sending the PR. I think it would be a good occasion to remove the code duplicates and only rely on the mixin.getUrlPrefix() method everywhere (like it's already done in different places like node-status.jsx for example).

Would you mind refactoring the code in that way ?
I'll test your code then and merge if there's no issue.

Thanks !

denniskline:
You're very welcome @adejanovski.  I'l absolutely make that change.  I debated doing that as well but opt'd for the least invasive change instead.  

Thank you!

denniskline:
@adejanovski please let me know if that last commit is ok or if you like to see other changes.

adejanovski:
Thanks @denniskline,

I just got back from vacation and will take a look ASAP.

src/ui/app/html_template.ejs
```diff
@@ -12,7 +12,11 @@
     <title>Cassandra Reaper<%= htmlWebpackPlugin.options.title %></title>
 
     <!-- <base href="/webui/"> -->
-    <base href="<%= htmlWebpackPlugin.options.baseUrl %>" />
+    <script type="text/javascript">
+        var pathName = document.location.pathname;
+        var contextPath = pathName.substring(0, pathName.indexOf("<%= htmlWebpackPlugin.options.baseUrl %>"));
+        document.write('<base href="' + contextPath + '<%= htmlWebpackPlugin.options.baseUrl %>" />');
+    </script>
```
adejanovski: The contextPath will not be correctly constructed if there's a `/` at the end of the baseUrl, so it breaks a change I've made lately to avoid having a blank page when going to `/webui` instead of `/webui/`.
Here's what I've changed to make it work in all situations : 

```
<script type="text/javascript">
        var pathName = document.location.pathname;
        var baseUrl = "<%= htmlWebpackPlugin.options.baseUrl %>".endsWith('/') ? "<%= htmlWebpackPlugin.options.baseUrl %>".substring(0,"<%= htmlWebpackPlugin.options.baseUrl %>".lastIndexOf('/')):"<%= htmlWebpackPlugin.options.baseUrl %>";
        var contextPath = pathName.substring(0, pathName.indexOf(baseUrl));
        document.write('<base href="' + contextPath + '<%= htmlWebpackPlugin.options.baseUrl %>" />');
    </script>
```

What do you think of it ?

adejanovski (changes requested):
Aside from one small change, I'm happy with the PR.

src/ui/app/html_template.ejs
denniskline: Ahh, sorry about missing that.  Your alteration looks perfect.  Thank you

denniskline:
@adejanovski Changes made, please let me know what you think.

adejanovski:
LGTM @denniskline 

Thanks a lot for you contribution !

adejanovski (approved):

PR #466, opened 2018-07-05
Systemd unit file + post-install script for log dirs.

nsteinmetz:
* Add systemd service file
* Add postinstall script to create log dirs


nsteinmetz:
Small fix on the doc by the way :-)

Issue #457, opened 2018-06-24
repair manager fails with IllegalStateException: endTime can only be set if segment is DONE


vrischmann:
Hi,

I use Cassandra 2.2.12 as storage and today I saw this error in the reaper logs:

    ERROR  [2018-06-24 13:06:09,148] [pool-1-thread-1] i.c.ReaperApplication - Couldn't resume running repair runs
    io.cassandrareaper.ReaperException: java.lang.IllegalStateException: endTime can only be set if segment is DONE
            at io.cassandrareaper.service.RepairManager.resumeRunningRepairRuns(RepairManager.java:103)
            at io.cassandrareaper.ReaperApplication.lambda$run$0(ReaperApplication.java:267)
            at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
            at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:308)
            at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:180)
            at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:294)
            at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
            at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
            at java.lang.Thread.run(Thread.java:748)
    Caused by: java.lang.IllegalStateException: endTime can only be set if segment is DONE
            at com.google.common.base.Preconditions.checkState(Preconditions.java:444)
            at io.cassandrareaper.core.RepairSegment$Builder.build(RepairSegment.java:228)
            at io.cassandrareaper.storage.CassandraStorage.createRepairSegmentFromRow(CassandraStorage.java:813)
            at io.cassandrareaper.storage.CassandraStorage.getSegmentsWithState(CassandraStorage.java:840)
            at io.cassandrareaper.service.RepairManager.lambda$abortAllRunningSegmentsWithNoLeader$0(RepairManager.java:111)
            at java.lang.Iterable.forEach(Iterable.java:75)
            at io.cassandrareaper.service.RepairManager.abortAllRunningSegmentsWithNoLeader(RepairManager.java:109)
            at io.cassandrareaper.service.RepairManager.resumeRunningRepairRuns(RepairManager.java:98)
[...]

PR #476, opened 2018-07-11
Ensure notifications for a given repairNo are processed in order

michaelsembwever:
The first commit is a code cleaning commit.
It
 - refactors all parameter/variable names of `commandId` to `repairNo`, as that is the terminology used by Cassandra
 - throws an exception if a segmentRunner is asked to handle a repairNo that is incorrect.

The second commit fixes both errors below, by providing a single separate threads per SegmentRunner for notifications to be processed on.
These errors were…
 - IllegalStateException: illegal multiple 'SUCCESS' and 'FAILURE'
 - IllegalStateException: endTime can only be set if segment is DONE

ref: https://github.com/thelastpickle/cassandra-reaper/issues/457

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/jmx/JmxProxyImpl.java
+++ b/src/server/src/main/java/io/cassandrareaper/jmx/JmxProxyImpl.java
@@ -123,8 +123,7 @@ final class JmxProxyImpl implements JmxProxy {
   private final String hostBeforeTranslation;
   private final JMXServiceURL jmxUrl;
   private final String clusterName;
-  private final ConcurrentMap<Integer, RepairStatusHandler> repairStatusHandlers =
-      Maps.newConcurrentMap();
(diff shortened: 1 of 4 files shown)


src/server/src/main/java/io/cassandrareaper/jmx/JmxProxyImpl.java
```diff
@@ -846,7 +815,11 @@ public String getAllEndpointsState() {
   @Override
   public void handleNotification(final Notification notification, Object handback) {
     // pass off the work immediately to a separate thread
-    EXECUTOR.submit(() -> {
+    int repairNo = ("repair").equals(notification.getType())
+        ? Integer.parseInt(((String) notification.getSource()).split(":")[1])
+        : ((int[]) notification.getUserData())[0];
+
+    repairStatusExecutors.get(repairNo).submit(() -> {
```
michaelsembwever: this is the guts of it.

adejanovski (approved):
LGTM

PR #484, opened 2018-07-15
Fix checkstyle error maxLineLength from LeftCurlyCheck

spodkowinski:
Attribute has been removed and will cause an error with recent
checkstyle releases, see https://github.com/checkstyle/checkstyle/issues/3671


michaelsembwever (approved):
LGTM

PR #486, opened 2018-07-17
Clean code: Replace RepairRun, RepairUnit and RepairSchedule's inner Builder classes

michaelsembwever:
Clean code: replace RepairRun, RepairUnit and RepairSchedule's inner Builder classes' verbose public constructor with a static `builder()` methods and builder methods.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/core/RepairRun.java
+++ b/src/server/src/main/java/io/cassandrareaper/core/RepairRun.java
@@ -17,6 +17,7 @@ package io.cassandrareaper.core;
 import java.util.Objects;
 import java.util.UUID;
 
+import com.google.common.base.Preconditions;
 import org.apache.cassandra.repair.RepairParallelism;
 import org.joda.time.DateTime;
 import org.joda.time.DateTimeComparator;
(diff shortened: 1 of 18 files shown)


Bj0rnen:
(minor rant)

Not a -1, but when you remove static guarantees and replace them with runtime checks, for the sake of better style, something isn't right. I guess that _something_ is the language in this case.

michaelsembwever:
> Not a -1, but when you remove static guarantees and replace them with runtime checks, for the sake of better style, something isn't right. I guess that something is the language in this case.

I hear you @Bj0rnen 

There's a trade-off here. The point of the _builder pattern_ is to avoid long and ambiguous constructors. What static guarantee a constructor with many parameters provides is limited, as the static guarantee depends on the type of each parameter. So the builder pattern provides more readable and arguably safer code, and then the preconditions add back the lost (and more) guarantees (albeit at runtime). That's my take on it at least.

And the main purpose of this PR was to remove/hide _all_ constructors. The use of constructors in Java isn't a great practice, especially on inner classes, as it gives less little room for api change.

And more importantly, this brings the codebase into a more consistent state.

PR #492, opened 2018-07-24
Use executorServices from DropWizard's lifecycle, in top-level singleton Managers and Resources

michaelsembwever:
This avoids executing late tasks, and the subsequent exceptions, created during reaper shutdowns (or flapping between reaper instances).

For example in the integration tests where GRIM_MIN and GRIM_MAX are set to different values to test a high-availability setup where instances are flapping a lot, this removes thousands of erroneous asynchronous executions and subsequent exceptions.


michaelsembwever:
Going to self-approve this one.

PR #494, opened 2018-07-24
Ignore late "start" notifications in SegmentRunner

michaelsembwever:
If a late `start` notification comes in (on this reaper instance, or on a different reaper instance after leadership has flipped), ignore it (it contains no information relevant for us).


ossarga (approved):
Change LGTM

PR #499, opened 2018-07-26
Add maven-enforcer-plugin to ensure known and correct java and maven versions are used to build

michaelsembwever:


ossarga (commented):
I think these changes are ok.

I was able to test the enforcement of the Java and Maven versions by running on an old Java and Maven version.
```
MAVEN_OPTS="-Xmx1g" mvn clean install
```

I was unable to fully test the `-Prelease` option because I think my `gpg` setup was incomplete. I ran this command instead which seemed to compile successfully.
```
MAVEN_OPTS="-Xmx1g" mvn clean install -Prelease -Dgpg.skip
```

ossarga (approved):

michaelsembwever:
> I was unable to fully test the -Prelease option because I think my gpg setup was incomplete. I ran this command instead which seemed to compile successfully.

That's a good thing. When running releases we need to upload gpg signed artefacts. (Usually done by travis)

PR #500, opened 2018-07-26
Integration tests for Upgrading Reaper mid-scenario

michaelsembwever:


src/server/src/test/resources/cassandra-reaper-cassandra-at.yaml
```diff
@@ -38,13 +38,6 @@ jmxPorts:
   127.0.0.7: 7700
   127.0.0.8: 7800
   
-jmxCredentials:
-  "test cluster":
-    username: cassandra
-    password: cassandra
-  test:
-    username: cassandra
-    password: cassandrapassword
```
michaelsembwever: this has to be reverted.

src/server/src/test/resources/io.cassandrareaper.acceptance/integration_reaper_functionality.feature
```diff
@@ -70,50 +80,38 @@ Feature: Using Reaper to launch repairs and schedule them
...
     Then reaper has 1 scheduled repairs for cluster called "test"
+    When reaper is upgraded to latest
+    Then reaper has 1 scheduled repairs for cluster called "test"
     And deleting cluster called "test" fails
     When all added schedules are deleted for the last added cluster
     And the last added cluster is deleted
     Then reaper has no longer the last added cluster in storage
+  ${cucumber.upgrade-versions}
 
-  Scenario: Create a cluster and a scheduled repair run with repair run history and delete them
```
michaelsembwever, line 89: This scenario outline must be restored. (Only currently removed as it fails and first priority was to generally get CI working.)

adejanovski (approved):
Awesome work Mick!

michaelsembwever:
Good to go now – just waiting on green CI.

The failing scenario is restored and working again.
It uncovered the failure that the `SchedulingManager` was a static class and would not work when multiple reaper servers were started in the one jvm, as they are in the upgrade tests.
`SchedulingManager` has been rewritten into a singleton. Some other bug fixes were applied to the tests too.

PR #501, opened 2018-07-27
Unit tests that the db migration scripts are ok

michaelsembwever:


ossarga (approved):
Ran these tests on my local build and completed successfully. Changes look good to me.

PR #508, opened 2018-08-04
Use cucumber command line tags instead of separate integration test classes

michaelsembwever:
Testing variations represented by cucumber tags are better done from the command line than hardcoded classes.


adejanovski (approved):
Much better this way! Good job!

PR #517, opened 2018-08-08
Display compaction stats in the node details UI

adejanovski:
Allows to display the list of running compactions on each node and their progress : 

compaction-stats-reaper

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/jmx/JmxProxyImpl.java
+++ b/src/server/src/main/java/io/cassandrareaper/jmx/JmxProxyImpl.java
@@ -48,6 +48,7 @@ import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.stream.Collectors;
+
 import javax.management.Attribute;
 import javax.management.AttributeList;
 import javax.management.InstanceNotFoundException;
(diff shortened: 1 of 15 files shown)


src/server/src/main/java/io/cassandrareaper/jmx/JmxProxyImpl.java
```diff
@@ -1268,4 +1280,28 @@ public void takeColumnFamilySnapshot(
     return ((StreamManagerMBean) smProxy).getCurrentStreams();
   }
 
+  public List<Compaction> listActiveCompactions()
```
michaelsembwever: i'm keen to stop any further "feature" additions to this class.

Can we rebase this PR onto `mck/feature-major-compaction` and this method into CompactionProxy.

michaelsembwever: I've done this work in https://github.com/thelastpickle/cassandra-reaper/commit/e58c6116eb4cd33b7ecf05291341305e7267a38a
if you'd like to bring it in.

src/server/src/main/java/io/cassandrareaper/service/MetricsGrabber.java
```diff
@@ -41,6 +44,7 @@
   private static final Logger LOG = LoggerFactory.getLogger(MetricsGrabber.class);
 
   private final AppContext context;
+  private final ExecutorService executor = Executors.newFixedThreadPool(5);
```
michaelsembwever: This needs to instead be a dropwizard lifecycle based executor, and also preferably wrapped with metrics instrumentation.

see work done in https://github.com/thelastpickle/cassandra-reaper/commit/b4c32482e600699c41e90ba8c9f517b81a870bf6

michaelsembwever: or if it is not used in the class it should just be removed.

michaelsembwever (approved):
LGTM

PR #518, opened 2018-08-08
[streams] Show tables using tags, so they are nicely separated

rzvoncek:
This PR fixes how the the UI shows tables involved in a stream.

Previously, the tables were concatenated without any delimiters:
screen shot 2018-08-08 at 17 27 37

After this PR, tables are going to be show using (read-only) tags:
screen shot 2018-08-08 at 17 28 11


michaelsembwever:
i think i'm late to the game on this one, but why does `StreamSessionFactory` have test methods in it?
Can this be moved to a class under `src/server/src/test/…`, of if it is used in production rename them to something less "test"-ish?

michaelsembwever (approved):

rzvoncek:
Rebased on master because out of sync. Merging without waiting another round of tests (they were all green before).

PR #520, opened 2018-08-11
Remove stateless services from context

michaelsembwever:
A number of stateless services have crept into the application context `AppContext` class.

We want to avoid cyclic dependency between `AppContext` and service classes. It should be only a `AppContext` <- service-class relationship.
The only current exception to that it our state service `RepairManager` that is contextual to a number other services.

This PR can't be merged (into master) until #517 is first merged.


adejanovski (approved):
LGTM

adejanovski:
Closing as the commits were merged and squashed into #517

PR #533, opened 2018-08-24
Update Cassandra-reaper to fix #184

### LINKED ISSUE #184: put reaper in homebrew
`brew install cassandra-reaper` needs to be a thing, right?

We'll need to pick some reasonable default settings.  I think this is a valid use case for H2 since it doesn't have any other dependencies.  

https://github.com/Homebrew/brew/blob/master/docs/How-to-Create-and-Maintain-a-Tap.md

kotaroyamazaki:
To fix #184, I made a formula for cassandra-reaper.
It needs an update to the `cassandra-reaper` script to be able to read .jar and .yaml from /usr/local/ .

https://github.com/KotaroYamazaki/homebrew-Cassandra-Reaper

### DIFF
--- a/src/packaging/bin/cassandra-reaper
+++ b/src/packaging/bin/cassandra-reaper
@@ -22,11 +22,15 @@ fi
 
 if [ -z "$CLASS_PATH" ]; then
   echo "Looking for reaper in /usr/share/cassandra-reaper/"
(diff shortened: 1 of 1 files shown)


adejanovski (approved):
LGTM

PR #536, opened 2018-08-25
Downgrade the enforced Maven version to 3.3.8 to allow the Docker build to work again

adejanovski:
The Docker build uses an image that sports the 3.3.8 version of Maven.
Since there doesn't seem to be a pressing issue that requires Maven 3.5.2, I think it's the easiest way to handle this issue.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -315,8 +315,8 @@
         <version.jdk>1.8</version.jdk>
         <!-- travis of July 2018 was still using _151 -->
         <build.jdk.minimum>1.8.0-151</build.jdk.minimum>
-        <!-- travis of July 2018 was still using maven-3.5.2 -->
-        <maven.version.minimum>3.5.2</maven.version.minimum>
+        <!-- travis of July 2018 was still using maven-3.5.2 but our Docker image for building packages ships with 3.3.8 -->
+        <maven.version.minimum>3.3.8</maven.version.minimum>
         <surefireArgLine></surefireArgLine>
 
         <!--


michaelsembwever (changes requested):
This doesn't fix the docker build.
(it was broken long before this change)

it's also pretty easy i believe to upgrade the docker image used so to use a newer maven version.

michaelsembwever:
Oh, I see…

Any idea why the docker CI task is failing?
(and why we don't have the docker build in question also in CI?)

> If not, what's making us enforce Maven 3.5.2 over 3.3.8 ?

Generally to reduce the surface area for behavioural incompatibilities.
But it's not worth wasting time on, just having a floor is the value.

pom.xml
```diff
@@ -297,7 +297,7 @@
         <!-- travis of July 2018 was still using _151 -->
         <build.jdk.minimum>1.8.0-151</build.jdk.minimum>
         <!-- travis of July 2018 was still using maven-3.5.2 -->
```
michaelsembwever: the comment needs to be updated.

michaelsembwever (changes requested):

pom.xml
adejanovski: I removed the comments as they seemed redundant with the actual variables and the version Travis uses should not enforce what we allow. Right ?

michaelsembwever: I think we want the comments there. We need a way to know who we're waiting on to raise the version.

For example these comments would tell us that it is `3.3.8` because of the docker build. That makes it easy to go check that first when we want to raise it. (It saves a lot of time over having to run CI tests in branches, even in master,  to see if there's a possibility to raise the version.

michaelsembwever (approved):
thanks for adding the comment

PR #553, opened 2018-09-16
Allow to configure a custom JMX port per cluster

### LINKED ISSUE #550: Reaper can not work with cluster with custom jmx port
In webui I can sucessfully add new cluster with custom jmx port using seed node in ip:custom_port format.
But webui show cluster hosts from another cluster witch worked on standard jmx port, not my custom port.
I tryed to set seed_hosts in reaper_db.cluster in ip:custom_port format. It temporarily fix problem, webui show correct hosts in cluster. But seed_hosts was updated by reaper with values without custom ports soon.

adejanovski:
Fixes #550 

Adding a cluster to Reaper now takes a jmx port parameter which will be stored in the backend for that cluster.
All JMX connection now require a full Cluster object to build the Node instance that is passed to the JmxConnectionFactory so we can get the custom port for that cluster.
The integration tests have been modified so that node 127.0.0.1 jmx port is now removed from the yaml file and the cluster is registered with port 7100.
The other nodes JMX ports are still in the yaml file and will prevail over the cluster's custom JMX port.


src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
```diff
@@ -327,16 +341,26 @@ public Response deleteCluster(
...
+            jmxPort.orElse(Cluster.DEFAULT_JMX_PORT));
     return () -> {
       try {
-        JmxProxy jmxProxy = connectAny(seeds, clusterName);
+        JmxProxy jmxProxy = connectAny(seeds, cluster);
         FailureDetectorProxy proxy = FailureDetectorProxy.create(jmxProxy);
-        return Optional.of(new NodesStatus(jmxProxy.getHost(), proxy.getAllEndpointsState(), proxy.getSimpleStates()));
+        return Optional.of(
+            new NodesStatus(
+                jmxProxy.getHost(), proxy.getAllEndpointsState(), proxy.getSimpleStates()));
```
michaelsembwever, line 376: there's no need to wrap this line (and there was no change)

src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
```diff
@@ -408,11 +434,29 @@ static String parseSeedHost(String seedHost) {
     return Optional.empty();
   }
 
-  private JmxProxy connectAny(Collection<String> seedHosts, String cluster) throws ReaperException {
+  private JmxProxy connectAny(Collection<String> seedHosts, String clusterName)
+      throws ReaperException {
```
michaelsembwever, line 453: no need to wrap this line.

src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
```diff
@@ -408,11 +434,29 @@ static String parseSeedHost(String seedHost) {
...
+                    Node.builder()
+                        .withCluster(context.storage.getCluster(clusterName).get())
+                        .withHostname(parseSeedHost(host))
+                        .build())
+            .collect(Collectors.toList()),
+        context.config.getJmxConnectionTimeoutInSeconds());
+  }
+
+  private JmxProxy connectAny(Collection<String> seedHosts, Cluster cluster)
+      throws ReaperException {
```
michaelsembwever, line 466: one line method declaration please.

src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
```diff
@@ -408,11 +434,29 @@ static String parseSeedHost(String seedHost) {
...
     return context.jmxConnectionFactory.connectAny(
         seedHosts
             .stream()
-            .map(host -> Node.builder().withClusterName(cluster).withHostname(parseSeedHost(host)).build())
+            .map(
+                host ->
+                    Node.builder().withCluster(cluster).withHostname(parseSeedHost(host)).build())
```
michaelsembwever, line 472: no need to wrap this line (and no change)

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
```diff
@@ -220,13 +220,16 @@ private boolean runRepair() {
     Thread.currentThread().setName(clusterName + ":" + segment.getRunId() + ":" + segmentId);
 
     try (Timer.Context cxt = context.metricRegistry.timer(metricNameForRunRepair(segment)).time()) {
-      JmxProxy coordinator
-          = context.jmxConnectionFactory.connectAny(
+      JmxProxy coordinator =
+          context.jmxConnectionFactory.connectAny(
```
michaelsembwever: can we keep the `=` on the newline.

in code style this clearly separates lines wrapped per parameter/argument versus operators.
this applies to the rest of the PR as well.

(the C* codebase also asks that operators carry when wrapping, it makes the code faster to read by providing this distinction)

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
```diff
@@ -237,9 +237,11 @@ private static void initializeAndUpgradeSchema(
...
   private void prepareStatements() {
     final String timeUdf = 0 < VersionNumber.parse("2.2").compareTo(version) ? "dateOf" : "toTimestamp";
-    insertClusterPrepStmt = session
-        .prepare("INSERT INTO cluster(name, partitioner, seed_hosts) values(?, ?, ?)")
-        .setConsistencyLevel(ConsistencyLevel.QUORUM);
+    insertClusterPrepStmt =
+        session
+            .prepare(
+                "INSERT INTO cluster(name, partitioner, seed_hosts, jmx_port) values(?, ?, ?, ?)")
+            .setConsistencyLevel(ConsistencyLevel.QUORUM);
```
michaelsembwever, line 250: the line wrapping doesn't need to be changed here.

src/server/src/test/java/io/cassandrareaper/acceptance/BasicSteps.java
```diff
@@ -251,11 +252,13 @@ public void an_add_cluster_request_is_made_to_reaper() throws Throwable {
                 Assertions.assertThat(responseEntity).isEmpty();
 
                 // follow to new location (to GET resource)
-                response = runner.callReaper("GET", response.getLocation().toString(), Optional.absent());
+                response =
+                    runner.callReaper("GET", response.getLocation().toString(), Optional.absent());
```
michaelsembwever: no need to wrap.

src/server/src/test/java/io/cassandrareaper/acceptance/BasicSteps.java
```diff
@@ -251,11 +252,13 @@ public void an_add_cluster_request_is_made_to_reaper() throws Throwable {
...
 
                 String responseData = response.readEntity(String.class);
                 Assertions.assertThat(responseData).isNotEmpty();
-                Map<String, Object> cluster = SimpleReaperClient.parseClusterStatusJSON(responseData);
+                Map<String, Object> cluster =
+                    SimpleReaperClient.parseClusterStatusJSON(responseData);
```
michaelsembwever: no need to wrap.

michaelsembwever (changes requested):
Threw out a few comments. Line wrapping is causing a headache, would be great to get that in order (consistent for the project).

Otherwise my questions:
 - how can we better deal with the Cluster partitioner parameter? sometimes the argument is null, sometimes a blank string. Something seems lacking here. 
 - our horizon here is to remove all cluster-based configuration from the yaml configuration (as it's infact application-state and not reaper configuration). while it needs not be part of this PR it would be good to know how what we're thinking here: how are we going to remove `jmxPorts` from the yaml? how are we going to support different jmx ports per node? as we move more cluster and node state from yaml to storage will we just keep adding db columns?

michaelsembwever:
I created #556 to remove the need for nit-picking over carrying the operator when wrapping lines in the future.

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
michaelsembwever: I created #556 to remove the need for nit-picking over carrying the operator when wrapping lines in the future.

src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
```diff
@@ -237,15 +241,21 @@ private Response addOrUpdateCluster(
...
 
     Set<String> seedHosts = parseSeedHosts(seedHost);
-    String cluster = parseClusterNameFromSeedHost(seedHost).orElse("");
+    String parsedClusterName = parseClusterNameFromSeedHost(seedHost).orElse("");
 
     try {
+      Cluster cluster =
+          new Cluster(
+              parsedClusterName,
+              null,
```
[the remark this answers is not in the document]
michaelsembwever: is related to https://github.com/thelastpickle/cassandra-reaper/pull/553#discussion_r217930325

was agreed offline with @adejanovski that the partitioner parameter should be make optional (or a constructor added without it) as it's not always supplied, and passing in either `null` or `""` is bad.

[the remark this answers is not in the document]
michaelsembwever: (the builder pattern could also be introduced)

michaelsembwever:
Answering these questions from offline discussion with @adejanovski 

> questions:
>
> - how can we better deal with the Cluster partitioner parameter? sometimes the argument is null, sometimes a blank string. Something seems lacking here.

See https://github.com/thelastpickle/cassandra-reaper/pull/553#discussion_r218601666

> - our horizon here is to remove all cluster-based configuration from the yaml configuration (as it's infact application-state and not reaper configuration). while it needs not be part of this PR it would be good to know how what  we're thinking here: how are we going to remove jmxPorts from the yaml? how are we going to support different jmx ports per node? as we move more cluster and node state from yaml to storage will we just keep adding db  columns?

Agreed that instead of adding a `jmx_port` column to the various backends schema it would be better to add a `cluster_configuration` text column. this column would store a json map of cluster-based configuration settings, for example the `jmx_port`.

Such a `cluster_configuration` column could also store data-center and node specific configurations for the cluster. Examples are different repair intensity defaults for different datacenters, and different jmx_ports for different nodes.

Hurdles identified with moving application-state away from the yaml file are:
 - passwords can't be stored in plaintext in the column. reaper might create it's own encryption key to store passwords encrypted…
 - node specific configurations, like different jmx_port on different nodes (eg ccm), will be difficult to implement in the UI

michaelsembwever:
@adejanovski , the only requested changes standing from my review is fixing the Cluster.partitioner parameter.

michaelsembwever (approved):

PR #558, opened 2018-09-19
Remove unused (and test) methods from StreamSessionFactory

michaelsembwever:


rzvoncek:
This is a code I've been using when I wanted to mock some responses to test the UI. While I'm sad to see it go, I understand it has no place in the tree.

In the future, would it be acceptable to replace this with some sort of a mock endpoint that would read this content from a resource file?

rzvoncek (approved):

michaelsembwever:
> In the future, would it be acceptable to replace this with some sort of a mock endpoint that would read this content from a resource file?

Could it have been under `src/server/src/test/java/` ?

PR #566, opened 2018-09-28
Remove the release profile and fix the version number evaluation with mvn

adejanovski:
Using the -B flag on maven didn't allow anymore to extract the version number, which prevented from pushing the artifacts to bintray.
This commit also removes rultor and the release profile, along with the Apache parent pom as they were not used/necessary at this stage.


michaelsembwever (approved):
LGTM. 
Some of this won't get tested until actually going through a release.

PR #582, opened 2018-10-29
Fix ui left node status and token count

adejanovski:
This PR brings two things in separate commits : 
- Remove nodes that left the cluster from the cluster view in the UI
- Fix the token count in the node detail screen and allow to display the list of tokens

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/jmx/JmxProxy.java
+++ b/src/server/src/main/java/io/cassandrareaper/jmx/JmxProxy.java
@@ -127,5 +127,4 @@ public interface JmxProxy extends NotificationListener {
   void close();
 
   void removeRepairStatusHandler(int repairNo);
-
 }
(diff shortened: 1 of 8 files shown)


src/server/src/main/java/io/cassandrareaper/jmx/JmxProxy.java
```diff
@@ -128,4 +128,6 @@ int triggerRepair(
 
   void removeRepairStatusHandler(int repairNo);
 
+  Map<String, List<String>> getTokensByNode();
```
michaelsembwever: Can we move this to `EndpointSnitchInfoProxy` or to a new domain-specific Proxy class please.

michaelsembwever (changes requested):

src/server/src/main/java/io/cassandrareaper/jmx/JmxProxy.java
adejanovski: @michaelsembwever, I made the requested changes and created the `StorageServiceProxy` class.

michaelsembwever (approved):

src/server/src/main/java/io/cassandrareaper/jmx/JmxProxy.java
michaelsembwever: I had avoided `StorageServiceProxy` as it's a "junk yard" where everything gets shoved into. I was kinda hoping for something more "domain" expressive. But it's not a blocker.

PR #584, opened 2018-11-02
Rename Reaper metrics to make them compatible with Prometheus

rzvoncek:


src/server/src/main/java/io/cassandrareaper/jmx/HostConnectionCounters.java
```diff
@@ -45,9 +45,9 @@ void incrementSuccessfulConnections(String host) {
       LOG.debug("Host {} has {} successfull connections", host, successes);
       if (null != metricRegistry) {
         metricRegistry
-            .counter(
-                MetricRegistry.name(
-                    JmxConnectionFactory.class, "connections", host.replace('.', '-')))
+            .counter(MetricRegistry.name(JmxConnectionFactory.class,
+                                         "connections",
+                                         host.replace('.', 'x').replaceAll("[^A-Za-z0-9]", "")))
```
michaelsembwever: Can you undo the line wrapping changes. I believe the additional .replaceAll call doesn't break the line-length code style rule.
```suggestion
            .counter(
                MetricRegistry.name(
                    JmxConnectionFactory.class, "connections",host.replace('.', 'x').replaceAll("[^A-Za-z0-9]", "")))                                        
```

src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
```diff
@@ -587,7 +587,8 @@ void killAndCleanupRunner() {
   private String metricName(String metric, String clusterName, String keyspaceName, UUID repairRunId) {
     String cleanClusterName = clusterName.replaceAll("[^A-Za-z0-9]", "");
     String cleanRepairRunId = repairRunId.toString().replaceAll("-", "");
-    return MetricRegistry.name(RepairRunner.class, metric, cleanClusterName, keyspaceName, cleanRepairRunId);
+    String cleanKeyspaceName = keyspaceName.replaceAll("[^A-Za-z0-9]", "");
```
michaelsembwever: hasn't `cleanKeyspaceName` already replaced those characters?

michaelsembwever:
Look good. But you'll need to fix the indentation.

And the tests are failing as SegmentRunnerTest doesn't set the keyspaceName in the RepairUnit it mocks. (SegmentRunnerTest should be fixed, as production code would never have a null keyspaceName.)

rzvoncek:
Reverted the indentation and fixed the SegmentRunnerTest. Had to remove one blank line because checkstyle was complaining about the method being too long (151 vs 150 allowed lines).

I've also rebased on current master.

michaelsembwever (approved):
:shipit:

rzvoncek:
Squashed.

michaelsembwever:
Manually merged to `1.3` and `master`.

In `1.3` it is a compatibility break, but in a rather controlled and cosmetic way I think we can clearly communicate and explain.

PR #585, opened 2018-11-04
Fixes #577.  Allow for automatically blacklisting any TWCS or DTCS ta…

### LINKED ISSUE #577: Disabling repair of tables with TWCS
We have a bunch of tables containing time series data with TTL’ed data using TWCS. Initially our plan was to simply enable the auto scheduling feature for all our keyspaces, but quickly noticed a huge difference in repair durations. After digging into why, we noticed that it was running a lot of compactions on TWCS tables containing our TTL’ed time series data. After digging a bit more, I found a post proposing to disable repair of such tables entirely and only run repairs when machines have been down longer than the configured hint window.

http://thelastpickle.com/blog/2016/12/08/TWCS-part1.html#comment-3558631910

Ideally I’d prefer to stick to auto scheduling as much as possible, but I’d rather not exclude entire keyspaces as not all tables in a given keyspace are using TWCS with TTL’ed data.

What is the recommended approach to handle this case? I was hoping was a flag that I could use to get this behaviour using the image from https://hub.docker.com/r/thelastpickle/cassandra-reaper/

denniskline:
…bles within a keyspace

Addressing the need to the automatically blacklist TWCS/DTCS tables per issue #577.

I decided to create a new Table object to handle both the Table.name and the Table.compactionStrategy in order to pull the required information in fewer calls.  Doing this lead to additional modifications to consumers of JmxProxy 

Please let me know if the approach looks good to you.


src/server/src/main/java/io/cassandrareaper/ReaperApplicationConfiguration.java
```diff
@@ -65,6 +65,11 @@
   @DefaultValue("false")
   private Boolean incrementalRepair;
 
+  @JsonProperty
+  @NotNull
+  @DefaultValue("true")
+  private Boolean blacklistTwcsTables;
```
michaelsembwever, line 71: We're moving towards (thinking about?) keeping application-state in the storage, and keeping the configuration purely about Reaper's configuration (and not about the configuration of clusters that a user registers in a running Reaper).
That change hasn't materialised yet, but am noting it here for forward-reference sake.

src/server/src/main/java/io/cassandrareaper/ReaperApplicationConfiguration.java
```diff
@@ -185,6 +190,14 @@ public void setIncrementalRepair(boolean incrementalRepair) {
     this.incrementalRepair = incrementalRepair;
   }
 
+  public boolean getBlacklistTwcsTables() {
+    return blacklistTwcsTables != null ? blacklistTwcsTables : false;
```
michaelsembwever, line 194: Isn't returning `false` here contradictory to the default being `false`?

michaelsembwever (approved):

michaelsembwever:
How will this work with Cassandra-4.0 and incremental repair?
That is with Cassandra-4.0 we want to recommend incremental repair, and to run it continuously. Then running it on TWCS makes sense (and the default to blacklist it should not be `true`).

src/server/src/main/java/io/cassandrareaper/ReaperApplicationConfiguration.java
adejanovski: The default is set to `true`, but AFAIK that annotation is ignored by Dropwizard. I suggest to remove it and indeed keep the actual default to `false`.

src/server/src/main/java/io/cassandrareaper/ReaperApplicationConfiguration.java
denniskline: That’s great to hear.  That will definitely make the app config cleaner. Plus, supporring multiple clusters is tricky for us since we need to update reaper config and cycle when new clusters are added (all contain their own unique jmx creds).

denniskline:
@michaelsembwever - Did you want me to work on allowing TWC repairs on 4.0 before this merged in?  I'm a bit confused if you want to see more added to this PR or if you like it the way it is.  I'm happy to make changes.  Thanks!

michaelsembwever:
> Did you want me to work on allowing TWC repairs on 4.0 before this merged in? 

It's not a blocker. I was mostly making the comment FTR. 

It's not so much specific to Cassandra 4.0 but to incremental repairs (yes, people are unfortunately still using incremental repairs pre Cassandra 4.0).
It's also on the caveat that TWCS is ok on incremental repairs so long as the incremental repairs are run with a frequency equal to or greater (more often) than the TWCS time_window.
If you would like to add the code to blacklist only when non-incremental, then that would be awesome. 

Instead of merging master into branches could you rebase the branch off master.
Merges going in both directions leads to lots of problems in git repositories, and on the tofu scale we merge into more stable branches and rebase a softer branch.

Otherwise we can merge this after 
 - the fix for https://github.com/thelastpickle/cassandra-reaper/pull/585#discussion_r231028560
 - rebase the commit off master (removing the merge commits from the PR)

michaelsembwever:
I took over your awesome work here @denniskline, and continued it in https://github.com/thelastpickle/cassandra-reaper/pull/602

PR #601, opened 2019-01-15
WIP - Sidecar mode

adejanovski:


src/server/src/main/java/io/cassandrareaper/jmx/JmxConnectionFactory.java
```diff
@@ -80,9 +82,11 @@ private void registerConnectionsGauge() {
     }
   }
 
-  protected JmxProxy connectImpl(Node node, int connectionTimeout) throws ReaperException, InterruptedException {
+  protected JmxProxy connectImpl(Node node, AppContext context) throws ReaperException, InterruptedException {
```
michaelsembwever: `AppContext` can be injected in via the constructor, rather than every method.

src/server/src/main/java/io/cassandrareaper/jmx/JmxConnectionFactory.java
```diff
@@ -80,9 +82,11 @@ private void registerConnectionsGauge() {
...
     // use configured jmx port for host if provided
-    String host = node.getHostname();
+    String host = context.config.isInSidecarMode()
+        ? LOCALHOST
+        : node.getHostname();
```
michaelsembwever, line 87: there was discussion to change this to `Preconditions.checkArgument(!context.config.isInSidecarMode() || LOCALHOST.equals(node.getHostname()));`
 so we smoke out all the places that are trying to make jmx connections to other nodes (there's a few places in the code!)

src/server/src/main/java/io/cassandrareaper/jmx/JmxConnectionFactory.java
adejanovski: Done.

We have a circular reference though as the jmxConnectionFactory now takes the context as argument and is stored in the same context.
Not sure that's really a problem though.

adejanovski:
replaced by #622

PR #602, opened 2019-01-16
Allow for automatically blacklisting any TWCS or DTCS tables within a keyspace

### LINKED ISSUE #577: Disabling repair of tables with TWCS

michaelsembwever:
Fixes #577.

 ref: https://github.com/thelastpickle/cassandra-reaper/pull/585

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/ReaperApplicationConfiguration.java
+++ b/src/server/src/main/java/io/cassandrareaper/ReaperApplicationConfiguration.java
@@ -1,6 +1,6 @@
 /*
  * Copyright 2014-2017 Spotify AB
- * Copyright 2016-2018 The Last Pickle Ltd
+ * Copyright 2016-2019 The Last Pickle Ltd
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -64,6 +64,11 @@ public final class ReaperApplicationConfiguration extends Configuration {
   @NotNull
   @DefaultValue("false")
   private Boolean incrementalRepair;
+
+  @JsonProperty
+  @NotNull
+  @DefaultValue("false")
+  private Boolean blacklistTwcsTables;
 
   @DefaultValue("7")
   private Integer scheduleDaysBetween;
@@ -183,6 +188,14 @@ public final class ReaperApplicationConfiguration extends Configuration {
 
   public void setIncrementalRepair(boolean incrementalRepair) {
     this.incrementalRepair = incrementalRepair;
+  }
+
+  public boolean getBlacklistTwcsTables() {
+    return blacklistTwcsTables != null ? blacklistTwcsTables : false;
+  }
+
+  public void setBlacklistTwcsTables(boolean blacklistTwcsTables) {
+    this.blacklistTwcsTables = blacklistTwcsTables;
   }
 
   public Integer getScheduleDaysBetween() {
--- a/src/server/src/main/java/io/cassandrareaper/service/ClusterRepairScheduler.java
+++ b/src/server/src/main/java/io/cassandrareaper/service/ClusterRepairScheduler.java
@@ -1,6 +1,6 @@
 /*
  * Copyright 2017-2017 Spotify AB
- * Copyright 2017-2018 The Last Pickle Ltd
+ * Copyright 2017-2019 The Last Pickle Ltd
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -30,7 +30,6 @@ import java.util.Set;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 
-import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Sets;
 import org.joda.time.DateTime;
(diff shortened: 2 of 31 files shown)


src/server/src/main/java/io/cassandrareaper/ReaperApplicationConfiguration.java
```diff
@@ -65,6 +65,11 @@
   @DefaultValue("false")
   private Boolean incrementalRepair;
 
+  @JsonProperty
+  @NotNull
+  @DefaultValue("false")
+  private Boolean blacklistTwcsTables;
```
adejanovski, line 71: The name of that variable suggests that we're blacklisting TWCS tables from all repair runs, which isn't exactly what it does. Currently we only blacklist for scheduled runs, not manual ones.
We could either:
- rename it to `blacklistTwcsTablesForScheduledRuns`
or
- blacklist TWCS/DTCS tables in manual runs as well.

src/server/src/main/java/io/cassandrareaper/service/ClusterRepairScheduler.java
```diff
@@ -109,10 +114,15 @@ private boolean keyspaceCandidateForRepair(Cluster cluster, String keyspace) {
...
 
+    Set<String> blacklistedTables = context.config.getBlacklistTwcsTables()
+        ? findBlacklistedCompactionStrategyTables(context, cluster, keyspace)
+        : Collections.emptySet();
+
     RepairUnit.Builder builder = RepairUnit.builder()
         .clusterName(cluster.getName())
         .keyspaceName(keyspace)
         .incrementalRepair(incrementalRepair)
+        .blacklistedTables(blacklistedTables)
```
adejanovski: We may want to blacklist them only if they're not in the explicit list of tables to repair. One could want to selectively repair TWCS/DTCS tables, which could be allowed if said table is passed in the list of tables to repair.

Also, it should not be possible to create a repair that contains both a blacklist and a list of table.
Thus, there's a little more work to do here : 
- if no table has been explicitly passed (repair the whole keyspace), it's ok to set a blacklist
- if tables were passed, remove (or not?) blacklisted tables from the table list instead of passing a blacklist.

We need to agree on wether or not we allow TWCS tables to be repaired if they're passed in the explicit table list (I think we should allow it).

src/server/src/main/java/io/cassandrareaper/ReaperApplicationConfiguration.java
michaelsembwever: Blacklisting now occurs when repairs are triggered, regardless of whether they were automatic, scheduled, or manual.

src/server/src/main/java/io/cassandrareaper/service/ClusterRepairScheduler.java
michaelsembwever: Tables that are explicitly listed are not blacklisted.

michaelsembwever:
See https://github.com/thelastpickle/cassandra-reaper/pull/602/files#diff-961ef0dbf187f606a6bf6d9419950f99R144

> Disables repairs of any tables that use either the `TimeWindowCompactionStrategy` or `DateTieredCompactionStrategy`. This automatic blacklisting is not stored in schedules or repairs. It is applied when repairs are triggered and visible in the UI for running repairs. Not storing which tables are TWCS/DTCS ensures changes to a table's compaction strategy are honored on every new repair.

I changed up the PR to take this approach instead.

This means that TWCS/DTCS tables don't appear in the blacklistings under schedules, nor are stored in the backend in the `RepairUnit`s. 
They will appear in the UI for repair runs.

src/server/src/main/java/io/cassandrareaper/service/RepairUnitService.java
```diff
@@ -67,6 +74,64 @@ public RepairUnit getOrCreateRepairUnit(Cluster cluster, RepairUnit.Builder para
...
+          .collect(Collectors.toSet());
+    } else {
+      // if tables have been specified then don't apply the twcsBlacklisting
+      tables = repairUnit.getColumnFamilies().stream()
+            .filter(tableName -> !repairUnit.getBlacklistedTables().contains(tableName))
+            .collect(Collectors.toSet());
+    }
+
+    Preconditions.checkState(
+        !(tables.isEmpty() && !repairUnit.getBlacklistedTables().isEmpty()),
```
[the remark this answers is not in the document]
michaelsembwever: That logic would be easier to read as
`repairUnit.getBlacklistedTables().isEmpty() || !tables.isEmpty() `

but i agree it doesn't really apply. What it was trying to catch is the situation where you've manually or automatically blacklisted all tables in the keyspace.
But this isn't so much a precondition, as it is a controller-layer error that needs to be replied to the user.

i will think about it more…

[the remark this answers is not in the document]
michaelsembwever: > Are you now always creating an explicit list of tables and storing it in the repair unit/run?

The answer to that is no.

> Your changes suggest that you're allowing both to be set and throw an error if you have a blacklist without a table list.

This wasn't what this check was about.
I'll add that check into `RepairRunResource.checkRequestForAddRepair(..)`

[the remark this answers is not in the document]
michaelsembwever: > I'll add that check into RepairRunResource.checkRequestForAddRepair(..)

 Done.

adejanovski (approved):
LGTM.
:shipit:

PR #604, opened 2019-01-21
Full Shiro protection

michaelsembwever:
- Authentication is now enabled by default. Default is the dummy admin/admin credentials, see shiro.ini
 - shiro.ini now embedded in app. can still be configured/overridden
 - REST endpoints are protected, requiring JWT specified in a request header ('Authorization: Bearer <jwt>')
 - The JWT for a user can be gained at the /jwt URL
 - add 'remember me' to login page

TODO
 - docs
 - spreaper support


michaelsembwever:
spreaper support added in #607

PR #607, opened 2019-01-22
Add support for authentication in spreaper

adejanovski:
spreaper will now accept --username and --password for all operations.
A login will be performed before running the desired command and the login JSESSIONID cookie will be added to all HTTP requests.

### DIFF
--- a/src/packaging/bin/spreaper
+++ b/src/packaging/bin/spreaper
@@ -46,7 +46,7 @@ if "-vv" in sys.argv:
     log_level = logging.DEBUG
 if "-q" in sys.argv or "--quiet" in sys.argv:
     if logging.WARN != log_level:
-        print "--verbose and --quiet options can't both be specified"
+        print("--verbose and --quiet options can't both be specified")
         exit(1)
     quiet = True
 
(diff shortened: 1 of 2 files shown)


src/packaging/bin/spreaper
```diff
@@ -397,12 +411,50 @@ class ReaperCLI(object):
...
+                "Authenticate to Reaper"
+            )
+            password = ReaperCLI.get_password()
+            printq("# Logging in...")
+            payload = {'username':args.username, 'password':password, 'rememberMe':False}
+            reply = reaper.postFormData("login", 
+                                payload=payload)
+            # Use shiro's session id to request a JWT
+            COOKIES["JSESSIONID"] = reply.cookies["JSESSIONID"]
+            jwt = reaper.get("jwt")
```
michaelsembwever: We want a way to read the `REAPER_JWT_TOKEN` from either a environment variable or the `~/.reaper` file, so to avoid the login request (which sends the password over the wire every spreaper call).

michaelsembwever (changes requested):
See comment. We need a way to avoid the password being sent over HTTP, ie a way to store/read the jwt token.

src/packaging/bin/spreaper
adejanovski: Setting an env variable that can be reused between calls doesn't seem to be possible.
I've just pushed a change to store the JWT in `~/.reaper_jwt` on top of accepting the `--jwt` arg.
The token isn't displayed anymore and the user is just told he is now authenticated to Reaper. Any 403 error would display a message asking to login again.

michaelsembwever: Thanks.

Wasn't thinking about setting the env variable here, just being able to read one already set by the user.

Will test it over the weekend. Am I free to squash, rebase and merge?

adejanovski: Sure thing

michaelsembwever (approved):

PR #618, opened 2019-02-16
 In NodesStatus when parsing gossipinfo status, also exclude those nodes that have been "removed", as well as those that are "left"

michaelsembwever:


adejanovski (approved):
LGTM.

Maybe we should list all statuses we can get and handle them all in this PR?

michaelsembwever:
> Maybe we should list all statuses we can get and handle them all in this PR?

agree.

michaelsembwever:
The possible states are listed here: https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/gms/VersionedValue.java#L67-L79

I don't think that `leaving` or `removing` should be added to the list, as they're operations in-progress, and may not complete.

`hibernate` is more complicated. A node that's started up again with `cassandra.join_ring=false` (usually so that it can repair before coming online and serving requests to clients) I think should still be listed (and red), as it's reducing the cluster's availability. Co-ordinator nodes that are started with this options don't own tokens, so I don't believe will be in `hibernate` state (although i don't know what state they will be in!)

So I'm going ahead and merging this.

PR #619, opened 2019-02-19
Encapsulate jmx calls

adejanovski:
Prep work for the sidecar mode.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/ReaperApplication.java
+++ b/src/server/src/main/java/io/cassandrareaper/ReaperApplication.java
@@ -178,7 +178,7 @@ public final class ReaperApplication extends Application<ReaperApplicationConfig
 
     if (context.jmxConnectionFactory == null) {
       LOG.info("no JMX connection factory given in context, creating default");
-      context.jmxConnectionFactory = new JmxConnectionFactory(context.metricRegistry);
+      context.jmxConnectionFactory = new JmxConnectionFactory(context);
 
       // read jmx host/port mapping from config and provide to jmx con.factory
       Map<String, Integer> jmxPorts = config.getJmxPorts();
@@ -266,7 +266,7 @@ public final class ReaperApplication extends Application<ReaperApplicationConfig
         "Cassandra backend storage is the only one allowing EACH datacenter availability modes.");
 
     ScheduledExecutorService scheduler = new InstrumentedScheduledExecutorService(
-            environment.lifecycle().scheduledExecutorService("ReaperApplication-scheduler").threads(1).build(),
+            environment.lifecycle().scheduledExecutorService("ReaperApplication-scheduler").threads(3).build(),
             context.metricRegistry);
 
     if (context.storage instanceof IDistributedStorage) {
@@ -274,6 +274,7 @@ public final class ReaperApplication extends Application<ReaperApplicationConfig
       // us to poll the database for running repairs regularly
       // only with Cassandra storage
       scheduleRepairManager(scheduler);
+      scheduleHandleMetricsRequest(scheduler);
     } else {
       // Storage is different than Cassandra, assuming we have a single instance
       context.repairManager.resumeRunningRepairRuns();
@@ -285,6 +286,7 @@ public final class ReaperApplication extends Application<ReaperApplicationConfig
     LOG.warn("Reaper is ready to get things done!");
   }
 
+
   private void scheduleRepairManager(ScheduledExecutorService scheduler) {
     scheduler.scheduleWithFixedDelay(
         () -> {
@@ -294,6 +296,23 @@ public final class ReaperApplication extends Application<ReaperApplicationConfig
             // test-pollution: grim_reaper trashes this log error
             //if (!Boolean.getBoolean("grim.reaper.running")) {
             LOG.error("Couldn't resume running repair runs", e);
+            //}
+          }
+        },
+        0,
+        10,
+        TimeUnit.SECONDS);
+  }
+
+  private void scheduleHandleMetricsRequest(ScheduledExecutorService scheduler) {
+    scheduler.scheduleWithFixedDelay(
+        () -> {
+          try {
+            context.repairManager.handleMetricsRequests();
+          } catch (ReaperException | RuntimeException e) {
+            // test-pollution: grim_reaper trashes this log error
+            //if (!Boolean.getBoolean("grim.reaper.running")) {
+            LOG.error("Couldn't handle metrics requests", e);
             //}
           }
         },
--- a/src/server/src/main/java/io/cassandrareaper/core/GenericMetric.java
+++ b/src/server/src/main/java/io/cassandrareaper/core/GenericMetric.java
@@ -1,0 +1,152 @@
+/*
+ *
+ * Copyright 2019-2019 The Last Pickle Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.cassandrareaper.core;
+
+import org.joda.time.DateTime;
+
+public final class GenericMetric {
+
+  private final String clusterName;
+  private final String metricDomain;
+  private final String metricType;
+  private final String metricScope;
+  private final String metricName;
+  private final String metricAttribute;
+  private final String host;
+  private final DateTime ts;
+  private final double value;
+
+  private GenericMetric(Builder builder) {
+    this.clusterName = builder.clusterName;
+    this.metricDomain = builder.metricDomain;
+    this.metricType = builder.metricType;
+    this.metricScope = builder.metricScope;
+    this.metricName = builder.metricName;
+    this.metricAttribute = builder.metricAttribute;
+    this.host = builder.host;
+    this.ts = builder.ts;
+    this.value = builder.value;
+  }
+
+  public String getClusterName() {
+    return clusterName;
+  }
+
+  public String getMetricDomain() {
+    return metricDomain;
+  }
+
+  public String getMetricType() {
+    return metricType;
+  }
+
+  public String getMetricScope() {
+    return metricScope;
+  }
+
+  public String getMetricName() {
+    return metricName;
+  }
+
+  public String getMetricAttribute() {
+    return metricAttribute;
+  }
+
+  public String getMetricFullId() {
+    return metricDomain + ":type=" + metricType + ",scope=" + metricScope + ",name=" + metricName;
+  }
+
+  public String getHost() {
+    return host;
+  }
+
+  public DateTime getTs() {
+    return ts;
+  }
+
+  public double getValue() {
+    return value;
+  }
+
+  public static Builder builder() {
+    return new Builder();
+  }
+
+  public static final class Builder {
+    private String clusterName;
+    private String metricDomain;
+    private String metricType;
+    private String metricScope;
+    private String metricName;
+    private String metricAttribute;
+    private String host;
+    private DateTime ts;
+    private double value;
+
+    private Builder() {}
+
+    public Builder withClusterName(String clusterName) {
+      this.clusterName = clusterName;
+      return this;
+    }
+
+    public Builder withMetricDomain(String metricDomain) {
+      this.metricDomain = metricDomain;
+      return this;
+    }
+
+    public Builder withMetricType(String metricType) {
+      this.metricType = metricType;
+      return this;
+    }
+
+    public Builder withMetricScope(String metricScope) {
+      this.metricScope = metricScope == null ? " " : metricScope;
+      return this;
+    }
+
+    public Builder withMetricName(String metricName) {
+      this.metricName = metricName;
+      return this;
+    }
+
+    public Builder withMetricAttribute(String metricAttribute) {
+      this.metricAttribute = metricAttribute;
+      return this;
+    }
+
+    public Builder withHost(String host) {
+      this.host = host;
+      return this;
+    }
+
+    public Builder withTs(DateTime ts) {
+      this.ts = ts;
+      return this;
+    }
+
+    public Builder withValue(double value) {
+      this.value = value;
+      return this;
+    }
+
+    public GenericMetric build() {
+      return new GenericMetric(this);
+    }
+  }
+}
--- a/src/server/src/main/java/io/cassandrareaper/jmx/ClusterFacade.java
+++ b/src/server/src/main/java/io/cassandrareaper/jmx/ClusterFacade.java
@@ -1,0 +1,673 @@
+/*
+ * Copyright 2018-2019 The Last Pickle Ltd
+ *
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.cassandrareaper.jmx;
+
+import io.cassandrareaper.AppContext;
+import io.cassandrareaper.ReaperApplicationConfiguration.DatacenterAvailability;
+import io.cassandrareaper.ReaperException;
+import io.cassandrareaper.core.Cluster;
+import io.cassandrareaper.core.Compaction;
+import io.cassandrareaper.core.DroppedMessages;
+import io.cassandrareaper.core.GenericMetric;
+import io.cassandrareaper.core.JmxStat;
+import io.cassandrareaper.core.MetricsHistogram;
+import io.cassandrareaper.core.Node;
+import io.cassandrareaper.core.Segment;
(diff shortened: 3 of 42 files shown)


src/server/src/main/java/io/cassandrareaper/ReaperApplicationConfiguration.java
```diff
@@ -508,7 +508,20 @@ public String toString() {
...
+     * @return true if we're in a collocated mode, false otherwise
+     */
+    public boolean isInCollocatedMode() {
+      switch (this) {
+        case EACH:
+          return true;
+        default:
+          return false;
+      }
+    }
```
michaelsembwever: i suspect this doesn't need to be part of this PR? (including the related patch chunk in `SegmentRunners.handlePotentialStuckRepairs(..)`)

src/server/src/main/java/io/cassandrareaper/jmx/JmxConnectionFactory.java
```diff
@@ -80,7 +76,7 @@ private void registerConnectionsGauge() {
     }
   }
 
-  protected JmxProxy connectImpl(Node node, int connectionTimeout) throws ReaperException, InterruptedException {
+  protected JmxProxy connectImpl(Node node) throws ReaperException, InterruptedException {
```
michaelsembwever, line 82: nice. this makes so much code more readable!

src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
```diff
@@ -46,6 +45,7 @@
 import com.google.common.util.concurrent.FutureCallback;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
+import jersey.repackaged.com.google.common.base.Preconditions;
```
michaelsembwever: wrong package

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
```diff
@@ -202,160 +208,205 @@ public CassandraStorage(ReaperApplicationConfiguration config, Environment envir
...
+        } catch (InvalidQueryException ex) {
+          throw new AssertionError(
+              "Failure preparing `SELECT… FROM repair_run WHERE… ALLOW FILTERING` should only happen on Cassandra-2",
+              ex);
+        }
       }
+      prepareMetricStatements();
+    } catch (RuntimeException e) {
+      LOG.error("Failed preparing Cassandra statements", e);
+      throw new ReaperException(e);
```
michaelsembwever: can we take the try-catch out of this PR please. it's generated such a huge diff that it's difficult to see what's actually changed here. i don't think it's specific to this PR either?

src/server/src/main/java/io/cassandrareaper/ReaperApplication.java
```diff
@@ -163,6 +164,7 @@ public void run(ReaperApplicationConfiguration config, Environment environment)
 
     context.repairManager = RepairManager.create(
         context,
+        ClusterFacade.create(context),
```
michaelsembwever: let's not inject a ClusterFacade into RepairManager (or RepairRunner or SegmentRunner or any other service class).

See comment in RepairManager for how to.

src/server/src/main/java/io/cassandrareaper/service/RepairManager.java
```diff
@@ -83,13 +87,21 @@ private RepairManager(
 
   public static RepairManager create(
       AppContext context,
+      ClusterFacade clusterFacade,
```
michaelsembwever: remove this parameter…

Instead, add a @VisibleForTesting package-protected static method like…

```
  private RepairManager(
      AppContext context,
      Supplier<ClusterFacade> clusterFacadeSupplier,
      ScheduledExecutorService executor,
      long repairTimeout,
      TimeUnit repairTimeoutTimeUnit,
      long retryDelay,
      TimeUnit retryDelayTimeUnit)  {

     …
  }

  @VisibleForTesting
  static RepairManager create(
      AppContext context,
      Supplier<ClusterFacade> clusterFacadeSupplier,
      ScheduledExecutorService executor,
      long repairTimeout,
      TimeUnit repairTimeoutTimeUnit,
      long retryDelay,
      TimeUnit retryDelayTimeUnit) {

    return new RepairManager(context, clusterFacadeSupplier, executor, repairTimeout, repairTimeoutTimeUnit, retryDelay, retryDelayTimeUnit);
  }

  // the default factory method delegates to the default facade behaviour
  public static RepairManager create(
      AppContext context,
      ScheduledExecutorService executor,
      long repairTimeout,
      TimeUnit repairTimeoutTimeUnit,
      long retryDelay,
      TimeUnit retryDelayTimeUnit) {

    return create(context, () -> ClusterFacade.create(context), executor, repairTimeout, repairTimeoutTimeUnit, retryDelay, retryDelayTimeUnit);
}
```

That allows you to pass in the mocks for testing, without (overly) polluting production code.

And if you find yourself having to inject ClusterFacade mocks for tests outside the package, it might be raising the question of why you require mocking something that's two levels away from the current layer you are *unit* testing.

src/server/src/test/resources/logback-test.xml
```diff
@@ -21,8 +21,18 @@
...
+        <Pattern>
+            %d{yyyy-MM-dd HH:mm:ss} [%thread] %-5level %logger{36} - %msg%n
+        </Pattern>
+    </layout>
+</appender>
+  
   <root>
     <level value="DEBUG" />
     <appender-ref ref="file"/>
+    <!-- <appender-ref ref="STDOUT"/> -->
```
michaelsembwever: i don't think any of this is part of the PR ?

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
```diff
@@ -104,10 +106,12 @@
   private final UUID leaderElectionId;
   private final AtomicBoolean successOrFailedNotified = new AtomicBoolean(false);
   private final AtomicBoolean completeNotified = new AtomicBoolean(false);
+  private final ClusterFacade clusterFacade;
 
 
   SegmentRunner(
       AppContext context,
+      ClusterFacade clusterFacade,
```
michaelsembwever: see https://github.com/thelastpickle/cassandra-reaper/pull/619#discussion_r259184312

src/server/src/main/java/io/cassandrareaper/service/SnapshotService.java
```diff
@@ -57,21 +58,23 @@
...
-  private SnapshotService(AppContext context, ExecutorService executor) {
+  private SnapshotService(AppContext context, ExecutorService executor, ClusterFacade clusterFacade) {
     this.context = context;
+    this.clusterFacade = clusterFacade;
     this.executor = new InstrumentedExecutorService(executor, context.metricRegistry);
   }
 
-  public static SnapshotService create(AppContext context, ExecutorService executor) {
-    return new SnapshotService(context, executor);
+  public static SnapshotService create(AppContext context, ExecutorService executor, ClusterFacade clusterFacade) {
```
michaelsembwever: see https://github.com/thelastpickle/cassandra-reaper/pull/619#discussion_r259184312

src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
```diff
@@ -57,34 +57,36 @@
...
   private final List<RingRange> parallelRanges;
   private final String metricNameForMillisSinceLastRepairPerKeyspace;
   private final String metricNameForMillisSinceLastRepair;
+  private final Optional<Cluster> cluster;
   private float repairProgress;
   private float segmentsDone;
   private float segmentsTotal;
 
-  RepairRunner(AppContext context, UUID repairRunId) throws ReaperException {
+  RepairRunner(AppContext context, UUID repairRunId, ClusterFacade clusterFacade) throws ReaperException {
```
michaelsembwever: see https://github.com/thelastpickle/cassandra-reaper/pull/619#discussion_r259184312

src/server/src/main/java/io/cassandrareaper/service/MetricsService.java
```diff
@@ -44,166 +41,49 @@
...
 
-  private MetricsService(AppContext context) {
+  private MetricsService(AppContext context, ClusterFacade clusterFacade) {
     this.context = context;
+    this.clusterFacade = clusterFacade;
   }
 
-  public static MetricsService create(AppContext context) {
-    return new MetricsService(context);
+  public static MetricsService create(AppContext context, ClusterFacade clusterFacade) {
```
michaelsembwever: see https://github.com/thelastpickle/cassandra-reaper/pull/619#discussion_r259184312

src/server/src/main/java/io/cassandrareaper/ReaperApplicationConfiguration.java
adejanovski: Reverted this change to make it part of the sidecar PR instead.

src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
adejanovski: Good catch

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
adejanovski: ok

src/server/src/test/resources/logback-test.xml
adejanovski: nope

src/server/src/main/java/io/cassandrareaper/ReaperApplication.java
adejanovski: ok, done

src/server/src/main/java/io/cassandrareaper/service/RepairManager.java
adejanovski: done

src/server/src/main/java/io/cassandrareaper/AppContext.java
```diff
@@ -49,6 +51,7 @@
   public JmxConnectionFactory jmxConnectionFactory;
   public ReaperApplicationConfiguration config;
   public MetricRegistry metricRegistry = new MetricRegistry();
+  public Set<String> accessibleDatacenters = Sets.newHashSet();
```
michaelsembwever: @alex i believe we can remove this.

it seems only to be used by tests. 
and those assertions in `ClusterFacadeTest` can be done instead against `context.jmxConnectionFactory.getHostConnectionCounters()` 

(sorry, i really should have spotted this in the first round of review!)

src/server/src/main/java/io/cassandrareaper/core/GenericMetric.java
```diff
@@ -0,0 +1,152 @@
+/*
+ * Copyright 2017-2017 Spotify AB
+ * Copyright 2017-2018 The Last Pickle Ltd
```
michaelsembwever: if this is a new file the copyright lines should be
```
 *
 * Copyright 2019-2019 The Last Pickle Ltd
```

And other touched files need to the second line to be updated to end with `-2019 The Last Pickle Ltd`

src/server/src/main/java/io/cassandrareaper/jmx/ClusterFacade.java
```diff
@@ -0,0 +1,674 @@
+/*
+ * Copyright 2017-2017 Spotify AB
+ * Copyright 2017-2019 The Last Pickle Ltd
```
michaelsembwever: see https://github.com/thelastpickle/cassandra-reaper/pull/619/files#r259656560

src/server/src/main/java/io/cassandrareaper/storage/cassandra/Migration018.java
```diff
@@ -0,0 +1,96 @@
+/*
+ * Copyright 2018-2018 The Last Pickle Ltd
```
michaelsembwever: see https://github.com/thelastpickle/cassandra-reaper/pull/619/files#r259656560

src/server/src/test/java/io/cassandrareaper/jmx/ClusterFacadeTest.java
```diff
@@ -0,0 +1,75 @@
+/*
+ * Copyright 2018-2018 The Last Pickle Ltd
```
michaelsembwever: see https://github.com/thelastpickle/cassandra-reaper/pull/619/files#r259656560

src/server/src/main/java/io/cassandrareaper/AppContext.java
adejanovski: it's used in ClusterFacade actually : https://github.com/thelastpickle/cassandra-reaper/pull/619/files#diff-3475cbd1fe679c34b1b40e562c7db6daR446

And it's a good first step to handle operations efficiently when it comes to use JMX or the backend. 
Please, let's keep it in this PR.

src/server/src/main/java/io/cassandrareaper/storage/cassandra/Migration018.java
adejanovski: This class doesn't belong in that PR, I removed it.

src/server/src/main/java/io/cassandrareaper/AppContext.java
michaelsembwever: > And it's a good first step to handle operations efficiently when it comes to use JMX or the backend.

Oh I missed that!
I agree that the code functionality in `ClusterFacade` should stay. It can be done though without adding the state to the content.

Could you still remove it, and add this method to `ClusterFacade` ?
```
@VisibleForTesting
boolean isDatacenterAccessible(String datacenter) {
  for (String host : context.jmxConnectionFactory.java.getHostConnectionCounters().getSuccessfullyConnectedHosts()) {
    if (EndpointSnitchInfoProxy.create(proxy).getDataCenter(host).equals(datacenter)) {
      return true;
    }
  }
  return false;
}
```

And in `HostConnectionCounters` the method `.getSuccessfullyConnectedHosts()` needs to be added which returns `Set<String>` for all the hosts that have one or more successful connections.

(I'm also curious as to whether there's a bug on https://github.com/thelastpickle/cassandra-reaper/pull/619/files#diff-739e34b5f2ca85ff72da4077afc1ea01R230 as it's only ever adding the local datacenter to the list, not the datacenter the host is in. My code above fixes that.)

adejanovski: My concern with your solution is that we'll have to issue a JMX call just to check if the node is accessible through JMX.
I could move `accessibleDatacenters` to `JmxConnectionFactory` instead if you don't like having it in `AppContext`.
Regarding https://github.com/thelastpickle/cassandra-reaper/pull/619/files#diff-739e34b5f2ca85ff72da4077afc1ea01R230 , it's storing the datacenter of the host we're connected to, which is the desired behavior (if we detect that the connection is possible to a host then we add its dc to the list of accessible ones). The call just avoids doing this at each successful connection and does it only if we previously failed or never connected to this node.

michaelsembwever: > Regarding https://github.com/thelastpickle/cassandra-reaper/pull/619/files#diff-739e34b5f2ca85ff72da4077afc1ea01R230 , it's storing the datacenter of the host we're connected to

too true, sorry about that.

> My concern with your solution is that we'll have to issue a JMX call just to check if the node is accessible through JMX.
> I could move `accessibleDatacenters` to `JmxConnectionFactory` instead if you don't like having it in `AppContext`.

The jmx call can be also avoided with caching its values. It would make sense to have a guava cache inside `EndpointSnitchInfoProxy.getDataCenter(host)` as it's a static and low-cardinality lookup.
But either way is fine for me, what's important for me is to go from the "global constant" to encapsulation.

I'm +1 on the PR once this has been solved.

michaelsembwever: > It would make sense to have a guava cache inside `EndpointSnitchInfoProxy.getDataCenter(host)` as it's a static and low-cardinality lookup.

In fact adding this type of logic was one of the reasons to create the separate AbcProxy classes above JmxProxy.

I've added the guava cache in https://github.com/thelastpickle/cassandra-reaper/commit/c7f72b4293bdb0abde5f0891d1d6a956e23d4156 if you wish to include it like that.

src/server/src/main/java/io/cassandrareaper/jmx/ClusterFacade.java
```diff
@@ -0,0 +1,674 @@
...
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.Lists;
+import org.apache.commons.lang3.tuple.Pair;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public final class ClusterFacade {
+  private static final Logger LOG = LoggerFactory.getLogger(ClusterFacade.class);
+  private static final String LOCALHOST = "127.0.0.1";
```
michaelsembwever: this constant doesn't appear to be used anywhere. (it's part of the subsequent PR that's to follow, i believe.)

src/server/src/main/java/io/cassandrareaper/ReaperApplicationConfiguration.java
```diff
@@ -508,7 +508,7 @@ public String toString() {
     /* We require jmx access to all nodes in the local datacenter */
     LOCAL,
     /* Each datacenter requires at minimum one reaper instance that has jmx access to all nodes in that datacenter */
-    EACH
+    EACH;
```
michaelsembwever: ??

src/server/src/main/java/io/cassandrareaper/AppContext.java
michaelsembwever: New commit https://github.com/thelastpickle/cassandra-reaper/commit/6d9bd1b73c6e4dce6d5bdb381c665c608b65f7aa
 that fixes the unit tests (`jmxProxyImpl.getUntranslatedHost()` needed mocking)

michaelsembwever (approved):
:shipit: 🚀

michaelsembwever:
Love the commit message detail, thanks for that @adejanovski

PR #622, opened 2019-03-01
Implement the sidecar mode

adejanovski:


src/server/src/main/java/io/cassandrareaper/ReaperApplication.java
```diff
@@ -67,6 +72,7 @@
 import io.prometheus.client.CollectorRegistry;
 import io.prometheus.client.dropwizard.DropwizardExports;
 import io.prometheus.client.exporter.MetricsServlet;
+import jersey.repackaged.com.google.common.collect.Sets;
```
michaelsembwever: wrong package

src/server/src/main/java/io/cassandrareaper/service/Heart.java
```diff
@@ -185,6 +213,44 @@ private void grabAndStoreNodeMetrics(IDistributedStorage storage, UUID runId, No
...
+      throws JsonProcessingException, MalformedObjectNameException, ReflectionException,
+          ReaperException, InterruptedException {
+    Node node = Node.builder().withClusterName(context.localClusterName).withHostname(context.localNodeAddress).build();
+    Set<CompositeData> activeStreams = ClusterFacade.create(context).listStreamsDirect(node);
+
+    ((IDistributedStorage) context.storage)
+        .storeStreams(context.localClusterName, context.localNodeAddress, activeStreams);
+
+    LOG.debug("Grabbing and storing streams for {}", context.localNodeAddress);
+  }
```
michaelsembwever: can we move these 3 methods down in to `MetricsService` ?

src/server/src/main/java/io/cassandrareaper/AppContext.java
```diff
@@ -49,6 +49,9 @@
   public JmxConnectionFactory jmxConnectionFactory;
   public ReaperApplicationConfiguration config;
   public MetricRegistry metricRegistry = new MetricRegistry();
+  public String localNodeAddress;
+  public String localClusterName;
+  public String localDatacenter;
```
michaelsembwever: for better encapsulation:
if the three new private methods added to `Heart` are moved down to `MetricsService`
then the three fields can also be moved to `MetricsService` (and made private).
this will make things a bit safer as well, as these fields are Optional, with values only present when in side-car mode.

`localNodeAddress` is slightly more difficult because `ClusterFacade` uses it, and its instantiation is based off the context and requires a jmx call.
But I would still remove it from here, duplicating it in `MetricsService` and `ClusterFacade`, bearing the cost of the extra jmx call (which could be alleviated with a guava cache, inside `JmxProxyImpl.getLocalEndpoint()`, as `host->address` is static information)

src/server/src/main/java/io/cassandrareaper/jmx/ClusterFacade.java
```diff
@@ -53,12 +55,23 @@
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Lists;
+import jersey.repackaged.com.google.common.base.Preconditions;
```
michaelsembwever: wrong package

src/server/src/main/java/io/cassandrareaper/jmx/ClusterFacade.java
```diff
@@ -53,12 +55,23 @@
...
 import org.slf4j.LoggerFactory;
 
+/**
+ * @author adejanovski
+ *
+ */
+/**
+ * @author adejanovski
+ *
+ */
```
michaelsembwever: remove author tags (especially duplicates :-)

src/server/src/main/java/io/cassandrareaper/AppContext.java
adejanovski: I would disagree here.
We can move them as static vars in `ClusterFacade` if you think it belongs there more than in `AppContext`, but I feel that's not what you're going after.
Duplicating the variables in both `ClusterFacade` and `MetricsService` would only add possible confusion and we're going to have to add it in `SegmentRunner` too. That's 3 places to init/maintain the same variable, which adds tech debt IMHO.

src/server/src/main/java/io/cassandrareaper/ReaperApplication.java
adejanovski: damn you Eclipse! :D

src/server/src/main/java/io/cassandrareaper/service/Heart.java
adejanovski: Sure thing, makes sense

src/server/src/main/java/io/cassandrareaper/AppContext.java
michaelsembwever: > We can move them as static vars in ClusterFacade if you think it belongs there more than in AppContext, but I feel that's not what you're going after.

Correct, static is no good, as it is related to the current `context.config`

> Duplicating the variables in both ClusterFacade and MetricsService would only add possible confusion…

No it shouldn't. It's a deterministic constant value, and encapsulation should be more important than DRY.

Let me take a closer look…

.travis.yml
```diff
@@ -84,6 +86,21 @@ jobs:
...
+      env: TEST_TYPE=sidecar CASSANDRA_VERSION=2.2.13 GRIM_MIN=1 GRIM_MAX=1
+    - stage: Sidecar Integration Tests
+      name: Cassandra 3.0.17
+      env: TEST_TYPE=sidecar CASSANDRA_VERSION=3.0.17 GRIM_MIN=1 GRIM_MAX=1
+    - stage: Sidecar Integration Tests
+      name: Cassandra 3.11.3
+      env: TEST_TYPE=sidecar CASSANDRA_VERSION=3.11.3 GRIM_MIN=1 GRIM_MAX=1
+    - stage: Sidecar Integration Tests
+      name: Cassandra 4.0
+      env: TEST_TYPE=sidecar CASSANDRA_VERSION=git:trunk GRIM_MIN=1 GRIM_MAX=1
```
michaelsembwever: you have to add the matching CASSANDRA_VERSION equivalents above as well.

src/server/src/test/resources/reaper-cassandra-sidecar2-at.yaml
```diff
@@ -0,0 +1,91 @@
+# Copyright 2017-2017 Spotify AB
+# Copyright 2017-2019 The Last Pickle Ltd
```
michaelsembwever: needs fixing

src/server/src/test/resources/reaper-cassandra-sidecar1-at.yaml
```diff
@@ -0,0 +1,98 @@
+# Copyright 2017-2017 Spotify AB
+# Copyright 2017-2019 The Last Pickle Ltd
```
michaelsembwever: needs fixing

src/server/src/test/java/io/cassandrareaper/jmx/ClusterProxyTest.java
```diff
@@ -0,0 +1,115 @@
+/*
+ * Copyright 2018-2018 The Last Pickle Ltd
```
michaelsembwever: needs fixing

src/server/src/test/java/io/cassandrareaper/acceptance/ReaperCassandraSidecarIT.java
```diff
@@ -0,0 +1,126 @@
+/*
+ * Copyright 2017-2017 Spotify AB
+ * Copyright 2017-2018 The Last Pickle Ltd
```
michaelsembwever: needs fixing

src/server/src/main/java/io/cassandrareaper/storage/cassandra/Migration018.java
```diff
@@ -0,0 +1,96 @@
+/*
+ * Copyright 2018-2018 The Last Pickle Ltd
```
michaelsembwever: needs fixing

src/server/src/main/java/io/cassandrareaper/storage/cassandra/Migration018.java
```diff
@@ -0,0 +1,96 @@
...
+  private static final String METRICS_V1_TABLE = "node_metrics_v1";
+  private static final String METRICS_V2_TABLE = "node_metrics_v2";
+  private static final String OPERATIONS_TABLE = "node_operations";
+
+  private Migration018() {
+  }
+
+  /**
+   * if Cassandra is running version less than 4.0
+   *  alter every table to set `dclocal_read_repair_chance` to zero
```
michaelsembwever: incorrect comment

src/server/src/main/java/io/cassandrareaper/storage/cassandra/Migration018.java
```diff
@@ -0,0 +1,96 @@
...
+          session.execute(
+                  "ALTER TABLE " + OPERATIONS_TABLE + " WITH compaction = {'class': 'TimeWindowCompactionStrategy', "
+                      + "'unchecked_tombstone_compaction': 'true', "
+                      + "'compaction_window_size': '1', "
+                      + "'compaction_window_unit': 'DAYS'}");
+
+          LOG.warn("{} was successfully altered to use TWCS.", OPERATIONS_TABLE);
+        }
+      } catch (RuntimeException e) {
+        LOG.error("Failed altering ");
```
michaelsembwever: can you make this `LOG.error("Failed altering tables to TWCS", e);` please.

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
```diff
@@ -248,6 +260,8 @@ private static void initializeAndUpgradeSchema(
         if (currentVersion <= 15) {
           Migration016.migrate(session, keyspace);
         }
+        // Switch metrics table to TWCS if possible
```
michaelsembwever: `// Switch metrics table to TWCS if possible, this is intentionally executed every startup`

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
```diff
@@ -410,6 +425,41 @@ private void prepareMetricStatements() {
         + " WHERE time_partition = ? AND run_id = ? AND node = ?");
     delNodeMetricsByNodePrepStmt = session.prepare("DELETE FROM node_metrics_v1"
         + " WHERE time_partition = ? AND run_id = ? AND node = ?");
+    storeMetricsPrepStmt
+        = session
+            .prepare(
+                "INSERT INTO node_metrics_v2 (cluster, metric_domain, metric_type, time_bucket, "
+                    + "host, metric_scope, metric_name, ts, metric_attribute, value) "
+                    + "VALUES(?, ?, ?, ?, ?, ?, ?, ?, ?, ?)")
+            .setConsistencyLevel(ConsistencyLevel.LOCAL_ONE);
```
michaelsembwever: everything is `ConsistencyLevel.LOCAL_ONE` by default.

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
```diff
@@ -410,6 +425,41 @@ private void prepareMetricStatements() {
...
+                    + "FROM node_metrics_v2 "
+                    + "WHERE metric_domain = ? and metric_type = ? and cluster = ? and time_bucket = ?")
+            .setConsistencyLevel(ConsistencyLevel.LOCAL_ONE);
+  }
+
+  private void prepareOperationsStatements() {
+    insertOperationsPrepStmt
+        = session.prepare(
+            "INSERT INTO node_operations(cluster, type, time_bucket, host, ts, data) "
+                + "values(?,?,?,?,?,?)");
```
michaelsembwever: does this need to be `QUORUM` ?

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
```diff
@@ -410,6 +425,41 @@ private void prepareMetricStatements() {
...
+
+    listOperationsForNodePrepStmt
+        = session.prepare(
+            "SELECT cluster, type, time_bucket, host, ts, data FROM node_operations "
+                + "WHERE cluster = ? AND type = ? and time_bucket = ? and host = ? LIMIT 1");
```
michaelsembwever: does this need to be `QUORUM` ?

src/server/src/main/java/io/cassandrareaper/storage/cassandra/Migration018.java
```diff
@@ -0,0 +1,96 @@
...
+import com.datastax.driver.core.VersionNumber;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public final class Migration018 {
+
+  private static final Logger LOG = LoggerFactory.getLogger(Migration018.class);
+  private static final String METRICS_V1_TABLE = "node_metrics_v1";
+  private static final String METRICS_V2_TABLE = "node_metrics_v2";
+  private static final String OPERATIONS_TABLE = "node_operations";
```
michaelsembwever: trivial: is there really any point in constants?

src/server/src/main/java/io/cassandrareaper/storage/cassandra/Migration018.java
```diff
@@ -0,0 +1,96 @@
...
+                      + "'compaction_window_unit': 'DAYS'}");
+
+          LOG.warn("{} was successfully altered to use TWCS.", METRICS_V2_TABLE);
+
+          LOG.warn("Altering {} to use TWCS...", OPERATIONS_TABLE);
+          session.execute(
+                  "ALTER TABLE " + OPERATIONS_TABLE + " WITH compaction = {'class': 'TimeWindowCompactionStrategy', "
+                      + "'unchecked_tombstone_compaction': 'true', "
+                      + "'compaction_window_size': '1', "
+                      + "'compaction_window_unit': 'DAYS'}");
```
michaelsembwever: the `node_operations` table's primary key does not look like it is suitable for TWCS…?

src/server/src/main/java/io/cassandrareaper/storage/cassandra/Migration018.java
```diff
@@ -0,0 +1,96 @@
...
+
+          LOG.warn("{} was successfully altered to use TWCS.", OPERATIONS_TABLE);
```
michaelsembwever: trivial: can we make all these log lines INFO instead?

src/server/src/main/java/io/cassandrareaper/storage/IDistributedStorage.java
```diff
@@ -56,4 +66,36 @@
...
+      throws JsonProcessingException;
+
+  List<Compaction> listCompactions(String clusterName, String host)
+      throws JsonProcessingException, IOException;
+
+  void storeStreams(String clusterName, String host, Set<CompositeData> activeStreams)
+      throws JsonProcessingException;
+
+  Set<CompositeData> listStreamingOperations(String clusterName, String host)
+      throws JsonProcessingException, IOException;
```
michaelsembwever, line 96: Some questions here:
 - do we really want to explode metric operations to individually defined interface methods here? 
 - do we want to expose `JsonProcessingException` and `IOException` here? (it goes against the grain of existing methods defined here, and i suspect is programmatic errors, not real checked exceptions defined by these behaviours) 

What about just the two methods:
 `void storeOperation(String clusterName, OpType opType, String host, String json);`
and
 `String getOperation(String clusterName, OpType opType, String host);`
?

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
```diff
@@ -1509,4 +1584,142 @@ public Snapshot getSnapshot(String clusterName, String snapshotName) {
...
+        : parseJson(streams.one().getString("data"), new TypeReference<Set<CompositeData>>(){});
+  }
+
+
+  private static <T> T parseJson(String json, TypeReference<T> ref) {
+    try {
+      return new ObjectMapper().readValue(json, ref);
+    } catch (IOException e) {
+      LOG.error("error parsing json", e);
+      throw new RuntimeException(e);
```
michaelsembwever: `RuntimeException` should be treated as an abstract class. this is an `IlegalStateException` (the json state stored is illegal)

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
```diff
@@ -1509,4 +1584,142 @@ public Snapshot getSnapshot(String clusterName, String snapshotName) {
...
+  @Override
+  public void storeStreams(String clusterName, String host, Set<CompositeData> activeStreams)
+      throws JsonProcessingException {
+    session.executeAsync(
+        insertOperationsPrepStmt.bind(
+            clusterName,
+            OP_STREAMING,
+            DateTime.now().toString(HOURLY_FORMATTER),
+            host,
+            DateTime.now().toDate(),
```
michaelsembwever, line 1747: this could be put into the CQL string as `now()`

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
```diff
@@ -1509,4 +1584,142 @@ public Snapshot getSnapshot(String clusterName, String snapshotName) {
...
+  @Override
+  public void storeCompactions(String clusterName, String host, List<Compaction> activeCompactions)
+      throws JsonProcessingException {
+    session.executeAsync(
+        insertOperationsPrepStmt.bind(
+            clusterName,
+            OP_COMPACTION,
+            DateTime.now().toString(HOURLY_FORMATTER),
+            host,
+            DateTime.now().toDate(),
```
michaelsembwever, line 1747: this could be put into the CQL string as `now()`

src/server/src/main/resources/db/cassandra/018_sidecar_mode.cql
```diff
@@ -0,0 +1,33 @@
...
+    metric_domain text,
+    metric_type text,
+    time_bucket text,
+    host text,
+    metric_scope text,
+    metric_name text,
+    ts timestamp,
+    metric_attribute text,
+    value double,
+    PRIMARY KEY ((cluster, metric_domain, metric_type, time_bucket), host, metric_scope, metric_name, ts, metric_attribute)
```
michaelsembwever: doesn't the `ts` column need to be the first clustering key?

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
```diff
@@ -410,6 +425,41 @@ private void prepareMetricStatements() {
...
+                    + "host, metric_scope, metric_name, ts, metric_attribute, value) "
+                    + "VALUES(?, ?, ?, ?, ?, ?, ?, ?, ?, ?)")
+            .setConsistencyLevel(ConsistencyLevel.LOCAL_ONE);
+    getMetricsForHostPrepStmt
+        = session
+            .prepare(
+                "SELECT cluster, metric_domain, metric_type, time_bucket, host, "
+                    + "metric_scope, metric_name, ts, metric_attribute, value "
+                    + "FROM node_metrics_v2 "
+                    + "WHERE metric_domain = ? and metric_type = ? and cluster = ? and time_bucket = ? and host = ?")
```
michaelsembwever: if the `ts` column were the first clustering column (for the sake of TWCS): then the range operator against `since` could be included here?

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
```diff
@@ -933,6 +983,31 @@ private static RepairSegment createRepairSegmentFromRow(Row segmentRow) {
     return Optional.empty();
   }
 
+  @Override
+  public Optional<RepairSegment> getNextFreeSegmentForRanges(
+      UUID runId, Optional<RingRange> parallelRange, List<RingRange> ranges) {
```
michaelsembwever: nit-pick: one parameter per line please.

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
```diff
@@ -933,6 +983,31 @@ private static RepairSegment createRepairSegmentFromRow(Row segmentRow) {
...
+    List<RepairSegment> segments
+        = Lists.<RepairSegment>newArrayList(getRepairSegmentsForRun(runId));
+    Collections.shuffle(segments);
+
+    for (RepairSegment seg : segments) {
+      if (seg.getState().equals(State.NOT_STARTED) && withinRange(seg, parallelRange)) {
```
michaelsembwever, line 1033: how do we handle  `seg.getState().equals(State.NOT_STARTED)`  changing to STARTED by someone else?

src/server/src/main/java/io/cassandrareaper/service/MetricsService.java
```diff
@@ -93,4 +107,43 @@ public static MetricsService create(AppContext context) {
 
     return metrics;
   }
+
```
michaelsembwever, line 127: can the exception list in the following three methods be cleaned up?
i'm thinking it should be only `ReaperException`, and other exceptions resolved as either `IllegalArgumentException` or `IllegalStateException` ?

src/server/src/main/java/io/cassandrareaper/jmx/ClusterFacade.java
```diff
@@ -339,6 +354,40 @@ public String getCassandraVersion(Cluster cluster, Collection<String> endpoints)
...
+          .stream()
+          .forEach(entry -> {
+            if (entry.getValue().contains(localEndpoint)) {
+              localRanges.add(
+                  new RingRange(new BigInteger(entry.getKey().get(0)), new BigInteger(entry.getKey().get(1))));
+            }
+          });
+
+      return localRanges;
+    } catch (RuntimeException e) {
```
michaelsembwever: i'm curious as to what unchecked exceptions are we catching here?

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
```diff
@@ -398,6 +410,27 @@ private boolean runRepair() {
     return true;
   }
 
+  private void updateNodeMetrics(JmxProxy coordinator) {
```
michaelsembwever: who calls this method?

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
```diff
@@ -295,6 +295,15 @@ private boolean runRepair() {
...
 
+      // If we're using a distributed storage, we need to synchronize with other Reaper instances
+      // So we don't start too many segments at the same time
+      if (!lockSegmentRunners()) {
+        LOG.debug(
+            "Not allowed to run the segment for now as another Reaper holds the lock for repair run {}. "
+            + "Will try again later",
+            segment.getRunId());
+        return false;
+      }
```
michaelsembwever: i'm keen to minimise the need for this locking, as described in https://github.com/thelastpickle/cassandra-reaper/pull/601#commitcomment-31906618

by redo-ing and offering https://github.com/thelastpickle/cassandra-reaper/compare/sidecar-mode...mck/sidecar-mode

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
michaelsembwever: confirmed that no one shall call it.

src/server/src/main/java/io/cassandrareaper/AppContext.java
michaelsembwever: `SegmentRunner` does not use any of these three fields. (Usage was from a method that can be removed: https://github.com/thelastpickle/cassandra-reaper/pull/622/files#r262444370

michaelsembwever: So, I'm thinking it makes sense to do the following…
 - `localClusterName` and `localDatacenter` are only used in `MetricsService`, and so can be moved to, and initialised, there. 
 - `localNodeAddress` is used in many places, but only when in side-car mode.
 - define it as `volatile String localNodeAddress = null;`
 - add the following getter…
```
public String getLocalNodeAddress() {
    Preconditions.checkState(context.config.isInSidecarMode());
    return localNodeAddress;
}
```

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
michaelsembwever: work-in-progress here: https://github.com/thelastpickle/cassandra-reaper/pull/624

the idea is that PR#624 would get merged to master first.

michaelsembwever: #624 has been merged to master.

off-line discussion agreed that side-car could add locking on top on the condition that it was optimistic: that is was only used when #624 was insufficient; and would be logged when required. This way over time the locking can be removed if it is proven to never be used.

src/ci/install.sh
```diff
@@ -58,6 +58,35 @@ case "${TEST_TYPE}" in
...
+          echo 'phi_convict_threshold: 16' >> ~/.ccm/test/node$i/conf/cassandra.yaml
+          if  echo "$CASSANDRA_VERSION" | grep -q "trunk"  ; then
+            sed -i 's/start_rpc: true//' ~/.ccm/test/node$i/conf/cassandra.yaml
+            echo '-Dcassandra.max_local_pause_in_ms=15000' >> ~/.ccm/test/node$i/conf/jvm-server.options
+            sed -i 's/#-Dcassandra.available_processors=number_of_processors/-Dcassandra.available_processors=2/' ~/.ccm/test/node$i/conf/jvm-server.options
+          else
+            sed -i 's/start_rpc: true/start_rpc: false/' ~/.ccm/test/node$i/conf/cassandra.yaml
+          fi
+        done
+        ;;
```
michaelsembwever, line 73: what's different about this to `"ccm"` ?
(it seems to be a fair bit to duplicate just for the switch case in `script.sh` ?)

src/server/src/main/java/io/cassandrareaper/AppContext.java
michaelsembwever: @adejanovski what's the status on this?

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
```diff
@@ -116,6 +124,8 @@
 
   private static final Logger LOG = LoggerFactory.getLogger(CassandraStorage.class);
 
+  private static boolean firstAttempt = true;
```
michaelsembwever: needs to be an `AtomicBoolean`

src/server/src/main/java/io/cassandrareaper/storage/cassandra/Migration018.java
adejanovski: Done

src/server/src/main/java/io/cassandrareaper/storage/cassandra/Migration018.java
adejanovski: To avoid repeating string values ¯\_(ツ)_/¯

src/server/src/main/java/io/cassandrareaper/storage/cassandra/Migration018.java
adejanovski: Everything expires after 5 minutes there, and the table is bucketed by time.
We only keep the latest value per bucket, and TWCS will ease reclaiming space for expired data and limit tombstone reads.

src/server/src/main/java/io/cassandrareaper/storage/cassandra/Migration018.java
adejanovski: Sure thing, not sure why I used `warn` in the first place.

src/server/src/main/java/io/cassandrareaper/storage/cassandra/Migration018.java
michaelsembwever: trivial: if everything expire after 5 minutes, then won't the table effectively only ever be doing STCS (because sstables never live to the second time_window)?

.travis.yml
adejanovski: Done, thanks

src/server/src/main/java/io/cassandrareaper/storage/cassandra/Migration018.java
adejanovski: 1 day is indeed way too big of a time window.
I reduced it to 1 hour which matches the bucketing window.

src/server/src/main/java/io/cassandrareaper/storage/IDistributedStorage.java
adejanovski: I made the suggested changes 👍

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
adejanovski: Well, I can't use the cql `now()` to fill in the hour that's used in the time bucket column, so I could end up with inconsistent times between the two fields.

src/server/src/main/resources/db/cassandra/018_sidecar_mode.cql
adejanovski: Since I have time buckets, I'm not using `ts` in the read queries. This clustering also allows me to get either a specific node or all nodes in the cluster in a single query.

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
adejanovski: it could but I'd lose the ability to get a metric for all nodes in the cluster. The `time_bucket` is more important for TWCS than `ts`'s position is here IMHO.
I've changed the TWCS window size to match the hourly bucketing scheme.

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
adejanovski: Done

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
adejanovski: In this case the segment will have a leader already and the repair manager will move to the next one.

src/server/src/main/java/io/cassandrareaper/service/MetricsService.java
adejanovski: Indeed, only `ReaperException` was necessary after all. I've cleaned up the `InterruptedException` from the signature.

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
adejanovski: Sure thing, it's been moved from here anyway as Json parsing shouldn't happen in the DAO.

src/server/src/main/java/io/cassandrareaper/jmx/ClusterFacade.java
adejanovski: right, I've removed the try/catch block.

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
adejanovski: Yup, deleted.

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
adejanovski: done

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
michaelsembwever: Wouldn't it be cheaper to first check against `STARTED` ? (it is meant as an optimisation)

src/server/src/main/java/io/cassandrareaper/jmx/ClusterFacade.java
michaelsembwever: has this been fixed? (i'm still seeing duplicate author tags in github)

src/server/src/main/java/io/cassandrareaper/jmx/ClusterFacade.java
michaelsembwever: did you forget to push that change?

src/server/src/main/java/io/cassandrareaper/service/MetricsService.java
michaelsembwever: was that change pushed?

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
```diff
@@ -300,6 +302,15 @@ private boolean runRepair() {
 
       LazyInitializer<Set<String>> busyHosts = new BusyHostsInitializer(cluster);
 
+      // If we're using a distributed storage, we need to synchronize with other Reaper instances
+      // So we don't start too many segments at the same time
+      if (!lockSegmentRunners()) {
```
michaelsembwever: this needs to be moved to happen after line 334 ( https://github.com/thelastpickle/cassandra-reaper/pull/622/files#diff-e6aa6503be815f976c3ed7a845621f4aR334 )
with a check first against `STARTED`

and the log line raised to `WARN` (so we get the feedback over time how required this locking is)

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
```diff
@@ -311,6 +322,7 @@ private boolean runRepair() {
         } catch (InterruptedException e) {
           LOG.debug("Interrupted while sleeping after a segment was postponed... weird stuff...");
         }
+        releaseSegmentRunners();
```
michaelsembwever: this `releaseSegmentRunners ()` won't be needed once https://github.com/thelastpickle/cassandra-reaper/pull/622/files#r284190582 is moved to be inside the `try {} finally {}` block.

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
```diff
@@ -619,7 +639,7 @@ private boolean nodesReadyForNewRepair(
 
     Collection<String> nodes = getNodesInvolvedInSegment(dcByNode);
     String dc = EndpointSnitchInfoProxy.create(coordinator).getDataCenter();
-    boolean requireAllHostMetrics = DatacenterAvailability.ALL == context.config.getDatacenterAvailability();
+    boolean requireAllHostMetrics = DatacenterAvailability.LOCAL != context.config.getDatacenterAvailability();
```
michaelsembwever, line 634: isn't this changing the way `EACH` works? and isn't that outside of the scope of this work?

src/server/src/main/java/io/cassandrareaper/service/Heart.java
```diff
@@ -83,7 +94,9 @@ synchronized void beat() {
   synchronized void beatMetrics() {
     if (context.storage instanceof IDistributedStorage
             && ReaperApplicationConfiguration.DatacenterAvailability.EACH
-                == context.config.getDatacenterAvailability()) {
+                == context.config.getDatacenterAvailability()
+        || ReaperApplicationConfiguration.DatacenterAvailability.SIDECAR
+            == context.config.getDatacenterAvailability()) {
```
michaelsembwever: curious, why doesn't `LOCAL` do `beatMetrics()` ?

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
```diff
@@ -571,26 +587,29 @@ private void handlePotentialStuckRepairs(LazyInitializer<Set<String>> busyHosts,
...
+
+      return !context.config.getDatacenterAvailability().isInCollocatedMode()
           ? Optional.empty()
           : getRemoteNodeMetrics(node, nodeDc);
     });
   }
 
   private Optional<NodeMetrics> getRemoteNodeMetrics(String node, String nodeDc) {
-    Preconditions.checkState(DatacenterAvailability.ALL != context.config.getDatacenterAvailability());
+    Preconditions.checkState(context.config.getDatacenterAvailability().isInCollocatedMode());
```
michaelsembwever, line 592: shouldn't this be `Preconditions.checkState(!context.config.getDatacenterAvailability().isInCollocatedMode());` ?

src/server/src/main/java/io/cassandrareaper/ReaperApplicationConfiguration.java
```diff
@@ -508,7 +548,24 @@ public String toString() {
...
+    EACH,
+    /* Sets Reaper in sidecar mode where each Cassandra node has a collocated Reaper instance */
+    SIDECAR;
+
+
+    /**
+     * Check if the current datacenter availability mode is to have collocation between Reaper and a DC/node.
+     * @return true if we're in a collocated mode, false otherwise
+     */
+    public boolean isInCollocatedMode() {
```
michaelsembwever, line 560: curious, if there's a more intuitive name for this… like `isOnlyLocalDcJmx()` or  inverted as `allowCrossDcJmx()` , wdyt?

src/server/src/main/java/io/cassandrareaper/AppContext.java
adejanovski: it's done, pushed and tested

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
adejanovski: With the lock that is applied in the sidecar mode, this kind of conflict couldn't even happen actually. For the other modes, the optimization would be very marginal I think, not worth the time to code and test.

src/ci/install.sh
adejanovski: it is removing all the parts around setting up JMX authentication, since it's not activated in local JMX mode.

src/server/src/main/java/io/cassandrareaper/service/RepairManager.java
```diff
@@ -73,7 +73,7 @@ private RepairManager(
       long repairTimeout,
       TimeUnit repairTimeoutTimeUnit,
       long retryDelay,
-      TimeUnit retryDelayTimeUnit)  {
+      TimeUnit retryDelayTimeUnit) throws ReaperException, InterruptedException  {
```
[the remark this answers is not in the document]
adejanovski: Indeed, forgot to remove those.

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
adejanovski: Not as far as I can tell.
`EACH` always required to have all host metrics in order to proceed with the segment. `LOCAL`  is the only mode where all hosts metrics aren't required (just those from the local DC).

src/server/src/main/java/io/cassandrareaper/service/Heart.java
adejanovski: Because LOCAL is used in the setup where you have one Reaper instance in a DC and no Reaper instance in the other DCs, so there's no metric request going on. Reaper will reach out directly to the nodes that are in the local DC and proceed.

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
adejanovski: No, we want to make sure that we're only trying to grab metrics through the backend when we're in either `EACH` or `SIDECAR`.
The precondition looks good to me.

src/server/src/main/java/io/cassandrareaper/ReaperApplicationConfiguration.java
adejanovski: It wouldn't really match the sidecar mode IMHO. It's not just about being local to a DC but also a node.

src/server/src/main/java/io/cassandrareaper/jmx/ClusterFacade.java
adejanovski: Right, it's still there. Removing it :)

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
michaelsembwever: Why can't `LOCAL` grab metrics?

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
michaelsembwever: Cool, thanks. (The check against`allLocalDcHostsChecked` is done in the `okToRepairSegment(..)` method.)

src/ci/install.sh
michaelsembwever: agreed to put lines 67-87 into a separate function that both switch cases can use.

src/server/src/main/java/io/cassandrareaper/service/Heart.java
michaelsembwever: I'm not sure that's necessarily true. `LOCAL` can also mean that availability and running repairsis preferred (over prevention of node overloading) when all reapers in a datacenter go down.

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
michaelsembwever: i will provide a PR for this.

src/server/src/main/java/io/cassandrareaper/ReaperApplicationConfiguration.java
michaelsembwever: that i don't really grok. `EACH` is not restricted to only be local to a node, and `LOCAL` can also be local to a node.

wallrj:
@adejanovski and @michaelsembwever We've been experimenting with this SIDECAR branch in https://github.com/sky-uk/cassandra-operator/pull/97.

We were originally planning to write a Kubernetes operator for Cassandra Reaper, but this seems like a much simpler deployment method, thanks.

Is there anything we can do to help get this merged? We'll test of course, but could also help write documentation if you'd like.

We're particularly interested to hear whether you consider SIDECAR mode to  be compatible with the reaper web UI. How will login work if there are multiple reapers in the cluster?

adejanovski:
Hi @wallrj, 

Both tests and documentation contributions will be greatly appreciated, thanks!!
The web UI will still work as expected, but if you switch from one Reaper instance to the other, you'll most probably need to re-authenticate.
I'll run some tests to verify this and report back.

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
adejanovski: `LOCAL` will grab metrics directly from the nodes in the local DC. It won't go through the backend to get metrics for remote nodes that aren't directly reachable through JMX.

src/server/src/main/java/io/cassandrareaper/service/Heart.java
adejanovski: I think considering `LOCAL` for this use case goes against what the documentation states and would complexify both the code and the tests.
I advise on keeping things as described in the docs so that we don't add confusion over what's already not very easy to understand.
I firmly believe that once the `SIDECAR` mode is out we'll want to reduce the `datacenterAvailability` options anyway to further simplify both usage and the code (I know we still disagree for now on this, but... 😉 ).

michaelsembwever: > I advise on keeping things as described in the docs so that we don't add confusion over what's already not very easy to understand.

The docs state for `LOCAL` that `In addition, metrics can be collected asynchronously through the Apache Cassandra storage.`
ref: http://cassandra-reaper.io/docs/configuration/reaper_specific/

This matches my understanding. The difference between `LOCAL` and `EACH` should be simplified to how failure (to jmx connect) is handled. To also also include the definition of how/what it does adds complexity.
That is, the simplest way to explain, as I have understood it, is: they work in the same way but `EACH` stops working if no reaper is running in a datacenter to ensure no C* node is overloaded, while `LOCAL` provides better Reaper availability by continuing to run in presuming the C* nodes in a datacenter where no Reaper is running are ok.

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
michaelsembwever: The docs do state for `LOCAL` that `In addition, metrics can be collected asynchronously through the Apache Cassandra storage.`

see https://github.com/thelastpickle/cassandra-reaper/pull/622#discussion_r287893882

src/server/src/main/java/io/cassandrareaper/service/Heart.java
michaelsembwever: > I firmly believe that once the SIDECAR mode is out we'll want to reduce the datacenterAvailability options anyway to further simplify both usage and the code.

Yes, I disagree. A cluster of 200 nodes means 100GB and 100+ virtual CPU usages resources dedicated to Reaper. That's a significant cost compared to running one Reaper process in a tiny virtual server.
And it is an opinionated approach, presuming that operators can colocate processes. Different platform container systems and setups vary a lot and I don't think we should be making this presumption.
It also breaks the ability for one Reaper UI to register and visualise all your Cassandra clusters in one UI.
These are huge trade-offs on the assumptions jmx connectivity is difficult for everyone. As we have seen many users have no problems with the jmx connectivity, even between datacenters around the world.

Promotion of the side-car mode should be, imho, via new functionality and features we add that are only available when using side-car mode.

src/ci/install.sh
adejanovski: Done

src/server/src/main/java/io/cassandrareaper/service/Heart.java
adejanovski: The doc is misleading as it is not how `LOCAL` works : https://github.com/thelastpickle/cassandra-reaper/blob/1.4/src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java#L653

We're only going through getting the metrics from the backend (and requesting them) if we're using `EACH` and it should stay that way.

It also matches the multi DC architecture docs: http://cassandra-reaper.io/docs/usage/multi_dc/

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
adejanovski: Same as above, the code matches `LOCAL`'s current behavior in 1.4

src/server/src/main/java/io/cassandrareaper/ReaperApplicationConfiguration.java
adejanovski: I can't think of a better name for now as `LOCAL` is also a form of collocation (to a single DC).

src/server/src/main/java/io/cassandrareaper/service/Heart.java
michaelsembwever: > The doc is misleading as it is not how LOCAL works

No, that doc has been intentionally written that way. There's different setups and scenarios that validate `LOCAL` in different ways.

From http://cassandra-reaper.io/docs/usage/multi_dc/

> Single Reaper instance with JMX accessible for a single DC only

^that is not accurate. It provides only one possible scenario of when `LOCAL` is desired. Reaper in `LOCAL` mode can also have multiple Reaper instances running in different datacenters.
`LOCAL` is not about definitely not having a Reaper instance, so much as about how to handle what happens when the (or all) Reaper instance(s) in a datacenter stop.

The idea is that it is about the trade-off between protecting Cassandra nodes from repairs that can overload them versus prioritising that repairs run even if some Reaper instances go down.
It allows the operator to decide what is most important in their environment: healthy nodes versus repaired data.
I don't think we can be opinionated about the Cassandra clusters, datacenters can be a mixture of online (OLTP) and offline (OLAP). In OLAP datacenters latency is less of a concern, so node load is less of a concern, while batch jobs depend upon repaired data for their consistency model.

 > how LOCAL works : https://github.com/thelastpickle/cassandra-reaper/blob/1.4/src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java#L653

That was a bug which has been fixed, if you look in master. It was also an important fix for our integration tests in CircleCI that run against multiple datacenters. We had a discussion about how this impacted CircleCI in the past but I'm guessing the relevance to this discussion wasn't obvious then.

Bear in mind it was I that argued for and introduced `LOCAL` as a compliment to `EACH`, and my understanding of the reasoning for it hasn't changed.
I suspect that the metrics requests need to know how to back-off, just like jmx hosts are blacklisted… so to cover the case where `LOCAL` is used where one or more datacenters don't (and won't ever) have a Reaper instance running.

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
michaelsembwever: https://github.com/thelastpickle/cassandra-reaper/pull/669

src/server/src/main/java/io/cassandrareaper/service/Heart.java
michaelsembwever: There's a discussion to be had here about a) improving `Heart` to do `beatMetrics()` when in `LOCAL` mode,  and b) better documenting what the different advantages `EACH` and `LOCAL` modes offer.

For now, as `Heart` did not already do `beatMetrics()` for `LOCAL` mode, this discussion can be taken out of the PR.

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
michaelsembwever: But it's not the behaviour in `master`.  The code was fixed intentionally in master, as 1.4 is identified as incorrect.

Similar to the decision in https://github.com/thelastpickle/cassandra-reaper/pull/622#discussion_r289596995
 this code should remain as it was in master beforehand, and we revisit the `LOCAL` vs `EACH` mode advantages outside of this PR.

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
```diff
@@ -92,6 +92,7 @@
   private static final ExecutorService METRICS_GRABBER_EXECUTOR = Executors.newFixedThreadPool(10);
   private static final long METRICS_POLL_INTERVAL_MS = TimeUnit.SECONDS.toMillis(5);
   private static final long METRICS_MAX_WAIT_MS = TimeUnit.MINUTES.toMillis(2);
+  private static final UUID GLOBAL_LOCK_ID = UUID.fromString("35446a30-86ed-11e9-bc42-526af7764f64");
```
michaelsembwever: Given that `SegmentRunner. processTriggeredSegment(..)` is called when the lock is held, does this not have the consequence that only segment will ever run in Reaper at any one time, regardless of how many clusters are registered in Reaper or how many Reaper instances there are?

adejanovski: Good point, it'll actually be only one segment that can *start* at once, not run.
Still it's suboptimal, I'll adjust this so we have one lock per cluster.

adejanovski: I've changed it to a cluster wide lock

michaelsembwever: > Good point, it'll actually be only one segment that can start at once, not run.

Are you sure about that?
Because it looks like `SegmentRunner. processTriggeredSegment(..)`  waits until the segment is complete.

adejanovski: Yes, because one SegmentRunner instance gets created per segment.

michaelsembwever: > Yes, because one SegmentRunner instance gets created per segment.

 Yes, but the global/cluster lock is not released until after `SegmentRunner. processTriggeredSegment(..)`. The instance of the SegmentRunner is irrelevant.

I think the lock needs to go back to how it was. (And that eventually we will see that we really don't need the lock, and that transactions are here and generally a really bad way of trying to balance/throttle the current throughput of repairs in the cluster.)

adejanovski: I disagree, we just need to release the lock earlier (before getting into `processTriggeredSegment()`).

michaelsembwever: I'm  `-1` on expanding the lock to be global or cluster because it breaks the premise of introducing the ~double-locking-idiom.

The purpose of the `WARN` log on https://github.com/thelastpickle/cassandra-reaper/pull/622/files#diff-e6aa6503be815f976c3ed7a845621f4aR297 was to prove or disprove the need for the LWT locking.
But by making the locking global or cluster-based the locking is going to be overly eager and won't prove or disprove anything. 

Can the lock be returned to how it was, so to guarantee the transition from `RepairSegment.State.STARTED` to `RepairSegment.State.RUNNING`.
The second aspect to this code (ie the result of the `canRepair(..)` method) is the nodes that the token range has replicas for are not already busy/overloaded. That is already not bound by the LWT transaction/lock: the jmx calls (and how these results are determined) are based on metrics that arn't strict regardless of how broad the lock is made.

michaelsembwever: > Can the lock be returned to how it was, so to guarantee the transition from RepairSegment.State.STARTED to RepairSegment.State.RUNNING.

Previously the lock was on the whole repair. That was the best we could do because we need to lock against all the nodes involved in the token range at hand. There's no way to achieve that without executing three separate LWTs locks. 

I'm `-0` on a cluster-wide lock, and would rather it returned to a `repairId` lock as it was. (ie i'm ok with the cluster-wide lock, but i would rather the repair based lock.)
Otherwise, either way, the first `WARN` log it not going to be specific enough to tell us the lock is never needed, as either repair or cluster based it is too broad. The second `WARN` log on line 310 should tell us though.

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
michaelsembwever: Thanks.

michaelsembwever:
LGTM. Though I haven't tested the latest changes. 

Two questions:
 - a quick look at CircleCi is warranted as many of the builds are failing a lot more than usual. CircleCi in master has few flaky tasks (apart from `c_4-0_flapping-reapers__all_nodes_reachable` which consistently fails for an yet unidentified reason),
 - Cassandra 4.0 integration tests work ok in CircleCI, so why are they failing in Travis? maybe it's something else, maybe the last commit to ignore 4.0 tests is the wrong approach??

It would not surprise me if there's a few small things to fix still, but it would be great to see this merged: a lot of great work and rework has gone into this! So if the circleci failures look to be small stuff do merge.

michaelsembwever:
CircleCI tests were fixed/improved.

michaelsembwever (approved):

PR #624, opened 2019-03-05
Improve how repair segments check whether they can start or not

michaelsembwever:
Improve how repair segments check whether they can start or not
 - simplify (reduce) logic in SegmentRunner.canRepair(..)
 - incremental repairs really checking for running repairs will have to use metrics grabbing as well
 - introduce State.STARTED
 - use strict consistency when changing a repair segment from NOT_STARTED to STARTED state

ref: https://github.com/thelastpickle/cassandra-reaper/pull/622/files#r262413082

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/core/RepairSegment.java
+++ b/src/server/src/main/java/io/cassandrareaper/core/RepairSegment.java
@@ -130,9 +130,11 @@ public final class RepairSegment {
   }
 
   public enum State {
+    // don't change order or insert items (ordial is used in db)
     NOT_STARTED,
     RUNNING,
-    DONE
+    DONE,
+    STARTED
   }
 
   @JsonPOJOBuilder(buildMethodName = "build", withPrefix = "with")
--- a/src/server/src/main/java/io/cassandrareaper/jmx/ClusterFacade.java
+++ b/src/server/src/main/java/io/cassandrareaper/jmx/ClusterFacade.java
@@ -440,9 +440,7 @@ public final class ClusterFacade {
    */
   public boolean nodeIsAccessibleThroughJmx(String nodeDc, String node) {
     return DatacenterAvailability.ALL == context.config.getDatacenterAvailability()
-        || DatacenterAvailability.LOCAL == context.config.getDatacenterAvailability()
-        || (DatacenterAvailability.EACH == context.config.getDatacenterAvailability()
-            && context.jmxConnectionFactory.getAccessibleDatacenters().contains(nodeDc));
+        || context.jmxConnectionFactory.getAccessibleDatacenters().contains(nodeDc);
   }
 
   /**
--- a/src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
+++ b/src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
@@ -36,6 +36,7 @@ import java.lang.management.ManagementFactory;
 import java.lang.management.OperatingSystemMXBean;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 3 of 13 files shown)


src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
```diff
@@ -441,29 +446,32 @@ private void processTriggeredSegment(final RepairSegment segment, final JmxProxy
...
+        default:
+          // Something went wrong on the coordinator node and we never got the RUNNING notification
+          // or we are in an undetermined state.
+          // Let's just abort and reschedule the segment.
+          LOG.info(
+              "Repair command {} on segment {} never managed to start within timeout.",
+              this.repairNo,
+              segmentId);
+          segmentFailed.set(true);
+          abort(resultingSegment, coordinator);
```
michaelsembwever, line 449: this diff chunk isn't strictly part of the PR, it's just replacing the if-else blocks with a switch statement.

src/server/src/main/java/io/cassandrareaper/jmx/ClusterFacade.java
```diff
@@ -440,9 +440,7 @@ public String getLocalEndpoint(Node node) throws ReaperException, InterruptedExc
    */
   public boolean nodeIsAccessibleThroughJmx(String nodeDc, String node) {
     return DatacenterAvailability.ALL == context.config.getDatacenterAvailability()
-        || DatacenterAvailability.LOCAL == context.config.getDatacenterAvailability()
-        || (DatacenterAvailability.EACH == context.config.getDatacenterAvailability()
-            && context.jmxConnectionFactory.getAccessibleDatacenters().contains(nodeDc));
+        || context.jmxConnectionFactory.getAccessibleDatacenters().contains(nodeDc);
```
michaelsembwever, line 443: `LOCAL` mode is not allowed to make cross-dc jmx connections either. only `ALL` can do that.

src/server/src/main/java/io/cassandrareaper/storage/IDistributedStorage.java
```diff
@@ -32,12 +32,8 @@
 
   boolean takeLead(UUID leaderId);
 
-  boolean takeLead(UUID leaderId, int ttl);
-
   boolean renewLead(UUID leaderId);
 
-  boolean renewLead(UUID leaderId, int ttl);
-
```
michaelsembwever, line 40: i took this stuff out because it's not used yet. if needed it should be included later in the side-car PR.

src/server/src/main/java/io/cassandrareaper/jmx/ClusterFacade.java
michaelsembwever: making this change caused the assertion on the following line (in multiple-reaper integration tests) to fail…
https://github.com/thelastpickle/cassandra-reaper/blob/master/src/server/src/test/java/io/cassandrareaper/acceptance/BasicSteps.java#L1521

src/server/src/main/java/io/cassandrareaper/core/RepairSegment.java
```diff
@@ -130,9 +130,11 @@ public Builder reset() {
   }
 
   public enum State {
+    // don't change order or insert items (ordial is used in db)
     NOT_STARTED,
     RUNNING,
-    DONE
+    DONE,
+    STARTED
```
michaelsembwever, line 137: UI needs updating to for this.

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
```diff
@@ -363,7 +342,6 @@ private boolean runRepair() {
 
             if (0 != repairNo) {
               processTriggeredSegment(segment, coordinator, repairNo);
```
michaelsembwever, line 345: can `processTriggeredSegment(..)` deal with cancelling too many running segments? (currently cassandra only supports cancelling all repair sessions)

@adejanovski , it was discussed and thought that it wasn't a good idea to cancel all repairs, but we do that already in `handlePotentialStuckRepairs(..)`, ref https://github.com/thelastpickle/cassandra-reaper/blob/f6c451b/src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java#L623-L644

So more to the point… when would we not want to cancel all repairs on a node if we saw that if either
 - was running an unrecognised repair, or
 - too many repairs.

A high maximum threshold on pending compactions and load could also be introduced where we cancel all repairs on a node.

src/server/src/main/java/io/cassandrareaper/storage/IDistributedStorage.java
adejanovski: I pushed a commit to address this.

src/server/src/main/java/io/cassandrareaper/core/RepairSegment.java
adejanovski: I pushed a commit to address this.

michaelsembwever: thanks @adejanovski 

still working on fixing the upgrade integration test. just haven't had as much time as i'd hope for this week :(

michaelsembwever: was addresses in #628 (the upgrade integration tests have been disabled for now)

PR #620, opened 2019-02-27
Add newline at the end of shiro.ini to fix REAPER_AUTH_USER

viossat:
When using `REAPER_AUTH_USER`/`REAPER_AUTH_PASSWORD` on Docker, an empty line is needed at the end of `shiro.ini`, otherwise the last line looks like `[users]user = password`, throwing this exception:

```
java.lang.IllegalArgumentException: There is no filter with name 'password' to apply to chain [[users]user] in the pool of available Filters.  Ensure a filter with that name/path has first been registered with the addFilter method(s).
	at org.apache.shiro.web.filter.mgt.DefaultFilterChainManager.addToChain(DefaultFilterChainManager.java:265)
	at org.apache.shiro.web.filter.mgt.DefaultFilterChainManager.createChain(DefaultFilterChainManager.java:148)
	at org.apache.shiro.web.config.IniFilterChainResolverFactory.createChains(IniFilterChainResolverFactory.java:192)
	at org.apache.shiro.web.config.IniFilterChainResolverFactory.buildChains(IniFilterChainResolverFactory.java:127)
	at org.apache.shiro.web.config.IniFilterChainResolverFactory.createInstance(IniFilterChainResolverFactory.java:80)
	at org.apache.shiro.web.config.IniFilterChainResolverFactory.createInstance(IniFilterChainResolverFactory.java:43)
	at org.apache.shiro.config.IniFactorySupport.createInstance(IniFactorySupport.java:123)
	at org.apache.shiro.util.AbstractFactory.getInstance(AbstractFactory.java:47)
	at org.apache.shiro.web.env.IniWebEnvironment.createFilterChainResolver(IniWebEnvironment.java:187)
	at org.apache.shiro.web.env.IniWebEnvironment.configure(IniWebEnvironment.java:102)
	at org.apache.shiro.web.env.IniWebEnvironment.init(IniWebEnvironment.java:92)
	at org.secnod.dropwizard.shiro.ShiroBundle.createFilter(ShiroBundle.java:62)
	at org.secnod.dropwizard.shiro.ShiroBundle.run(ShiroBundle.java:43)
	at io.cassandrareaper.ReaperApplication$1.run(ReaperApplication.java:127)
	at io.cassandrareaper.ReaperApplication$1.run(ReaperApplication.java:123)
	at io.dropwizard.setup.Bootstrap.run(Bootstrap.java:200)
	at io.dropwizard.cli.EnvironmentCommand.run(EnvironmentCommand.java:42)
[...]

PR #625, opened 2019-03-07
Small fix for webUI auth.

b13n1u:
There is no next line before the username so it creates wrong entry.  For example for user: foo
The entry in shiro.ini looks like this:
[users]foo = bar

Causing: 
```
java.lang.IllegalArgumentException: There is no filter with name '"foo"' to apply to chain [[users]"foo"] in the pool of available Filters.
```

Adding next line in cat solves it :)


michaelsembwever:
@b13n1u did this not get fixed already in https://github.com/thelastpickle/cassandra-reaper/pull/620 and https://github.com/thelastpickle/cassandra-reaper/commit/7f46e2054005cf7da16bda214d404db179df42cc ?

b13n1u:
Yes, indeed it got fixed. I'll close this PR then.

Issue #629, opened 2019-03-18
1.4.1 doesn't repair Cassandra 1.2.19/2.0 anymore


rborer:
Hi again,

I've just bumped reaper to version 1.4.1 and just discovered that it doesn't repair my cluster on Cassandra 1.2.19 anymore (yeah I know, pretty old version...).

Here are the logs:

```
WARN   [2019-03-18 15:09:54,207] [staging-cluster:00000000-0000-0076-0000-000000000000:00000000-0000-16ed-0000-000000000000] i.c.s.SegmentRunner - Failed to connect to a coordinator node for segment 00000000-0000-16ed-0000-00000000
0000 
java.lang.reflect.UndeclaredThrowableException: null
        at com.sun.proxy.$Proxy56.getCompactionParameters(Unknown Source)
        at io.cassandrareaper.jmx.JmxProxyImpl.getTablesForKeyspace(JmxProxyImpl.java:406)
        at io.cassandrareaper.service.RepairUnitService.getTablesToRepair(RepairUnitService.java:94)
        at io.cassandrareaper.service.SegmentRunner.runRepair(SegmentRunner.java:269)
        at io.cassandrareaper.service.SegmentRunner.run(SegmentRunner.java:143)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at com.google.common.util.concurrent.TrustedListenableFutureTask$TrustedFutureInterruptibleTask.runInterruptibly(TrustedListenableFutureTask.java:111)
        at com.google.common.util.concurrent.InterruptibleTask.run(InterruptibleTask.java:58)
        at com.google.common.util.concurrent.TrustedListenableFutureTask.run(TrustedListenableFutureTask.java:75)
        at com.codahale.metrics.InstrumentedScheduledExecutorService$InstrumentedRunnable.run(InstrumentedScheduledExecutorService.java:241)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:180)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:293)
[...]

PR #631, opened 2019-03-19
Restore compatibility with Cassandra 1.2 & 2.0

rborer:
Fixes #629.

As [`ColumnFamilyStoreMBean.getCompactionParameters()`](https://github.com/apache/cassandra/blob/9bb75358dfdf1b9824f9a454e70ee2c02bc64a45/src/java/org/apache/cassandra/db/ColumnFamilyStoreMBean.java#L331) is only available since Cassandra 2.1, this PR restore compatibility with older versions which doesn't expose this method. 

On such versions, the default compaction strategy is used (`TimeWindowCompactionStrategy` & `DateTieredCompactionStrategy` strategies are not available for those versions).

This PR also configures Travis CI to run integration tests with Cassandra versions 1.2.19 & 2.0.17 (hopefully since I have no clue how to test this).

I've manually deployed this version on my cluster running Cassandra 1.2.19 and so far it's running repairs properly.

PR #632, opened 2019-03-19
Travis: enable integration tests under versions 1.2.19 & 2.0.17

rborer:
As discussed under #629 we may want to extend integration tests to run with older Cassandra versions.

~~As of now the tests will fail since they attempt to start an incremental repair which is only supported from versions 2.1 onwards.~~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,7 +52,13 @@ jobs:
   - env: TEST_TYPE=upgrade CASSANDRA_VERSION=git:trunk CO="-t ~@all_nodes_reachable"
   include:
     - stage: Integration Tests
-      name: Memory, H2 and Postgresql
+      name: Memory, H2 and Postgresql on Cassandra 1.2.19
+      env: TEST_TYPE=ccm CASSANDRA_VERSION=1.2.19 CUCUMBER_OPTIONS="--tags ~@cassandra_2_1_onwards"
+    - stage: Integration Tests
+      name: Memory, H2 and Postgresql on Cassandra 2.0.17
+      env: TEST_TYPE=ccm CASSANDRA_VERSION=2.0.17 CUCUMBER_OPTIONS="--tags ~@cassandra_2_1_onwards"
+    - stage: Integration Tests
+      name: Memory, H2 and Postgresql on Cassandra 2.1.20
       env: TEST_TYPE=ccm CASSANDRA_VERSION=2.1.20
     - stage: Integration Tests
       name: Cassandra 2.1.20
@@ -70,7 +76,7 @@ jobs:
       name: Cassandra 4.0
       env: TEST_TYPE=ccm CASSANDRA_VERSION=git:trunk GRIM_MIN=1 GRIM_MAX=1
     - stage: Distributed Reaper Integration Tests
-      name: Two Reapers on  Cassandra 2.1.20
+      name: Two Reapers on Cassandra 2.1.20
       env: TEST_TYPE=ccm CASSANDRA_VERSION=2.1.20 GRIM_MIN=2 GRIM_MAX=2
     - stage: Distributed Reaper Integration Tests
       name: Two Reapers on Cassandra 2.2.13
--- a/src/server/src/test/java/io/cassandrareaper/acceptance/BasicSteps.java
+++ b/src/server/src/test/java/io/cassandrareaper/acceptance/BasicSteps.java
@@ -49,6 +49,7 @@ import com.datastax.driver.core.Host;
 import com.datastax.driver.core.Session;
 import com.datastax.driver.core.SocketOptions;
 import com.datastax.driver.core.VersionNumber;
+import com.datastax.driver.core.exceptions.AlreadyExistsException;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Maps;
@@ -934,6 +935,8 @@ public final class BasicSteps {
   @And("^the last added repair has twcs table \"([^\"]*)\" in the blacklist$")
   public void the_last_added_repair_has_twcs_table_in_the_blacklist(String twcsTable) throws Throwable {
     synchronized (BasicSteps.class) {
+      final VersionNumber lowestNodeVersion = getCassandraVersion();
+
       RUNNERS
           .parallelStream()
           .forEach(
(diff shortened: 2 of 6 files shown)


rborer:
@michaelsembwever I have extracted Travis changes to run integration tests with versions 1.2 & 2.0 here instead of #631.

.travis.yml
```diff
@@ -51,6 +51,12 @@ jobs:
   - env: TEST_TYPE=upgrade CASSANDRA_VERSION=git:trunk CO="-t @all_nodes_reachable"
   - env: TEST_TYPE=upgrade CASSANDRA_VERSION=git:trunk CO="-t ~@all_nodes_reachable"
   include:
+    - stage: Integration Tests
+      name: Memory, H2 and Postgresql
```
michaelsembwever: Each name also should be updated, to include Cassandra version, so it's obvious which build is succeeding/failing. (my bad for not seeing this before.)

michaelsembwever:
Thanks @rborer 

I believe the best way forward with this is to create and add cucumber version tags `@Cassandra_1_2` and `@Cassandra_2_0` to the scenarios in `integration_reaper_functionality.feature`, and add to the `CUCUMBER_OPTIONS` environment variable for the new CI tasks: `"-t ~@Cassandra_1_2 -t ~@Cassandra_2_0"`.

.travis.yml
rborer: Definitely make sense :+1:

.travis.yml
```diff
@@ -52,13 +52,13 @@ jobs:
   - env: TEST_TYPE=upgrade CASSANDRA_VERSION=git:trunk CO="-t ~@all_nodes_reachable"
   include:
     - stage: Integration Tests
-      name: Memory, H2 and Postgresql
+      name: Cassandra 1.2.19: memory, H2 and Postgresql
```
michaelsembwever: to nitpick: would this read better in the format: `Memory, H2, and Postgresql (repairing Cassandra 1.2.19)` ?
it's awkward… i'm not sure how to best distinguish between what storage versus what c* version we're testing against…

rborer: Or `Memory, H2 and Postgresql on Cassandra 1.2.19` to better match the other tests like `Two Reapers on Cassandra 2.1.20` and `Four Flapping Reaper on Cassandra 2.2.13`?

rborer:
@michaelsembwever I have tried to adapt cucumber tests into 228cab0335a5b39fc12cf721f43a6cc57075af40 . Since I have 0 experience on cucumber I had to read a little bit, and I've tried a different approach than [what you proposed earlier](https://github.com/thelastpickle/cassandra-reaper/pull/632#issuecomment-474706923). I have added a tag `@cassandra_2_1_onwards` that we can add to integration tests that supports Cassandra 2.1 and superior. Such tests are excluded when running versions prior to this one.

Since I really don't know cucumber at all let me know if it make sense this way?

michaelsembwever:
> @michaelsembwever I have tried to adapt cucumber tests into 228cab0 . Since I have 0 experience on cucumber I had to read a little bit, and I've tried a different approach than what you proposed earlier. I have added a tag @cassandra_2_1_onwards that we can add to integration tests that supports Cassandra 2.1 and superior. Such tests are excluded when running versions prior to this one.

Looks perfect to me. Thanks.

michaelsembwever (approved):
LGTM.

Could you…
 - fix the integration test names (to your suggestion),
 - squash the 3 commits to one, and
 - rebase off master.

Then I'll merge with great fanfare and happiness :-)

rborer:
@michaelsembwever done, happy to help :-)

michaelsembwever:
Uh :(
Test failures 
 - https://travis-ci.org/thelastpickle/cassandra-reaper/jobs/517681061
 - https://travis-ci.org/thelastpickle/cassandra-reaper/jobs/517681062

If you don't get to look at these first, I should be able to tomorrow…

rborer:
I had a very quick look, I think the tests under 1.2.19 are failing because support for lightweight transactions has been added in 2.0 only (https://www.datastax.com/dev/blog/lightweight-transactions-in-cassandra-2-0).

As such the tables cannot be created with `IF NOT EXISTS` here: 

https://github.com/thelastpickle/cassandra-reaper/blob/0dfb4281c18aaeec1f2969345560bab1a3442610/src/server/src/test/java/io/cassandrareaper/acceptance/BasicSteps.java#L1665-L1668

:crying_cat_face:

michaelsembwever:
>  I had a very quick look, I think the tests under 1.2.19 are failing because support for lightweight transactions has been added in 2.0 only 

👍 

Want to merge this into your branch? https://github.com/rborer/cassandra-reaper/pull/2

michaelsembwever:
I have a bit more work to do in fixing that schema setup…

michaelsembwever:
> I have a bit more work to do in fixing that schema setup…

Another commit for you here: https://github.com/rborer/cassandra-reaper/commit/499db2523ad3d8d707ae6499346fd583acacdcd4

src/server/src/test/java/io/cassandrareaper/acceptance/BasicSteps.java
```diff
@@ -1695,7 +1717,11 @@ private static void createTable(String keyspaceName, String tableName) {
         }
       }
 
-      tmpSession.execute(createTableStmt);
+      try {
+        if (null == tmpSession.getCluster().getMetadata().getKeyspace(keyspaceName).getTable(tableName)) {
+          tmpSession.execute(createTableStmt);
```
rborer, line 1724: @michaelsembwever I'm wondering: the usual recommended way when writing integration tests with a remote datastore is to ensure that the datastore is empty at the beginning of each test (a test data should not leak in another test).

Here don't we take the risk to already have a keyspace & table from a previous test (but with a different configuration) and then re-use it? If the configuration is not the same it may yield strange failures.

Of course it's no little change, so maybe a better candidate for another PR. What do you think?

rborer:
@michaelsembwever 1.2 is still failing the background setup as it tries to create a table `booya_twcs` with a DateTiered/TimeWindow compaction strategy which is unavailable.

https://github.com/thelastpickle/cassandra-reaper/pull/632/files#diff-f7cde3f37a58bd10458e98869b49222aR21

`
 And cluster "test" has keyspace "booya" with tables "booya_twcs"
 `

And I think it is done this way to check that tables with such compaction strategies are blacklisted by default.

michaelsembwever:
@rborer 
> 1.2 is still failing the background setup as it tries to create a table booya_twcs with a DateTiered/TimeWindow compaction strategy which is unavailable.

Can you replace last commit with this one? 
https://github.com/thelastpickle/cassandra-reaper/commit/9da773e522956364c312970976e3086110fdd411

EDIT: still working on this one, the blacklisting tests have to be skipped in older C* versions…

src/server/src/test/java/io/cassandrareaper/acceptance/BasicSteps.java
michaelsembwever: i think it's the same test just running the setup code concurrently… but haven't looked into it any further.

michaelsembwever:
One more time please… I picked up a bug in the integration tests…

New commit is https://github.com/thelastpickle/cassandra-reaper/commit/d05b7ece5028e64f18531f820062ba0178827aad

Feel free to squash these additional commits into your initial PR commit.

michaelsembwever:
Integration tests now work repairing a Cassandra 1.2 cluster, but continue to fail repairing a Cassandra 2.0 cluster.
So far I haven't been able to see the error, maybe a jmx configuration error (seems it can't connect to `127.0.0.2`)
ref: https://travis-ci.org/thelastpickle/cassandra-reaper/jobs/518560161

michaelsembwever:
@rborer, I am looking into this. Have found a few bugs (not from you) and am working on the same branch name but in the origin repository.

UPDATE: the logs/errors are a bit more manageable to read now, but there's still an obvious bug in jmx connections to other hosts in Cassandra 2.0.17

michaelsembwever:
Silly me, Cassandra 2.0 tests were *also* failing because disabling of `LOCAL_JMX` has to be done differently: https://github.com/apache/cassandra/blob/cassandra-2.0/conf/cassandra-env.sh#L269

michaelsembwever:
Merged manually with https://github.com/thelastpickle/cassandra-reaper/commit/3dc22554e9458178d15485a5ecc5dea5cd119a7a

Thanks so much for the effort in making this happen @rborer !! It is of huge value to quality assurance of the project.

rborer:
You're welcome, I'm happy to help on this extremely useful project!

PR #636, opened 2019-03-27
Fix LongCollectionSqlType.getValue()

mmonemali:
Simple fix for LongCollectionSqlType.getValue(), it was always returning empty list because of [this](https://github.com/thelastpickle/cassandra-reaper/blob/7bf92be033a01dedebb175f4059531d5e45958a5/src/server/src/main/java/io/cassandrareaper/core/RepairSchedule.java#L260)

Moved the class outside of RepairSchedule.java and added a regression test.

We found this bug while investigating an issue with duplicate repair runs happening at the same time for a keyspace. While this won't fix the concurrency problem but it will reduce the window where duplicate runs could happen when [comparing](https://github.com/thelastpickle/cassandra-reaper/blob/7bf92be033a01dedebb175f4059531d5e45958a5/src/server/src/main/java/io/cassandrareaper/service/SchedulingManager.java#L161) RepairSchedules.


michaelsembwever:
looking into this…

src/server/src/main/java/io/cassandrareaper/storage/postgresql/LongCollectionSqlType.java
```diff
@@ -0,0 +1,39 @@
...
+
+import java.util.Collection;
+
+import com.google.common.collect.Lists;
+
+/**
+ * This is required to be able to map in generic manner into Postgres array
+ * types through JDBI.
+ */
+public final class LongCollectionSqlType {
```
michaelsembwever: this class can't be in this package, as it then creates a package tangle. 
Other classes in Reaper do not reference this package, with the exception of `PostgresStorage`, which it is private to.

could you please move it back to being an inner class in `RepairSchedule`.

michaelsembwever:
The fix looks good, and thanks for catching that rather clumsy mistake @mmonemali !

The only adjustment needed before merging this is the `LongCollectionSqlType` (and its test `LongCollectionSqlTypeTest`) needs to be moved back to the `io.cassandrareaper.core` package, and preferable with  `LongCollectionSqlType`  as an inner class to `ReaperSchedule` as it was previously.

michaelsembwever (changes requested):
LongCollectionSqlType needs to be moved back to where it was.

michaelsembwever:
Perfect! Thanks so much @mmonemali for the fix, the test, and the extra work re-arranging it. very much appreciated :)

michaelsembwever:
Merged to `1.4` and `master` with https://github.com/thelastpickle/cassandra-reaper/commit/cf34fbb1ac7d3043119ee6920ab0a148fccbc402

PR #639, opened 2019-04-03
Add automated retries when connecting to the storage backend

### LINKED ISSUE #448: Retry on failed connection
With 1.1.0 on Oracle Linux 7 installed from RPM (init.d script), Reaper dies on a system reboot. 

This is because it is configured to connect to 127.0.0.1 and Cassandra may not be up yet.  I'd like to see some configurable retry logic at least during startup.

adejanovski:
fixes #448

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
+++ b/src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
@@ -53,6 +53,7 @@ import java.util.SortedSet;
 import java.util.UUID;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.stream.Collectors;
 
 import javax.annotation.Nullable;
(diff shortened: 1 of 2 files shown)


src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
```diff
@@ -115,6 +115,8 @@
 
   private static final Logger LOG = LoggerFactory.getLogger(CassandraStorage.class);
 
+  private static boolean firstAttempt = true;
```
michaelsembwever: is this always accessed by the same thread? (eg in tests)
maybe make it an `AtomicBoolean` instead?

michaelsembwever (approved):
LGTM. just the one comment regarding thread-safety.

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
```diff
@@ -183,6 +186,11 @@ public CassandraStorage(ReaperApplicationConfiguration config, Environment envir
 
     // https://docs.datastax.com/en/developer/java-driver/3.5/manual/metrics/#metrics-4-compatibility
     cassandraFactory.setJmxEnabled(false);
+    if (!CassandraStorage.firstAttempt.get()) {
```
michaelsembwever: nit-picking: this should be 
```
if (!firstAttempt.compareAndSet(true, false)) {
      // If there's been a past connection attempt, metrics are already registered
      cassandraFactory.setMetricsEnabled(false);
}

cassandra = cassandraFactory.build(environment);
```
(and to nit-pick on the nit-pick… i'd be calling the static variable `firstInstantiation` as it's not an attempt: `there is no try. do or do not.`)

adejanovski: fair enough, I've applied the suggested changes.

michaelsembwever:
CircleCI had a high number of failures on the flapping integration tests. Am waiting to see if they improve/fix.

PR #641, opened 2019-04-04
WIP: Fix ClassCastException in integration tests

aahutsal:
Only these two are mostly affected. Stuck on the problem screenshotted in @gitcoinbot discussion thread

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,7 +43,7 @@ aliases:
             # fallback to using the latest cache if no exact match is found
             - v1-dependencies-
 
-        - run: 
+        - run:
             name: Download dependencies
             command: |
               mvn -B dependency:go-offline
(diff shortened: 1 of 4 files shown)


aahutsal:
I've considered 2 ways to fix `ClassCastException`.

1. Exclude JARs that contains duplicated classes from surefire classpath. Worked well initially, but stuck at this [point](https://github.com/gutsal-arsen/cassandra-reaper/blob/f57bc8ab1f9d705261e03d5aac067802564cbe2d/src/server/pom.xml#L613). If I keep excluding JARs we get ClassNotFound exception. So, that's not the solution.
2. Another solution would be to use custom ClassLoader as you started to do, but this is not going to work as well, as Jersey/Dropwizard escapes from the sandbox in some places calling `ClassLoader.getSystemClassloader()` like [it happens here](https://github.com/dropwizard/dropwizard/blob/a21eb86e30433d8655c496ba835cfa876bf962d5/dropwizard-core/src/test/java/io/dropwizard/server/DefaultServerFactoryTest.java#L50). Once this happens incorrect versions of classes are loaded and we get `ClassCastException` again.

Since neither works, I'd keep my research trying to include missed JARs into shaded JAR, then run surefire with `useSystemClassloader=true` and see how it performs.

@michaelsembwever Don't merge this PR yet until I finish my research. Hopefully it will contain really working integration tests once I'm done.

adejanovski:
@aahutsal, the work done here is mostly incompatible with the current state of the code.
I'll close this PR as I'm cleaning up our board. Thanks for what was done and let us know if you're planning more work on Reaper's CI.

aahutsal:
@adejanovski if you have some new requests - I'd be glad to review.

PR #642, opened 2019-04-09
Upgrade integration tests to provide the forward-compatibility on the yaml configuration

michaelsembwever:
Tell the jackson object mapper not to fail when deserialising unknown properties…

This is done with a call to `bootstrap.getObjectMapper().disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES)`. But this has to be done via reflection in `ReaperTestJettyRunner`.

### DIFF
--- a/src/server/src/test/java/io/cassandrareaper/acceptance/ReaperTestJettyRunner.java
+++ b/src/server/src/test/java/io/cassandrareaper/acceptance/ReaperTestJettyRunner.java
@@ -101,7 +101,8 @@ public final class ReaperTestJettyRunner {
         Object methodFilter = Proxy.newProxyInstance(
             appClass.getClassLoader(),
             new Class[] {methodFilterCls},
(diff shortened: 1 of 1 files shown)


aahutsal:
@michaelsembwever mistake
you should repalce  `disable(...)` with `configure(..., false)`, so it reads:

`mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)`

Reference: https://stackoverflow.com/questions/5455014/ignoring-new-fields-on-json-objects-using-jackson

michaelsembwever:
>  you should repalce disable(...) with configure(..., false), so it reads:
> `mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)`

fixed (and forced pushed). Thanks @gutsal-arsen .

Issue #603, opened 2019-01-17
Error while repairing – invalid RunState and timestamps


ynsdl:
Hi,

I have a cassandra cluster setup with 3 nodes. On each node JMX authentication is enabled

When ı start to repair, ı see some errors on cassandra-reaper.log

What does it means ?

ERROR  [2019-01-17 07:20:54,850] [ReaperApplication-scheduler] i.c.ReaperApplication - Failed purging repair runs from history 
java.lang.NullPointerException: null
        at io.cassandrareaper.service.PurgeService.lambda$purgeRepairRunsByHistoryDepth$1(PurgeService.java:94)
        at java.util.TimSort.countRunAndMakeAscending(TimSort.java:355)
        at java.util.TimSort.sort(TimSort.java:220)
        at java.util.Arrays.sort(Arrays.java:1512)
        at java.util.ArrayList.sort(ArrayList.java:1454)
        at io.cassandrareaper.service.PurgeService.purgeRepairRunsByHistoryDepth(PurgeService.java:93)
        at io.cassandrareaper.service.PurgeService.purgeDatabase(PurgeService.java:71)
        at io.cassandrareaper.ReaperApplication.lambda$schedulePurge$1(ReaperApplication.java:308)
        at com.codahale.metrics.InstrumentedScheduledExecutorService$InstrumentedRunnable.run(InstrumentedScheduledExecutorService.java:235)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:308)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:180)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:294)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
        at java.lang.Thread.run(Thread.java:745)

ynsdl:
Hi,

cassandra-reaper working in memory so ı have restarted cassandra-reaper and send a new repair request to cassandra. So ı cant see last repair history but ı can share new repair history.

storageType: memory

[...]

PR #651, opened 2019-04-24
Fix RepairRun timestamps, honouring different run states

michaelsembwever:
ref:
 - https://github.com/thelastpickle/cassandra-reaper/issues/603

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/storage/cassandra/Migration008.java
+++ b/src/server/src/main/java/io/cassandrareaper/storage/cassandra/Migration008.java
@@ -20,9 +20,9 @@ package io.cassandrareaper.storage.cassandra;
 
 import com.datastax.driver.core.Session;
 
-public final class Migration011 {
+public final class Migration008 {
 
-  private Migration011() {
+  private Migration008() {
   }
 
   /**
--- a/src/server/src/main/java/io/cassandrareaper/storage/cassandra/Migration012.java
+++ b/src/server/src/main/java/io/cassandrareaper/storage/cassandra/Migration012.java
@@ -1,6 +1,6 @@
 /*
  * Copyright 2014-2017 Spotify AB
- * Copyright 2016-2018 The Last Pickle Ltd
+ * Copyright 2016-2019 The Last Pickle Ltd
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -27,9 +27,9 @@ public final class Migration012 {
 
   /**
    * fix segment start and end times in the repair_run table.
-   *  delegates to Migration009 as that does everything already.
+   *  delegates to FixRepairSegmentTimestamps as that does everything already.
    */
   public static void migrate(Session session) {
-    Migration009.migrate(session);
+    FixRepairSegmentTimestamps.migrate(session);
   }
 }
--- a/src/server/src/main/resources/db/h2/V12_0_0__fix_repair_run_timestamps.sql
+++ b/src/server/src/main/resources/db/h2/V12_0_0__fix_repair_run_timestamps.sql
@@ -13,11 +13,30 @@
 --  See the License for the specific language governing permissions and
 --  limitations under the License.
 --
+-- Fix timestamp inconsistencies in the repair_run table
 --
--- Support for custom cluster properties
---
 
-ALTER TABLE cluster 
-ADD properties TEXT;
+UPDATE repair_run
+SET start_time = NULL
+WHERE start_time is NOT NULL
+AND state =  'NOT_STARTED';
 
+UPDATE repair_run
+SET end_time = NULL
+WHERE end_time IS NOT NULL
+AND state NOT IN ('ERROR', 'DONE', 'ABORTED', 'PAUSED');
 
+UPDATE repair_run
+SET end_time = start_time
+WHERE end_time IS NULL
+AND state IN ('ERROR', 'DONE', 'ABORTED', 'PAUSED');
+
+UPDATE repair_run
+SET start_time = end_time
+WHERE start_time is NULL
+AND end_time IS NOT NULL;
+
+UPDATE repair_run
+SET pause_time = start_time
+WHERE pause_time IS NULL
+AND state = 'PAUSED';
--- a/src/server/src/main/resources/db/postgres/V12_0_0__fix_repair_run_timestamps.sql
+++ b/src/server/src/main/resources/db/postgres/V12_0_0__fix_repair_run_timestamps.sql
@@ -1,0 +1,42 @@
+--
+--  Copyright 2019-2019 The Last Pickle Ltd
+--
+--  Licensed under the Apache License, Version 2.0 (the "License");
+--  you may not use this file except in compliance with the License.
+--  You may obtain a copy of the License at
+--
+--      http://www.apache.org/licenses/LICENSE-2.0
+--
+--  Unless required by applicable law or agreed to in writing, software
+--  distributed under the License is distributed on an "AS IS" BASIS,
+--  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+--  See the License for the specific language governing permissions and
+--  limitations under the License.
+--
+-- Fix timestamp inconsistencies in the repair_run table
+--
+
+UPDATE repair_run
+SET start_time = CAST(NULL as TIMESTAMP WITH TIME ZONE)
+WHERE start_time is NOT NULL
+AND state = 'NOT_STARTED';
+
+UPDATE repair_run
+SET end_time = CAST(NULL as TIMESTAMP WITH TIME ZONE)
+WHERE end_time IS NOT NULL
+AND state NOT IN ('ERROR', 'DONE', 'ABORTED', 'PAUSED');
+
+UPDATE repair_run
+SET end_time = start_time
+WHERE end_time IS NULL
+AND state IN ('ERROR', 'DONE', 'ABORTED', 'PAUSED');
+
+UPDATE repair_run
+SET start_time = end_time
+WHERE start_time is NULL
+AND end_time IS NOT NULL;
+
+UPDATE repair_run
+SET pause_time = start_time
+WHERE pause_time IS NULL
+AND state = 'PAUSED';
--- a/src/server/src/main/java/io/cassandrareaper/core/RepairRun.java
+++ b/src/server/src/main/java/io/cassandrareaper/core/RepairRun.java
@@ -1,6 +1,6 @@
 /*
  * Copyright 2014-2017 Spotify AB
- * Copyright 2016-2018 The Last Pickle Ltd
+ * Copyright 2016-2019 The Last Pickle Ltd
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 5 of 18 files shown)


src/server/src/main/java/io/cassandrareaper/storage/cassandra/Migration008.java
```diff
@@ -0,0 +1,35 @@
...
+import com.datastax.driver.core.Session;
+
+public final class Migration008 {
+
+  private Migration008() {
+  }
+
+  /**
+   * fix segment start and end times in the repair_run table.
+   *  delegates to Migration009 as that does everything already.
```
SesquipedalianDefenestrator:
```suggestion
   *  delegates to FixRepairSegmentTimestamps as that does everything already.
```

src/server/src/main/java/io/cassandrareaper/storage/cassandra/Migration011.java
```diff
@@ -30,6 +30,6 @@ private Migration011() {
    *  delegates to Migration009 as that does everything already.
```
SesquipedalianDefenestrator:
```suggestion
   *  delegates to FixRepairSegmentTimestamps as that does everything already.
```

src/server/src/main/java/io/cassandrareaper/storage/cassandra/Migration012.java
```diff
@@ -30,6 +30,6 @@ private Migration012() {
    *  delegates to Migration009 as that does everything already.
```
SesquipedalianDefenestrator:
```suggestion
   *  delegates to FixRepairSegmentTimestamps as that does everything already.
```

SesquipedalianDefenestrator (changes requested):
A few nitpicky comment updates to match the code updates

SesquipedalianDefenestrator:
For the failing test, I suspect the postgres UPDATE statements need = CAST(NULL as timestamp) or possibly = CAST(NULL as TIMESTAMP WITH TIME ZONE)

SesquipedalianDefenestrator:
The Cassandra fix logic makes sense to me, but I'm having a bit of trouble with both instances of `V12_0_0__fix_repair_run_timestamps.sql`.  I don't see where it's setting `end_date` to anything but NULL. Am I missing something?

michaelsembwever:
> The Cassandra fix logic makes sense to me, but I'm having a bit of trouble with both instances of `V12_0_0__fix_repair_run_timestamps.sql`. I don't see where it's setting `end_date` to anything but NULL. Am I missing something?

I'm not sure I follow the question.
Both sql scripts only ever set `end_date` to `NULL`.

This is to meet the precondition here:  https://github.com/thelastpickle/cassandra-reaper/pull/651/files#diff-206e460c603d749f438e724e1023f236R303

src/server/src/main/resources/db/h2/V12_0_0__fix_repair_run_timestamps.sql
```diff
@@ -0,0 +1,52 @@
...
+--  See the License for the specific language governing permissions and
+--  limitations under the License.
+--
+-- Fix timestamp inconsistencies in the repair_run table
+--
+
+UPDATE repair_run
+SET start_time = NULL
+WHERE start_time is NOT NULL
+AND state = 0;
```
SesquipedalianDefenestrator:
```suggestion
AND state = 'NOT_STARTED';
```

src/server/src/main/resources/db/h2/V12_0_0__fix_repair_run_timestamps.sql
```diff
@@ -0,0 +1,52 @@
...
+
+UPDATE repair_run
+SET end_time = NULL
+WHERE end_time IS NOT NULL
+AND state != 2;
```
SesquipedalianDefenestrator:
```suggestion
AND state != 'ERROR';
```

src/server/src/main/resources/db/h2/V12_0_0__fix_repair_run_timestamps.sql
```diff
@@ -0,0 +1,52 @@
...
+
+UPDATE repair_run
+SET end_time = NULL
+WHERE end_time IS NOT NULL
+AND state != 3;
```
SesquipedalianDefenestrator:
```suggestion
AND state != 'DONE';
```

src/server/src/main/resources/db/h2/V12_0_0__fix_repair_run_timestamps.sql
```diff
@@ -0,0 +1,52 @@
...
+
+UPDATE repair_run
+SET end_time = NULL
+WHERE end_time IS NOT NULL
+AND state != 5;
```
SesquipedalianDefenestrator:
```suggestion
AND state != 'ABORTED';
```

src/server/src/main/resources/db/h2/V12_0_0__fix_repair_run_timestamps.sql
```diff
@@ -0,0 +1,52 @@
...
+
+UPDATE repair_run
+SET pause_time = NULL
+WHERE pause_time IS NOT NULL
+AND state != 4;
```
SesquipedalianDefenestrator:
```suggestion
AND state != 'PAUSED';
```

src/server/src/main/resources/db/h2/V12_0_0__fix_repair_run_timestamps.sql
```diff
@@ -0,0 +1,52 @@
...
+
+UPDATE repair_run
+SET start_time = end_time
+WHERE start_time is NULL
+AND end_time IS NOT NULL;
+
+UPDATE repair_run
+SET pause_time = start_time
+WHERE pause_time IS NULL
+AND state = 4;
```
SesquipedalianDefenestrator:
```suggestion
AND state = 'PAUSED';
```

src/server/src/main/resources/db/postgres/V12_0_0__fix_repair_run_timestamps.sql
```diff
@@ -0,0 +1,52 @@
...
+--  See the License for the specific language governing permissions and
+--  limitations under the License.
+--
+-- Fix timestamp inconsistencies in the repair_run table
+--
+
+UPDATE repair_run
+SET start_time = CAST(NULL as TIMESTAMP WITH TIME ZONE)
+WHERE start_time is NOT NULL
+AND state = 0;
```
SesquipedalianDefenestrator:
```suggestion
AND state = 'NOT_STARTED';
```

src/server/src/main/resources/db/postgres/V12_0_0__fix_repair_run_timestamps.sql
```diff
@@ -0,0 +1,52 @@
...
+
+UPDATE repair_run
+SET end_time = CAST(NULL as TIMESTAMP WITH TIME ZONE)
+WHERE end_time IS NOT NULL
+AND state != 2;
```
SesquipedalianDefenestrator:
```suggestion
AND state != 'ERROR';
```

src/server/src/main/resources/db/postgres/V12_0_0__fix_repair_run_timestamps.sql
```diff
@@ -0,0 +1,52 @@
...
+
+UPDATE repair_run
+SET end_time = CAST(NULL as TIMESTAMP WITH TIME ZONE)
+WHERE end_time IS NOT NULL
+AND state != 3;
```
SesquipedalianDefenestrator:
```suggestion
AND state != 'DONE';
```

src/server/src/main/resources/db/postgres/V12_0_0__fix_repair_run_timestamps.sql
```diff
@@ -0,0 +1,52 @@
...
+
+UPDATE repair_run
+SET end_time = CAST(NULL as TIMESTAMP WITH TIME ZONE)
+WHERE end_time IS NOT NULL
+AND state != 5;
```
SesquipedalianDefenestrator:
```suggestion
AND state != 'ABORTED';
```

src/server/src/main/resources/db/postgres/V12_0_0__fix_repair_run_timestamps.sql
```diff
@@ -0,0 +1,52 @@
...
+
+UPDATE repair_run
+SET pause_time = CAST(NULL as TIMESTAMP WITH TIME ZONE)
+WHERE pause_time IS NOT NULL
+AND state != 4;
```
SesquipedalianDefenestrator:
```suggestion
AND state != 'PAUSED';
```

src/server/src/main/resources/db/postgres/V12_0_0__fix_repair_run_timestamps.sql
```diff
@@ -0,0 +1,52 @@
...
+
+UPDATE repair_run
+SET start_time = end_time
+WHERE start_time is NULL
+AND end_time IS NOT NULL;
+
+UPDATE repair_run
+SET pause_time = start_time
+WHERE pause_time IS NULL
+AND state = 4;
```
SesquipedalianDefenestrator:
```suggestion
AND state = 'ABORTED';
```

SesquipedalianDefenestrator (changes requested):
Sorry, I was wrong on the cause of the type mismatch on the SQL migrations (or there's multiple).
H2 and Postgres both use a string representation for State.

SesquipedalianDefenestrator:
What I mean is, the original error that prompted this change is that RepairRun.endTime being set to null causes an error and a failure to purge old entries.  `FixRepairRunTimestamps.java` now sets endTime = startTime for ABORTED runs, which prevents that problem for the Cassandra back-end.

For H2 and Postgres backend migrations, though, I can't find an equivalent that's setting endTime = startTime for ABORTED runs.  Don't we need to do something like:
```
UPDATE repair_run
SET end_time = start_time
WHERE end_time is NULL
AND status = 'ABORTED';
```
(isTerminated() looks to be true for state 'ABORTED', so end_time is allowed and required for that state from what I can tell, but won't be set on older aborted runs)

michaelsembwever:
> What I mean is, the original error that prompted this change is that RepairRun.endTime being set to null causes an error and a failure to purge old entries. FixRepairRunTimestamps.java now sets endTime = startTime for ABORTED runs, which prevents that problem for the Cassandra back-end.

my boolean logic was whack. have replaced the "terminated" `SELECT` statements with a `WHERE state NOT IN (..)`

> For H2 and Postgres backend migrations, though, I can't find an equivalent that's setting endTime = startTime for ABORTED runs. Don't we need to do something like:

Indeed! Completely went by me, thanks for picking up on it. Fixed (see both `V12..cql` scripts).

SesquipedalianDefenestrator (approved):
Revamped logic makes total sense to me now.
Failed tests don't look as concerning, anyways.
One of them was a "git clone" failure, so I'll chalk that up to test infra
One of the test failures I don't understand, but smells a bit like a flaky test rather than a true failure.
The flapping reaper test looks like maybe a credential mismatch?

You're probably more familiar with the project's test infra state than I am, but the failures all look likely to be unrelated to these changes from what I can tell.

michaelsembwever:
> One of the test failures I don't understand, but smells a bit like a flaky test rather than a true failure.

Yes it looks that way!

I am taking your last comment as a reviewer `+1` , hope this is correct?
Once I get the tests passing I will merge.

SesquipedalianDefenestrator:
Yep, +1.  I thought I'd had "Approve" selected, but either I didn't or it didn't count it as a real approval because I don't have write access.

michaelsembwever:
Thanks. Yes you did approve the changes :-)

Issue #664, opened 2019-05-15
[1.4.2] Error 500 on /cluster/<cluster_name> and /repair_run


nsteinmetz:
I did the upgrade from 1.4.1 to 1.4.2 with a "pending repair", I mean a repair I defined but which was not yet activated. All others repairs are terminated.

I have in the logs:

```
ERROR  [ReaperApplication-scheduler] i.c.ReaperApplication - Couldn't resume running repair runs 
io.cassandrareaper.ReaperException: java.lang.IllegalStateException: endTime only valid when runState is terminated. PAUSED 2019-05-14T09:33:14.182Z
	at io.cassandrareaper.service.RepairManager.resumeRunningRepairRuns(RepairManager.java:112)
	at io.cassandrareaper.ReaperApplication.lambda$scheduleRepairManager$0(ReaperApplication.java:293)
	at com.codahale.metrics.InstrumentedScheduledExecutorService$InstrumentedRunnable.run(InstrumentedScheduledExecutorService.java:241)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:308)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:180)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:294)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
Caused by: java.lang.IllegalStateException: endTime only valid when runState is terminated. PAUSED 2019-05-14T09:33:14.182Z
	at com.google.common.base.Preconditions.checkState(Preconditions.java:721)
	at io.cassandrareaper.core.RepairRun$Builder.build(RepairRun.java:303)
	at io.cassandrareaper.storage.CassandraStorage.buildRepairRunFromRow(CassandraStorage.java:1192)
	at io.cassandrareaper.storage.CassandraStorage.getRepairRunsWithStateForCluster(CassandraStorage.java:699)
	at io.cassandrareaper.storage.CassandraStorage.getRepairRunsWithState(CassandraStorage.java:679)
[...]

PR #665, opened 2019-05-16
Automatically run every Cassandra Migration class when upgrading

michaelsembwever:
ref: https://github.com/thelastpickle/cassandra-reaper/issues/664


ossarga (commented):
I checked out and built Reaper 1.4.0, started a repair and paused halfway through. Then checked out your branch and build that. I was able to resume the repair and it completed successfully.

ossarga (approved):

PR #669, opened 2019-05-21
Make the segment locking LWTs happen inside an initial opportunistic check against `canRepair(..)`

michaelsembwever:
Make the segment locking LWTs happen inside an initial opportunistic check against `canRepair(..)`

The `canRepair(..) -> lockSegmentRunners() -> canRepair(..)` flow mimics the double-locking-idoim.
 Suspicion is the LWTs is not required. (If removed the second check on `canRepair(..)` must also be removed.)

ref: https://github.com/thelastpickle/cassandra-reaper/pull/622#discussion_r284190582

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
+++ b/src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
@@ -87,7 +87,9 @@ final class SegmentRunner implements RepairStatusHandler, Runnable {
   private static final Pattern REPAIR_UUID_PATTERN
       = Pattern.compile("[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}");
 
-  private static final long SLEEP_TIME_AFTER_POSTPONE_IN_MS = 10000;
+  private static final long SLEEP_TIME_AFTER_POSTPONE_IN_MS
+      = Integer.getInteger(SegmentRunner.class.getName() + ".sleep_time_after_postpone_in_ms", 10000);
+
   private static final ExecutorService METRICS_GRABBER_EXECUTOR = Executors.newFixedThreadPool(10);
   private static final long METRICS_POLL_INTERVAL_MS = TimeUnit.SECONDS.toMillis(5);
   private static final long METRICS_MAX_WAIT_MS = TimeUnit.MINUTES.toMillis(2);
@@ -304,27 +306,13 @@ final class SegmentRunner implements RepairStatusHandler, Runnable {
 
       LazyInitializer<Set<String>> busyHosts = new BusyHostsInitializer(cluster);
 
-      // If we're using a distributed storage, we need to synchronize with other Reaper instances
-      // So we don't start too many segments at the same time
-      if (!lockSegmentRunners()) {
-        LOG.debug(
-            "Not allowed to run the segment for now as another Reaper holds the lock for repair run {}. "
-            + "Will try again later",
-            segment.getRunId());
-        return false;
-      }
       if (!canRepair(segment, keyspace, coordinator, cluster, busyHosts)) {
         LOG.info(
-            "Cannot run segment {} for repair {} at the moment. Will try again later",
-            segmentId,
-            segment.getRunId());
-        SEGMENT_RUNNERS.remove(segment.getId());
+            "Cannot run segment {} for repair {} at the moment. Will try again later", segmentId, segment.getRunId());
+
         try {
           Thread.sleep(SLEEP_TIME_AFTER_POSTPONE_IN_MS);
-        } catch (InterruptedException e) {
-          LOG.debug("Interrupted while sleeping after a segment was postponed... weird stuff...");
-        }
-        releaseSegmentRunners();
+        } catch (InterruptedException ignore) { }
         return false;
       }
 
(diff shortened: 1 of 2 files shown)


src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
```diff
@@ -302,34 +304,45 @@ private boolean runRepair() {
...
-        releaseSegmentRunners();
+        } catch (InterruptedException ignore) { }
         return false;
       }
 
       try (Timer.Context cxt1 = context.metricRegistry.timer(metricNameForRepairing(segment)).time()) {
         try {
           LOG.debug("Enter synchronized section with segment ID {}", segmentId);
           synchronized (condition) {
+            if (!lockSegmentRunners()) {
```
adejanovski, line 323: I don't see the corresponding `releaseSegmentRunners()` call in this commit. What did you have in mind about this?

adejanovski (commented):
I'm not convinced about the approach because `canRepair()` will generate some delay in starting segment. In sidecar and each mode, it takes a little while to start segments because of the metrics requests so doing this twice will double this time.
It would be good to run some repairs with both versions to compare total durations.

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
michaelsembwever: the code is inside the `try-finally` block. see https://github.com/thelastpickle/cassandra-reaper/pull/669/commits/5a4ede8c187714c8cfe1dcfb4522fe37d17e5d3e#diff-e6aa6503be815f976c3ed7a845621f4aR387

michaelsembwever:
> In sidecar and each mode, it takes a little while to start segments because of the metrics requests so doing this twice will double this time.

Are the metrics requests cleared and re-requested?

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
adejanovski: Cool, I missed it.

adejanovski:
> > In sidecar and each mode, it takes a little while to start segments because of the metrics requests so doing this twice will double this time.
> 
> Are the metrics requests cleared and re-requested?

Yes, otherwise it would defeat the double check idiom : https://github.com/thelastpickle/cassandra-reaper/blob/alex/new-sidecar-mode/src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java#L628

michaelsembwever:
> Yes, otherwise it would defeat the double check idiom : https://github.com/thelastpickle/cassandra-reaper/blob/alex/new-sidecar-mode/src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java#L628

`canRepair(..)` involves two checks: one against the storage and one with jmx metrics (direct or requested).

Either way the trade-off, the first check of `canRepair(..)` is expected to be faster than the LWT locking.
Furthermore, it is unknown how often we'll depend on the second call to `canRepair(..)`, and that questions the logic to it…
Maybe the second call to `canRepair(..)` should be replaced with only a check against the storage backend. Since the jmx metrics arn't expected to change that much over the "double-locking"?

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
```diff
@@ -302,34 +304,45 @@ private boolean runRepair() {
...
+                  segmentId,
+                  segment.getRunId());
+
+              return false;
+            }
+
+            // double-locking-idiom, only applies to non-incremental and distributed storage
+            if (!repairUnit.getIncrementalRepair()
+                && context.storage instanceof IDistributedStorage
+                && !canRepair(segment, keyspace, coordinator, cluster, busyHosts)) {
```
michaelsembwever: does the jmx metrics have to happen inside the lock? if not we can simplify the second `canRepair(..)` (this one).

michaelsembwever: possibility to do `assert nodesReadyForNewRepair (..)` here…

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
```diff
@@ -1146,6 +1180,58 @@ long intensityBasedDelayMillis(double intensity) {
...
+  private boolean lockSegmentRunners() {
+    if (this.repairUnit.getIncrementalRepair()) {
+      return true;
+    }
+
+    try (Timer.Context cx
+        = context.metricRegistry.timer(MetricRegistry.name(SegmentRunner.class, "lockSegmentRunners")).time()) {
+
+      boolean result = context.storage instanceof IDistributedStorage
+          ? ((IDistributedStorage) context.storage).takeLead(repairRunner.getRepairRunId(), LOCK_DURATION)
```
adejanovski: Instead of using `repairRunner.getRepairRunId()`, let's use a unique constant UUID that would be declared as `final static` member of SegmentRunner (like `13814000-1dd2-11b2-8080-808080808080` for example, any valid timeuuid would work here I guess).

This will allow to have a global lock which is what we want, not a repair level lock.

Thi

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
```diff
@@ -1146,6 +1180,58 @@ long intensityBasedDelayMillis(double intensity) {
...
+  private boolean renewLockSegmentRunners() {
+    if (this.repairUnit.getIncrementalRepair()) {
+      return true;
+    }
+
+    try (Timer.Context cx
+        = context.metricRegistry.timer(MetricRegistry.name(SegmentRunner.class, "renewLockSegmentRunners")).time()) {
+
+      boolean result = context.storage instanceof IDistributedStorage
+          ? ((IDistributedStorage) context.storage).renewLead(repairRunner.getRepairRunId(), LOCK_DURATION)
```
adejanovski: Same here, we need to use the global lock, not a per repair lock.

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
```diff
@@ -1146,6 +1180,58 @@ long intensityBasedDelayMillis(double intensity) {
...
+      return result;
+    }
+  }
+
+  private void releaseSegmentRunners() {
+    if (!this.repairUnit.getIncrementalRepair()) {
+      try (Timer.Context cx
+          = context.metricRegistry.timer(MetricRegistry.name(SegmentRunner.class, "releaseSegmentRunners")).time()) {
+        if (context.storage instanceof IDistributedStorage) {
+          ((IDistributedStorage) context.storage).releaseLead(repairRunner.getRepairRunId());
```
adejanovski: To be consistent with the other methods, this needs to use the global lock too.

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
michaelsembwever: I'm left a bit uncertain on this.

The work in this PR is to add validation onto whether we need the locking at all. I suspect that we can't 100% avoid the situation of duplicate segment repairs no matter what we do, and that the opportunistic provided here will do the job in all production environments.

Can we focus this PR on just adding the opportunistic check and leave any corrections to the LWT separate?

michaelsembwever:
> Maybe the second call to canRepair(..) should be replaced with only a check against the storage backend.

this has been done.

adejanovski (approved):
I agree on keeping the changes to the LWT as separate from this PR.

PR #675, opened 2019-05-29
Improve performance of JsonParseUtils

michaelsembwever:
- was taking up 92% of calls to `storage.getNextFreeSegmentInRange(..)`
 - which is 81%of `RepairRunner.run()`
 - was taking up 99% of cpu time to `ClusterResource.getCluster(..)`


adejanovski (approved):
I always wondered what was the cost of instantiating the ObjectMapper on demand. I guess I have the answer now 😅 

LGTM

PR #676, opened 2019-05-29
Add a one minute cache to `ClusterFacade.getTablesForKeyspace(..)`

michaelsembwever:
- If the cluster is remote then a lot of time to  `RepairRunResource.listRepairRuns(..)` is taken by the jmx calls behind `ClusterFacade.getTablesForKeyspace(..)`

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/jmx/ClusterFacade.java
+++ b/src/server/src/main/java/io/cassandrareaper/jmx/ClusterFacade.java
@@ -44,6 +44,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


adejanovski (approved):
I'll approve the PR to get performance enhancements but I think we should review the code to persist the list of tables in the repair run instance instead of computing it when listing runs.
wdyt?

michaelsembwever:
> I'll approve the PR to get performance enhancements but I think we should review the code to persist the list of tables in the repair run instance instead of computing it when listing runs.

Completely agree, caches typically point towards a fault. It makes sense to persist in the repair run. Though it must not be persisted in the schedule.
If a table is altered to/from TWCS while a repair run is running, the UI will be inaccurate either way.

PR #677, opened 2019-05-29
WIP : upgrade the admin HTML theme from 1.0.5 to 4.0.4 

adejanovski:
This will allow us to upgrade to Bootstrap 4 and have a more modern look for the UI.

### DIFF
--- a/src/ui/app/html_template.ejs
+++ b/src/ui/app/html_template.ejs
@@ -42,7 +42,7 @@
 
 </head>
 
-<body>
+<body id="page-top">
 
     <div id="wrapper">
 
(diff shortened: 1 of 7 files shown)


adejanovski:
@ossarga, is this migration something you think you could handle? (since you already did Webpack 4 😉)
I've partially migrated the cluster screen which "mostly works", except the modals don't show up anymore.
It was a quick and dirty attempt to see if the migration was possible or not and it seems like we could achieve it without having to change too many things.

adejanovski:
Upgraded the theme to v3 and refreshed the UI in another PR. Will be enough for now.

PR #680, opened 2019-05-31
Add 'tables' column to repair_run table

michaelsembwever:
This allows the automatic blacklisting of TWCS to be calculated just once when the repair run is created.
Previously the TWCS blacklisting was done when reading/displaying repair runs,
  and if jmx latency was high it took minutes to display UI pages listing repair runs.
This improvement also ensures that all repair segment in one repair run repair the same tables.
 Before it was possible for repair segments to differ if table compaction strategies were changed during a repair run.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
+++ b/src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
@@ -95,8 +95,6 @@ import org.cognitor.cassandra.migration.Database;
 import org.cognitor.cassandra.migration.MigrationRepository;
 import org.cognitor.cassandra.migration.MigrationTask;
 import org.joda.time.DateTime;
-import org.joda.time.format.DateTimeFormat;
-import org.joda.time.format.DateTimeFormatter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import systems.composable.dropwizard.cassandra.CassandraFactory;
@@ -111,11 +109,10 @@ public final class CassandraStorage implements IStorage, IDistributedStorage {
   private static final String SELECT_REPAIR_UNIT = "SELECT * FROM repair_unit_v1";
   private static final String SELECT_LEADERS = "SELECT * FROM leader";
   private static final String SELECT_RUNNING_REAPERS = "SELECT reaper_instance_id FROM running_reapers";
-  private static final DateTimeFormatter HOURLY_FORMATTER = DateTimeFormat.forPattern("yyyyMMddHH");
 
   private static final Logger LOG = LoggerFactory.getLogger(CassandraStorage.class);
 
-  private static AtomicBoolean firstInstantiation = new AtomicBoolean(true);
+  private static final AtomicBoolean UNINITIALISED = new AtomicBoolean(true);
 
   private final com.datastax.driver.core.Cluster cassandra;
   private final Session session;
@@ -185,7 +182,7 @@ public final class CassandraStorage implements IStorage, IDistributedStorage {
 
     // https://docs.datastax.com/en/developer/java-driver/3.5/manual/metrics/#metrics-4-compatibility
     cassandraFactory.setJmxEnabled(false);
-    if (!CassandraStorage.firstInstantiation.compareAndSet(true, false)) {
+    if (!CassandraStorage.UNINITIALISED.compareAndSet(true, false)) {
       // If there's been a past connection attempt, metrics are already registered
       cassandraFactory.setMetricsEnabled(false);
     }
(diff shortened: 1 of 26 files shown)


src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
```diff
@@ -111,11 +110,10 @@
   private static final String SELECT_REPAIR_UNIT = "SELECT * FROM repair_unit_v1";
   private static final String SELECT_LEADERS = "SELECT * FROM leader";
   private static final String SELECT_RUNNING_REAPERS = "SELECT reaper_instance_id FROM running_reapers";
-  private static final DateTimeFormatter HOURLY_FORMATTER = DateTimeFormat.forPattern("yyyyMMddHH");
 
   private static final Logger LOG = LoggerFactory.getLogger(CassandraStorage.class);
 
-  private static AtomicBoolean firstInstantiation = new AtomicBoolean(true);
+  private static final AtomicBoolean UNINITIALISED = new AtomicBoolean(true);
```
michaelsembwever, line 115: not strictly part of the PR.

adejanovski: you bad boy 😱

src/server/src/main/resources/db/cassandra/020_repair_run_tables.cql
```diff
@@ -0,0 +1,18 @@
...
+--
+--  Unless required by applicable law or agreed to in writing, software
+--  distributed under the License is distributed on an "AS IS" BASIS,
+--  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+--  See the License for the specific language governing permissions and
+--  limitations under the License.
+--
+-- Upgrade to allow segments to contain several token ranges
+
+ALTER TABLE repair_run ADD tables text STATIC;
```
adejanovski: `repair_unit_v1` uses `set<text>` types for table list storage. Any specific reason made you choose a plain text field instead?

adejanovski (commented):
LGTM aside from a minor comment on the types used to store the list of tables in `repair_run`.

adejanovski:
Apparently the list of tables is persisted even when the TWCS auto blacklist setting is set to false or if there's no blacklist.
Since I guess most of the times we won't have blacklists, would it be possible to leave the table list empty (meaning all tables) then? That would lighten the UI for keyspaces with many tables.

adejanovski:
Also, integration tests are broken it seems because of the way table names are stored : 
```
ERROR [Thread-5] 2019-05-31 11:37:16,184 StorageService.java (line 2647) Repair session failed:
java.lang.IllegalArgumentException: Unknown keyspace/cf pair (booya.((((((((booya2)
	at org.apache.cassandra.db.Keyspace.getColumnFamilyStore(Keyspace.java:167)
	at org.apache.cassandra.service.StorageService.getValidColumnFamilies(StorageService.java:2381)
	at org.apache.cassandra.service.StorageService.forceKeyspaceRepair(StorageService.java:2701)
	at org.apache.cassandra.service.StorageService$4.runMayThrow(StorageService.java:2643)
	at org.apache.cassandra.utils.WrappedRunnable.run(WrappedRunnable.java:28)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.lang.Thread.run(Thread.java:748)
```

Looking at the traces, it seems you have parenthesis in some table names : 
```
[{"cause":"scheduled run (schedule id 00000000-0000-0005-0000-000000000000)","owner":"test_user","id":"00000000-0000-0001-0000-000000000000","state":"RUNNING","intensity":0.8999999761581421,"duration":"5 minutes 2 seconds","cluster_name":"test",
"column_families":["((((((((((((booya2","            booya1","            booya_twcs))))))))))))"],
"keyspace_name":"booya","incremental_repair":false,"total_segments":36,"repair_parallelism":"SEQUENTIAL","segments_repaired":0,"last_event":"Triggered repair of segment 00000000-0000-001d-0000-000000000000 via host 127.0.0.2","nodes":[],"datacenters":[],"blacklisted_tables":[],"repair_thread_count":1,"creation_time":"2019-05-31T11:33:45Z","start_time":"2019-05-31T11:33:46Z","end_time":null,"pause_time":null,"current_time":"2019-05-31T11:38:48Z","estimated_time_of_arrival":null}]
```

src/server/src/main/resources/db/cassandra/020_repair_run_tables.cql
michaelsembwever: good point! i will change the code to use the `set<text>` approach.

michaelsembwever:
> Apparently the list of tables is persisted even when the TWCS auto blacklist setting is set to false or if there's no blacklist.
> Since I guess most of the times we won't have blacklists, would it be possible to leave the table list empty (meaning all tables) then? That would lighten the UI for keyspaces with many tables.

I don't know what to do here. If we allow the table list to be blank, when it's to indicate 'all tables', it opens us up to faults when new tables are added, blacklisting setting is changed, or compaction strategies are altered.
For created and started repair runs there is a definite benefit in all situations of putting the table list in explicitly. It makes the table list immutable during the life of the repair run and for all its repair segments.

The trade-off though is the burden it puts on the UI.

michaelsembwever:
> Also, integration tests are broken it seems because of the way table names are stored :

I still need to work with the integration tests, a few scenarios have now changed behaviour…

EDIT: done.

adejanovski (approved):
LGTM.

I've pushed a UI change to better display the list of tables and unclutter the repair screen:  
reaper_table_tooltips2

michaelsembwever:
> I've pushed a UI change to better display the list of tables and unclutter the repair screen:

Brilliant. Thanks so much!!

PR #692, opened 2019-06-03
Fix initialisation ordering of AppContext

### LINKED ISSUE #654: Exception in AppContext init throws NPE
When `AppContext` initialises, it has some exception handling if it's unable to determine a host address.
https://github.com/thelastpickle/cassandra-reaper/blob/d7d045f7eaa4e214584c4ec8082265ba45c36d9b/src/server/src/main/java/io/cassandrareaper/AppContext.java#L53-L62
I'm getting a NullPointerException on Line 58, which I think is because, above...
https://github.com/thelastpickle/cassandra-reaper/blob/d7d045f7eaa4e214584c4ec8082265ba45c36d9b/src/server/src/main/java/io/cassandrareaper/AppContext.java#L40-L43
the logger isn't initialised till after this function runs. 

I may have misunderstood what's going on here, but this seems like a simple oversight in the ordering. Happy to submit a PR to fix if you agree on the cause, but not sure how best to set up a test case.

treuherz:
Previously, `initialiseInstanceAddress()` was called before `LOG` and `DEFAULT_INSTANCE_ADDRESS` were initialised, but relied on both of them, so threw an NPE instead of correctly assigning the default and starting up.

Fixes #654

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/AppContext.java
+++ b/src/server/src/main/java/io/cassandrareaper/AppContext.java
@@ -35,12 +35,14 @@ import org.slf4j.LoggerFactory;
  * Single class to hold all application global interfacing objects, and app global options.
  */
 public final class AppContext {
+  
+  private static final Logger LOG = LoggerFactory.getLogger(AppContext.class);
+
+  private static final String DEFAULT_INSTANCE_ADDRESS = "127.0.0.1";
 
   public static final UUID REAPER_INSTANCE_ID = UUID.randomUUID();
   public static final String REAPER_INSTANCE_ADDRESS = initialiseInstanceAddress();
 
-  private static final String DEFAULT_INSTANCE_ADDRESS = "127.0.0.1";
-  private static final Logger LOG = LoggerFactory.getLogger(AppContext.class);
 
   public final AtomicBoolean isRunning = new AtomicBoolean(true);
   public IStorage storage;


michaelsembwever:
Thanks for the contribution @treuherz !

The patch as is won't build because it breaks our code style, specifically the 'Declaration Order' rule:
```
[ERROR] /home/travis/build/thelastpickle/cassandra-reaper/src/server/src/main/java/io/cassandrareaper/AppContext.java:38: Line matches the illegal pattern 'Trailing whitespace'. [Regexp]
[ERROR] /home/travis/build/thelastpickle/cassandra-reaper/src/server/src/main/java/io/cassandrareaper/AppContext.java:43:3: Variable access definition in wrong order. [DeclarationOrder]
[ERROR] /home/travis/build/thelastpickle/cassandra-reaper/src/server/src/main/java/io/cassandrareaper/AppContext.java:44:3: Variable access definition in wrong order. [DeclarationOrder]
```

I will put a fix based on your work in.

michaelsembwever:
merged with https://github.com/thelastpickle/cassandra-reaper/commit/534f5aae0dde877910e8394653c8b81307441428

treuherz:
Appreciate the quick response. My IDE was misconfigured and the checks didn't run when I thought they had, apologies for that. It looks like it might have been that same DeclarationOrder inspection that led to this bug in the first place, though.

PR #715, opened 2019-07-11
Request remote nodes metrics only if there are several reapers running

adejanovski:
Skips `getRemoteNodeMetrics()` call if there's only one reaper instance running.
Will make LOCAL more versatile and help with integration tests success rate.


michaelsembwever (approved):
LGTM

PR #716, opened 2019-07-15
Don't include *-sources.jar into the classpath, and be a bit smarter …

michaelsembwever:
and be a bit smarter in finding the jar file in a current project directory (regardless of the current folder in the project directory)

### DIFF
--- a/src/packaging/bin/cassandra-reaper
+++ b/src/packaging/bin/cassandra-reaper
@@ -1,5 +1,6 @@
 #!/bin/bash
 # Copyright (c) 2014-2015 Spotify AB
+# Copyright 2019-2019 The Last Pickle Ltd
 #
 #  Licensed under the Apache License, Version 2.0 (the "License");
 #  you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


src/packaging/bin/cassandra-reaper
```diff
@@ -22,7 +23,7 @@ fi
 
 if [ -z "$CLASS_PATH" ]; then
   echo "Looking for reaper in /usr/share/cassandra-reaper/"
-  CLASS_PATH="/usr/share/cassandra-reaper/*:/usr/local/share/cassandra-reaper/*"
+  CLASS_PATH="$(find /usr/local/share -path 'cassandra-reaper/cassandra-reaper-*[0-9rT].jar'):$(find /usr/share -path 'cassandra-reaper/cassandra-reaper-*[0-9rT].jar')"
```
michaelsembwever, line 26: If operators were intentionally putting other jar files into the `.../share/cassandra-reaper/` folder, eg log or metric spi implementations, then this will break that as it now only loads the cassandra-reaper shaded jar file.

The correct approach for operators in such a situation is to manually declare `$CLASSPATH`.

adejanovski (approved):

PR #719, opened 2019-07-17
WIP – avoid cluster conflicts on ipaddresses

michaelsembwever:
**Add `State` and `Last Contact` fields to the Cluster object**  
  - Introduces cluster states: UNKNOWN, ACTIVE, UNREACHABLE, DELETED
 - The UNKNOWN state is needed as a cluster object is often created on-the-fly to pass through jmx port and credentials as connection information
 - Clusters can only be marked as DELETED if they have no active schedules and no running repairs
 - Clusters are marked ACTIVE (and last_contact) updated on successful jmx requests are made against them
 - Clusters are sorted by state then name
 - DELETED clusters are not listed

**Introduce `Cluster.Builder`**  
  - IStorage no longer returns Optionals against Cluster methods, if the name argument is wrong it throws an IllegalArgumentException instead,
 - Node.withClusterName(..) method and usage removed, as it was never used anywhere,
 - Rationalise ClusterFacade connection methods, simplify connection method names.
 - Remove JmxConnectionFactory.connect(node) method, only tests were using it.

WAS
> Add "&force=true` query parameter to the delete cluster rest endpoint that forcible deletes a cluster (by first deleting all schedules and repairs belonging to it).
> 
> Also add the functionality to detect conflicting clusters (two clusters using seeds of the same ipaddress) and to delete the cluster(s) that no longer exists.
> 
> The significant patch is the first commit.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
+++ b/src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
@@ -20,6 +20,8 @@ package io.cassandrareaper.resources;
 import io.cassandrareaper.AppContext;
 import io.cassandrareaper.ReaperException;
 import io.cassandrareaper.core.Cluster;
+import io.cassandrareaper.core.RepairRun;
+import io.cassandrareaper.core.RepairSchedule;
 import io.cassandrareaper.jmx.ClusterFacade;
 import io.cassandrareaper.jmx.JmxProxy;
 import io.cassandrareaper.resources.view.ClusterStatus;
@@ -27,6 +29,7 @@ import io.cassandrareaper.resources.view.NodesStatus;
 import io.cassandrareaper.service.ClusterRepairScheduler;
 
 import java.net.URI;
+import java.time.LocalDate;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -86,27 +89,26 @@ public final class ClusterResource {
   }
 
   @GET
-  public Response getClusterList(@QueryParam("seedHost") Optional<String> seedHost) throws ReaperException {
+  public Response getClusterList(@QueryParam("seedHost") Optional<String> seedHost) {
     LOG.debug("get cluster list called");
-    Collection<Cluster> clusters = context.storage.getClusters();
-    List<String> clusterNames = new ArrayList<>();
-    for (Cluster cluster : clusters) {
-      if (seedHost.isPresent()) {
-        if (cluster.getSeedHosts().contains(seedHost.get())) {
-          clusterNames.add(cluster.getName());
-        }
-      } else {
-        clusterNames.add(cluster.getName());
-      }
-    }
-    return Response.ok().entity(clusterNames).build();
+
+    Collection<String> clusters = context.storage.getClusters()
+        .stream()
+        .filter(c -> seedHost.isPresent()
+            ? c.getSeedHosts().contains(seedHost.get())
+            : Cluster.State.DELETED != c.getState())
+        .sorted()
+        .map(c -> c.getName())
+        .collect(Collectors.toList());
+
+    return Response.ok().entity(clusters).build();
   }
 
   @GET
   @Path("/{cluster_name}")
   public Response getCluster(
       @PathParam("cluster_name") String clusterName,
-      @QueryParam("limit") Optional<Integer> limit) throws ReaperException {
+      @QueryParam("limit") Optional<Integer> limit) {
 
     LOG.debug("get cluster called with cluster_name: {}", clusterName);
     try {
@@ -120,19 +122,20 @@ public final class ClusterResource {
             context.jmxConnectionFactory.getJmxCredentialsForCluster(clusterName).get().getPassword());
       }
       Cluster cluster = context.storage.getCluster(clusterName);
+      if (Cluster.State.DELETED != cluster.getState()) {
 
-      ClusterStatus clusterStatus = new ClusterStatus(
-            cluster,
-            jmxUsername,
-            jmxPasswordIsSet,
-            context.storage.getClusterRunStatuses(cluster.getName(), limit.orElse(Integer.MAX_VALUE)),
-            context.storage.getClusterScheduleStatuses(cluster.getName()),
-            getNodesStatus(cluster));
+        ClusterStatus clusterStatus = new ClusterStatus(
+              cluster,
+              jmxUsername,
+              jmxPasswordIsSet,
+              context.storage.getClusterRunStatuses(cluster.getName(), limit.orElse(Integer.MAX_VALUE)),
+              context.storage.getClusterScheduleStatuses(cluster.getName()),
+              getNodesStatus(cluster));
 
-      return Response.ok().entity(clusterStatus).build();
-    } catch (IllegalArgumentException ex) {
-      return Response.status(404).entity("cluster with name \"" + clusterName + "\" not found").build();
-    }
+        return Response.ok().entity(clusterStatus).build();
+      }
+    } catch (IllegalArgumentException ignore) { }
+    return Response.status(404).entity("cluster with name \"" + clusterName + "\" not found").build();
   }
 
   @GET
@@ -265,13 +268,15 @@ public final class ClusterResource {
       if (context.config.getEnableDynamicSeedList() && !liveNodes.isEmpty()) {
         seedHosts = ImmutableSet.copyOf(liveNodes);
       }
-      LOG.debug("Seeds {}", seedHosts);
+      LOG.debug("Cluster {}", seedHosts);
 
       return Cluster.builder()
               .withName(clusterName)
               .withPartitioner(partitioner)
               .withSeedHosts(seedHosts)
               .withJmxPort(jmxPort.orElse(Cluster.DEFAULT_JMX_PORT))
+              .withState(Cluster.State.ACTIVE)
+              .withLastContact(LocalDate.now())
               .build();
     } catch (ReaperException e) {
       LOG.error("failed to find cluster with seed hosts: {}", seedHosts, e);
@@ -299,7 +304,12 @@ public final class ClusterResource {
           cluster.getName(), StringUtils.join(previousNodes, ','), StringUtils.join(liveNodes, ','));
 
       if (!cluster.getSeedHosts().equals(liveNodes)) {
-        cluster = cluster.with().withSeedHosts(liveNodes).build();
+        cluster = cluster.with()
+              .withSeedHosts(liveNodes)
+              .withState(Cluster.State.ACTIVE)
+              .withLastContact(LocalDate.now())
+              .build();
+
         context.storage.updateCluster(cluster);
       }
       return cluster;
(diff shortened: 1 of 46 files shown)


src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
```diff
@@ -347,6 +347,23 @@ public Response deleteCluster(@PathParam("cluster_name") String clusterName)
...
+            .forEach((uuid) -> context.storage.deleteRepairRun(uuid));
+
+        try {
+          Thread.sleep(100);
+        } catch (InterruptedException ignore) {
+          break;
+        }
+      } while (9 > attempts && (context.storage.getRepairSchedulesForCluster(clusterName).isEmpty()
+            || !context.storage.getRepairRunsForCluster(clusterName, Optional.empty()).isEmpty()));
+    }
```
adejanovski: That condition is hard to read and understand.
You're looping as long as you haven't reached 10 attempts and [there are no more schedules or there are still repair runs].
Shouldn't the condition be [there are still schedules or there are still repair runs] ?

src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
```diff
@@ -267,14 +272,12 @@ private Cluster findClusterWithSeedHost(String seedHost, Optional<Integer> jmxPo
...
-          = new Cluster(
+      Cluster cluster = new Cluster(
               parsedClusterName,
               Optional.empty(),
               Sets.newHashSet(seedHost),
-              ClusterProperties.builder()
-                  .withJmxPort(jmxPort.orElse(Cluster.DEFAULT_JMX_PORT))
-                  .build());
+              ClusterProperties.builder().withJmxPort(jmxPort.orElse(Cluster.DEFAULT_JMX_PORT)) .build());
+
```
adejanovski: Seems like you've been re-formatting code that hasn't been modified in this PR.
Makes the review more challenging ;)

src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
```diff
@@ -84,21 +85,26 @@ public ClusterResource(AppContext context, ExecutorService executor) {
   }
 
   @GET
-  public Response getClusterList(@QueryParam("seedHost") Optional<String> seedHost)
-      throws ReaperException {
-
+  public Response getClusterList(@QueryParam("seedHost") Optional<String> seedHost) throws ReaperException {
     LOG.debug("get cluster list called");
-    Collection<Cluster> clusters = context.storage.getClusters();
+    Collection<Cluster> clusters = removeConflictingClusters(context.storage.getClusters());
```
adejanovski: I'm worried by the additional cost we're adding on getting the cluster list, especially for large installs.
Having conflicting clusters is not something that should happen that much, unless you have a large infrastructure with IP addresses that get re-used a lot.
I'd be in favor of creating a new rest operation, or run the command in the background on a regular basis (once per day for example, or upon startup of Reaper), instead of doing this check every time we list the clusters. We list them A LOT in the UI.

src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
```diff
@@ -347,6 +347,23 @@ public Response deleteCluster(@PathParam("cluster_name") String clusterName)
...
+      int attempts = 0;
+      do {
+        context.storage.getRepairSchedulesForCluster(clusterName)
+            .forEach((schedule) -> context.storage.deleteRepairSchedule(schedule.getId()));
+
+        context.storage.getRepairRunIdsForCluster(clusterName)
+            .forEach((uuid) -> context.storage.deleteRepairRun(uuid));
+
+        try {
+          Thread.sleep(100);
```
adejanovski: Are we deleting at `LOCAL_ONE` in the context of Cassandra, and is that why you're adding a sleep here?

Couldn't we move to a more consistent CL in that case and avoid adding this block?

adejanovski (changes requested):
It will be good to have a way of deleting clusters which still have history, but we must complement it with changes in the UI (if removing the cluster fails, we can offer to force delete and ask for confirmation).
I think the conflict handling should happen on some occasions only, and not each time we list the clusters. That'll bite us back in the case of large Reaper installs.

src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
michaelsembwever: that's why there's two commits in the PR. review just the first commit ;)

michaelsembwever:
I'm too leaning away from the idea of hard deleting the cluster, schedules, and repair runs.

I've been thinking about how to add `Cluster.State` with values of `REGISTERED|DELETED|UNREACHABLE`

Such an approach would solve most of the questions raised above, i do believe…

michaelsembwever:
This is ready for a preliminary review.

Have updated the description according to the new approach.

Still on my todo list

- [ ]  move some of the trivial code cleaning changes from the second commit forward to the first commit,
- [ ]  handling changing clusters to UNREACHABLE state when last_contact is older than X days,
- [ ]  handling updating the cluster's last_contact value
- [ ]  figure out how to re-add (or re-use) deleted clusters,
- [ ]  handling prevention of activating schedules or running repairs on clusters that have been marked as deleted
- [ ]  unit tests

src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
```diff
@@ -82,17 +82,14 @@ private RepairRunner(
...
-
-    Preconditions.checkArgument(cluster.isPresent(), "Cluster couldn't be found in storage");
+    this.clusterName = cluster.getName();
     String keyspace = repairUnitOpt.getKeyspaceName();
-    int parallelRepairs
-        = getPossibleParallelRepairsCount(
-            clusterFacade.getRangeToEndpointMap(cluster.get(), keyspace),
-            clusterFacade.getEndpointToHostId(cluster.get()));
+
+    int parallelRepairs = getPossibleParallelRepairsCount(
```
adejanovski, line 90: hem... weren't you the one asking to put the operator on the next line so it carries?
Readability of the code is better if the method call starts on the next line, but I won't be dogmatic about this.

src/server/src/main/java/io/cassandrareaper/service/SnapshotService.java
```diff
@@ -106,16 +106,12 @@ public static SnapshotService create(AppContext context, ExecutorService executo
               .build();
 
       context.storage.saveSnapshot(snapshot);
+      LOG.info("Cluster : {} ; Cluster obj : {}", clusterName, cluster);
```
adejanovski, line 109: I guess we should move this to debug

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
```diff
@@ -500,99 +499,102 @@ public boolean isStorageConnected() {
...
-    for (Row row : results) {
-      if (row.getString("cluster_name").equals(clusterName)) {
-        UUID id = row.getUUID("id");
-        assert getRepairRunsForUnit(id).isEmpty() : StringUtils.join(getRepairRunsForUnit(id));
-        session.executeAsync(deleteRepairUnitPrepStmt.bind(id));
-      }
-    }
-    session.executeAsync(deleteClusterPrepStmt.bind(clusterName));
-    return Optional.ofNullable(new Cluster(clusterName, null, null));
+    assert getRepairRunsWithState(RepairRun.RunState.RUNNING).stream()
```
adejanovski, line 600: That check/call is done in the ClusterResource method already, twice in case there are matching repairs. Since listing repair runs by state can be quite expensive in Cassandra, we might want to limit the number of calls.

src/server/src/main/java/io/cassandrareaper/storage/PostgresStorage.java
```diff
@@ -139,26 +136,27 @@ public boolean isStorageConnected() {
...
-      int rowsAdded
-          = getPostgresStorage(h)
-              .insertCluster(
-                  newCluster.getName(),
-                  newCluster.getPartitioner().get(),
-                  newCluster.getSeedHosts(),
-                  properties);
+      Preconditions.checkState(newCluster.getPartitioner().isPresent(), "Cannot insert cluster with no partitioner.");
+
+      int rowsAdded = getPostgresStorage(h).insertCluster(
```
adejanovski, line 147: What is going to happen if we try to add back a cluster that was in `DELETED` state?

src/server/src/main/resources/db/cassandra/022_cluster_states.cql
```diff
@@ -0,0 +1,18 @@
...
+--
+--  Unless required by applicable law or agreed to in writing, software
+--  distributed under the License is distributed on an "AS IS" BASIS,
+--  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+--  See the License for the specific language governing permissions and
+--  limitations under the License.
+--
+-- Upgrade to store cluster state and last jmx connection time
+
+ALTER TABLE cluster ADD state text, last_contact date;
```
adejanovski: Why not use a timestamp for `last_contact`? It would match the other storage backend types.

src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
```diff
@@ -195,108 +185,102 @@ private Response addOrUpdateCluster(
...
 
-        LOG.info(msg);
-        return Response.status(Response.Status.BAD_REQUEST).entity(msg).build();
-      }
-      Optional<Cluster> existingCluster = context.storage.getCluster(cluster.getName());
-      URI location = uriInfo.getBaseUriBuilder().path("cluster").path(cluster.getName()).build();
-      if (existingCluster.isPresent()) {
-        LOG.debug("Attempting updating nodelist for cluster {}", existingCluster.get().getName());
+    Preconditions.checkState(
+        1 >= matchingClusters.get().count(),
```
adejanovski, line 211: I think this will prevent re-adding a cluster that's been put in `DELETED` state. There would be no way to add back a cluster that was mistakenly deleted.

adejanovski (commented):
Appreciate the cleanups overall.
I need to check if the changes to ClusterFacade connect* methods didn't break some specific cases.
I'll have to deploy this branch to run some tests.
I'm not sure what's the workflow to delete and then re-add a cluster looking at the code, but I'll have to run some tests there as well.

michaelsembwever:
>  I need to check if the changes to ClusterFacade connect* methods didn't break some specific cases.
> I'll have to deploy this branch to run some tests.
> I'm not sure what's the workflow to delete and then re-add a cluster looking at the code, but I'll have to run some tests there as well.

let's make sure those checks  also become tests.
that said, i'm still going those the integration tests with this patch.

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
michaelsembwever: It's intentional to be in both places. At the resource level it is required so to provide correct response handling to the user, ie application level. At the storage level it is provided for program correctness. If performance is a problem in production systems it is intended that assertions can be disabled.

src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
michaelsembwever: The rules for wrapping is to apply one (and only one) of the following approaches:
 - carry the operator,
 - carry the method calls, 
 - carry every argument.

In this example the arguments are getting carried.

src/server/src/main/java/io/cassandrareaper/storage/PostgresStorage.java
michaelsembwever: see todo list above: https://github.com/thelastpickle/cassandra-reaper/pull/719#issuecomment-517219637

>  figure out how to re-add (or re-use) deleted clusters,

src/server/src/main/resources/db/cassandra/022_cluster_states.cql
michaelsembwever: the field will only store the day. (is there a sql equivalent?)

i may need to change it to timestamp though for compatibility across c* versions…

src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
michaelsembwever: That's not what this precondition is checking, and it won't prevent re-adding an existing deleted cluster (so long as nodes overlap).

adejanovski:
After giving it some thought, I don't see a benefit in implementing the `DELETED` state for clusters. It adds complexity and I do not know what to do with clusters in such state. They won't get displayed anyway.
We should purely delete those from storage and probably cleanup all tables holding data for them (but the schedules maybe).

michaelsembwever:
> After giving it some thought, I don't see a benefit in implementing the DELETED state for clusters. It adds complexity and I do not know what to do with clusters in such state. They won't get displayed anyway.
> We should purely delete those from storage and probably cleanup all tables holding data for them (but the schedules maybe).

I disagree with the approach of deleting the data. The Correct™ way forward is to handle clusters with different states. The added benefits it comes with does introduce some extra overhead.

Fortunately much of the ground work for a more manageable code and design around cluster objects in this PR is independent of whether we go for cluster states or not. This groundwork is in the first two commits. 

Let's focus on getting that work into master while we parking lot the cluster state discussion and design. 
The cleanup and ground work is in https://github.com/thelastpickle/cassandra-reaper/pull/733

michaelsembwever:
Closing out. First two commits got merged in #733. We're re-evaluating how to do/introduce `Cluster.State`…

PR #720, opened 2019-07-17
Add authorization to rest endpoints 

michaelsembwever:
…based on their verbs (http method), add "operator" (read+write) and "user" (read-only) roles, let the existing admin user be an operator and add a new user "user" with the "user" role.

### DIFF
--- a/src/server/src/main/docker/configure-webui-authentication.sh
+++ b/src/server/src/main/docker/configure-webui-authentication.sh
@@ -38,6 +38,6 @@ fi
 
 if [ ! -z "${REAPER_AUTH_USER}" ]; then
 cat <<EOT2 >> /etc/shiro.ini
-${REAPER_AUTH_USER} = ${REAPER_AUTH_PASSWORD}
+${REAPER_AUTH_USER} = ${REAPER_AUTH_PASSWORD}, operator
 EOT2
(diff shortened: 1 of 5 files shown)


adejanovski:
That's a great first step for dealing with authorization in Reaper.
Maybe we should make the UI handle this as well so that we don't end up with buttons that are clickable but triggering operations that fail.
We could add a dummy POST operation that we would use to test the rights of the currently connected user and disable buttons/forms that trigger something else than reads. wdyt?

michaelsembwever:
>  Maybe we should make the UI handle this as well so that we don't end up with buttons that are clickable but triggering operations that fail.
> We could add a dummy POST operation that we would use to test the rights of the currently connected user and disable buttons/forms that trigger something else than reads. wdyt?

i'd rather push that out to a follow up task. (at least it is not a quick task for me.)

One thing outstanding here is the `Unauthorized` error message that gets displayed is shown as raw html. The UI needs to parse the html error message in some manner…

michaelsembwever:
> One thing outstanding here is the Unauthorized error message that gets displayed is shown as raw html. The UI needs to parse the html error message in some manner…

Done.

adejanovski (approved):

PR #724, opened 2019-07-30
Fix travis deploy

adejanovski:
The Makefile couldn't evaluate anymore the maven command to parse the version number.
Fixed by exporting the VERSION var in before_deploy.sh and then use it directly in the Makefile.

### DIFF
--- a/src/packaging/Makefile
+++ b/src/packaging/Makefile
@@ -13,12 +13,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Using mvn:
-VERSION := `mvn org.apache.maven.plugins:maven-help-plugin:2.1.1:evaluate -Dexpression=project.version -f ../../pom.xml | grep -v '\['`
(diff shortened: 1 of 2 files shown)


src/packaging/Makefile
```diff
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 # Using mvn:
-VERSION := `mvn org.apache.maven.plugins:maven-help-plugin:2.1.1:evaluate -Dexpression=project.version -f ../../pom.xml | grep -v '\['`
+#VERSION := `mvn org.apache.maven.plugins:maven-help-plugin:2.1.1:evaluate -Dexpression=project.version -f ../../pom.xml | grep -v '\['`
```
michaelsembwever: can we just remove the line now?

michaelsembwever (approved):
LGTM :shipit: 

why does/would it not apply to `1.4` as well?

PR #730, opened 2019-07-31
Fix the find command to correctly identify the reaper jar on all systems while excluding source/javadoc jars

adejanovski:


michaelsembwever (approved):
LGTM.

but don't you want this into `1.4` rather than `master`?

adejanovski:
replaced by #731

PR #732, opened 2019-07-31
Self register clusters in sidecar mode since the nodes are collocated with Reaper

adejanovski:
In order to achieve this, I did the following:

To avoid code duplication, I made public the `addOrUpdateCluster()` method from `ClusterResource` so that it could be invoked upon startup, when initializing the sidecar mode.
Since the instances will now self register the cluster in sidecar mode, it was required to update the integration tests so that they wouldn't check the number of clusters in storage for sidecar, and retrieve the cluster name even if it already existed.
The jetty runners needed to have an additional member that exposes their `datacenterAvailability` to the Cucumber tests.

### DIFF
--- a/src/packaging/resource/cassandra-reaper-cassandra-sidecar.yaml
+++ b/src/packaging/resource/cassandra-reaper-cassandra-sidecar.yaml
@@ -22,16 +22,18 @@ repairIntensity: 0.9
 scheduleDaysBetween: 7
 repairRunThreadCount: 15
 hangingRepairTimeoutMins: 30
-storageType: memory
+storageType: cassandra
 enableCrossOrigin: true
 incrementalRepair: false
 blacklistTwcsTables: true
 enableDynamicSeedList: true
 repairManagerSchedulingIntervalSeconds: 10
+activateQueryLogger: false
 jmxConnectionTimeoutInSeconds: 5
 useAddressTranslator: false
 # purgeRecordsAfterInDays: 30
 # numberOfRunsToKeepPerUnit: 10
+enableConcurrentMigrations: true
 
 # datacenterAvailability has three possible values: ALL | LOCAL | EACH | SIDECAR
 # the correct value to use depends on whether jmx ports to C* nodes in remote datacenters are accessible
@@ -42,19 +44,8 @@ useAddressTranslator: false
 # If jmx access is restricted to localhost, then configure to SIDECAR.
 #
 # The default is ALL
-datacenterAvailability: ALL
+datacenterAvailability: SIDECAR
 
-
-# any cassandra nodes on 127.0.0.x addresses we presume are ccm nodes
-jmxPorts:
-  127.0.0.1: 7100
-  127.0.0.2: 7200
-  127.0.0.3: 7300
-  127.0.0.4: 7400
-  127.0.0.5: 7500
-  127.0.0.6: 7600
-  127.0.0.7: 7700
-  127.0.0.8: 7800
 
 #jmxAuth:
 #  username: myUsername
@@ -63,6 +54,14 @@ jmxPorts:
 logging:
   level: INFO
   loggers:
+    com.datastax.driver.core.QueryLogger.NORMAL:
+      level: DEBUG
+      additive: false
+      appenders:
+        - type: file
+          currentLogFilename: /var/log/cassandra-reaper/query-logger.log
+          archivedLogFilenamePattern: query-logger-%d.log.gz
+          archivedFileCount: 2
     io.dropwizard: WARN
     org.eclipse.jetty: WARN
   appenders:
@@ -88,6 +87,18 @@ server:
   requestLog:
     appenders: []
 
+cassandra:
+  clusterName: "test"
+  contactPoints: ["127.0.0.1"]
+  keyspace: reaper_db
+  loadBalancingPolicy:
+    type: tokenAware
+    shuffleReplicas: true
+    subPolicy:
+      type: dcAwareRoundRobin
+      localDC:
+      usedHostsPerRemoteDC: 0
+      allowRemoteDCsForLocalConsistencyLevel: false
 autoScheduling:
   enabled: false
   initialDelayPeriod: PT15S
--- a/src/server/src/main/java/io/cassandrareaper/ReaperApplication.java
+++ b/src/server/src/main/java/io/cassandrareaper/ReaperApplication.java
@@ -19,6 +19,7 @@ package io.cassandrareaper;
 
 import io.cassandrareaper.ReaperApplicationConfiguration.DatacenterAvailability;
 import io.cassandrareaper.ReaperApplicationConfiguration.JmxCredentials;
+import io.cassandrareaper.core.Cluster;
 import io.cassandrareaper.core.Node;
 import io.cassandrareaper.jmx.ClusterFacade;
 import io.cassandrareaper.jmx.JmxConnectionFactory;
@@ -45,6 +46,7 @@ import io.cassandrareaper.storage.PostgresStorage;
 
 import java.util.EnumSet;
 import java.util.Map;
+import java.util.Optional;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 
@@ -254,7 +256,7 @@ public final class ReaperApplication extends Application<ReaperApplicationConfig
     }
 
     initializeJmxSeedsForAllClusters();
-    maybeInitializeSidecarMode();
+    maybeInitializeSidecarMode(addClusterResource);
     LOG.info("resuming pending repair runs");
 
     Preconditions.checkState(
@@ -314,10 +316,11 @@ public final class ReaperApplication extends Application<ReaperApplicationConfig
 
   /**
    * If Reaper is in sidecar mode, grab the local host id and the associated broadcast address
+   * @param addClusterResource
    *
    * @throws ReaperException any caught runtime exception
    */
-  private void maybeInitializeSidecarMode() throws ReaperException {
+  private void maybeInitializeSidecarMode(ClusterResource addClusterResource) throws ReaperException {
     if (context.config.isInSidecarMode()) {
       ClusterFacade clusterFacade = ClusterFacade.create(context);
       Node host = Node.builder().withHostname(context.config.getEnforcedLocalNode().orElse("127.0.0.1")).build();
@@ -327,11 +330,29 @@ public final class ReaperApplication extends Application<ReaperApplicationConfig
                 .orElse(clusterFacade.getLocalEndpoint(host));
 
         LOG.info("Sidecar mode. Local node is : {}", context.localNodeAddress);
+        selfRegisterClusterForSidecar(addClusterResource, context.config.getEnforcedLocalNode().orElse("127.0.0.1"));
       } catch (RuntimeException | InterruptedException | ReaperException e) {
         LOG.error("Failed connecting to the local node in sidecar mode {}", host, e);
         throw new ReaperException(e);
       }
     }
+  }
+
+  private boolean selfRegisterClusterForSidecar(ClusterResource addClusterResource, String seedHost)
+      throws ReaperException {
+    final Optional<Cluster> cluster = addClusterResource.findClusterWithSeedHost(seedHost, Optional.empty());
+    if (!cluster.isPresent()) {
+      return false;
+    }
+
+    if (context.storage.getClusters().stream().noneMatch(c -> c.getName().equals(cluster.get().getName()))) {
+      LOG.info("registering new cluster : {}", cluster.get().getName());
+
+      // it is ok for this be called in parallel by different sidecars on the same cluster,
+      // so long we're not split-brain
+      context.storage.addCluster(cluster.get());
+    }
+    return true;
   }
 
   private void scheduleRepairManager(ScheduledExecutorService scheduler) {
--- a/src/server/src/main/java/io/cassandrareaper/ReaperApplicationConfiguration.java
+++ b/src/server/src/main/java/io/cassandrareaper/ReaperApplicationConfiguration.java
@@ -150,6 +150,10 @@ public final class ReaperApplicationConfiguration extends Configuration {
   @JsonProperty
   private Optional<String> enforcedLocalDatacenter = Optional.empty();
 
+  @JsonProperty
+  @DefaultValue("false")
+  private Boolean enableConcurrentMigrations;
+
   public int getSegmentCount() {
     return segmentCount == null ? 0 : segmentCount;
   }
@@ -434,7 +438,13 @@ public final class ReaperApplicationConfiguration extends Configuration {
     this.enforcedLocalDatacenter = enforcedLocalDatacenter;
   }
 
+  public boolean getEnableConcurrentMigrations() {
+    return this.enableConcurrentMigrations == null ? false : this.enableConcurrentMigrations;
+  }
 
+  public void setEnableConcurrentMigrations(boolean enableConcurrentMigrations) {
+    this.enableConcurrentMigrations = enableConcurrentMigrations;
+  }
 
   public static final class JmxCredentials {
 
--- a/src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
+++ b/src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
@@ -41,7 +41,6 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.stream.Collectors;
 
-import javax.annotation.Nullable;
 import javax.ws.rs.DELETE;
 import javax.ws.rs.GET;
 import javax.ws.rs.POST;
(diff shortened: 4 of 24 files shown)


src/server/src/test/java/io/cassandrareaper/acceptance/ReaperTestJettyRunner.java
```diff
@@ -51,10 +52,16 @@
...
+  public final DatacenterAvailability datacenterAvailability;
+
   final ReaperJettyTestSupport runnerInstance;
   private SimpleReaperClient reaperClientInstance;
 
-  public ReaperTestJettyRunner(String yamlConfigFile, Optional<String> version) {
+  public ReaperTestJettyRunner(
+      String yamlConfigFile,
+      Optional<String> version,
+      Optional<DatacenterAvailability> dcAvailability) {
```
michaelsembwever: this added parameter is not needed. The `DatacenterAvailability` is already specified in the yaml, and available via `getContext().config`

src/server/src/test/java/io/cassandrareaper/acceptance/ReaperCassandraIT.java
```diff
@@ -103,7 +103,7 @@ public void upgradeReaperRunner(Optional<String> version) throws InterruptedExce
   }
 
   private static void createReaperTestJettyRunner(Optional<String> version) throws InterruptedException {
-    ReaperTestJettyRunner runner = new ReaperTestJettyRunner(CASS_CONFIG_FILE, version);
+    ReaperTestJettyRunner runner = new ReaperTestJettyRunner(CASS_CONFIG_FILE, version, Optional.empty());
```
michaelsembwever: Even if this added parameter is needed (i don't think it is) this is the wrong value to be passing in here.

src/server/src/main/java/io/cassandrareaper/ReaperApplication.java
```diff
@@ -332,13 +335,22 @@ private void maybeInitializeSidecarMode() throws ReaperException {
...
       }
     }
   }
 
+  private void selfRegisterClusterForSidecar(ClusterResource addClusterResource) {
+    addClusterResource.addOrUpdateCluster(
+        Optional.empty(),
+        Optional.empty(),
+        Optional.of("127.0.0.1"),
+        Optional.of(Cluster.DEFAULT_JMX_PORT));
```
michaelsembwever:
```
  public boolean selfRegister(String seedHost, int jmxPort) {

    final Cluster cluster = findClusterWithSeedHost(seedHost, jmxPort);
    if (null == cluster) {
      return false; // shutdown reaper? no point in starting side-car when C* is not there and running first…
    }

    if (context.storage.getClusters().stream().noneMatch(c -> c.getName().equals(cluster.getName()))) {
      LOG.info("registering new cluster : {}", cluster.getName());

      // it is ok for this be called in parallel by different sidecars on the same cluster, so long we're not split-brain
      context.storage.addCluster(cluster);

      if (context.config.hasAutoSchedulingEnabled()) {
        try {
          clusterRepairScheduler.scheduleRepairs(cluster);
        } catch (ReaperException e) {
          LOG.error("failed to automatically schedule repairs for sidecar cluster {}", cluster.getName(), e);
          return false; // what to do in this situation for sidecar??
        }
      }
    }
    return true;
  }
```

src/server/src/test/java/io/cassandrareaper/acceptance/ReaperTestJettyRunner.java
```diff
@@ -17,6 +17,7 @@
 
 package io.cassandrareaper.acceptance;
 
+import io.cassandrareaper.ReaperApplicationConfiguration.DatacenterAvailability;
```
michaelsembwever: until we figure out https://github.com/thelastpickle/cassandra-reaper/pull/641 it's problematic to add production classes to imports here.

src/server/src/test/java/io/cassandrareaper/acceptance/ReaperTestJettyRunner.java
adejanovski: Good point, I hadn't noticed I could do this.

src/server/src/test/java/io/cassandrareaper/acceptance/ReaperTestJettyRunner.java
adejanovski: Removed since I can use `getContext()`

src/server/src/test/java/io/cassandrareaper/acceptance/ReaperCassandraIT.java
adejanovski: Removed since I can use `getContext()`

src/server/src/main/java/io/cassandrareaper/ReaperApplication.java
adejanovski: I don't see a benefit in duplicating some of the logic from `addOrUpdateCluster()` here.
Also, Cassandra could be down on the local node when Reaper starts, so it's fine to continue startup even if adding the cluster fails.
Auto scheduling will kick in eventually as well.

michaelsembwever: > I don't see a benefit in duplicating some of the logic from addOrUpdateCluster() here.

`addOrUpdateCluster()` in its current form returns a Response and is intended for rest clients.
while this code is definitely not a rest client. there's value in having a more appropriate api signature. (dry can do as much damage as good sometimes.)

the original method also has additional and different logic to what is needed here: updating existing clusters is not required, and as you point out how to handle a down C* node, and/or auto-scheduling failing, is to be dealt with differently here.

michaelsembwever: i'm easy which way you go, just pointing out the mismatches… 

i do think we need to be clear and decisive about what functionality we are re-using here, particularly that around the "updating an existing cluster with a new node list"

michaelsembwever (approved):

src/packaging/resource/cassandra-reaper-cassandra-sidecar.yaml
```diff
@@ -0,0 +1,126 @@
...
+segmentCountPerNode: 16
+repairParallelism: DATACENTER_AWARE
+repairIntensity: 0.9
+scheduleDaysBetween: 7
+repairRunThreadCount: 15
+hangingRepairTimeoutMins: 30
+storageType: cassandra
+enableCrossOrigin: true
+incrementalRepair: false
+blacklistTwcsTables: false
```
michaelsembwever: trivial: we don't recommend enabling `blacklistTwcsTables` by default now?

src/packaging/resource/cassandra-reaper-cassandra-sidecar.yaml
```diff
@@ -0,0 +1,126 @@
...
+enableDynamicSeedList: true
+repairManagerSchedulingIntervalSeconds: 10
+activateQueryLogger: false
+jmxConnectionTimeoutInSeconds: 5
+useAddressTranslator: false
+# purgeRecordsAfterInDays: 30
+# numberOfRunsToKeepPerUnit: 10
+enableConcurrentMigrations: true
+
+# datacenterAvailability has three possible values: ALL | LOCAL | EACH
```
michaelsembwever: trivial: this comment now needs to include info about `SIDECAR` (same goes for all our *.yaml files)

src/server/src/main/java/io/cassandrareaper/ReaperApplicationConfiguration.java
```diff
@@ -150,6 +150,9 @@
   @JsonProperty
   private Optional<String> enforcedLocalDatacenter = Optional.empty();
 
+  @JsonProperty
+  private Optional<Boolean> enableConcurrentMigrations = Optional.of(false);
```
michaelsembwever: why is this an `Optional`? (i would have thought `false` is the safe default whenever it is unspecified?)

src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
```diff
@@ -241,8 +241,8 @@ private Response addOrUpdateCluster(
     return Response.created(location).build();
   }
 
-  @Nullable // if cluster can't be found over jmx
-  private Cluster findClusterWithSeedHost(String seedHost, Optional<Integer> jmxPort) {
+  @Nullable // if cluster can't be found
+  public Cluster findClusterWithSeedHost(String seedHost, Optional<Integer> jmxPort) {
```
michaelsembwever: trivial: ideally we should have changed the return type from `@Nullable` to `Optional<Cluster>` when making the method public.

src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
```diff
@@ -167,14 +169,20 @@ UUID getRepairRunId() {
 
   static int getPossibleParallelRepairsCount(
       Map<List<String>, List<String>> ranges,
-      Map<String, String> hostsInRing) throws ReaperException {
+      Map<String, String> hostsInRing,
+      DatacenterAvailability datacenterAvailability) throws ReaperException {
```
michaelsembwever, line 173: trivial: could the method could have just been made non-static? (not sure how much change that would require in the tests, but i suspect not much…)

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
```diff
@@ -213,92 +219,236 @@ public CassandraStorage(
...
+            List<String> otherRunningReapers
+                = session
+                    .execute("SELECT reaper_instance_host FROM running_reapers")
+                    .all()
+                    .stream()
+                    .map((row) -> row.getString("reaper_instance_host"))
+                    .filter(
+                        (reaperInstanceHost) ->
+                            !AppContext.REAPER_INSTANCE_ADDRESS.equals(reaperInstanceHost))
+                    .collect(Collectors.toList());
```
michaelsembwever: can you restore this to the 5 lines it previously was. this is an excessive amount of wrapping.

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
```diff
@@ -213,92 +219,236 @@ public CassandraStorage(
...
-      } else {
-        LOG.info(
-            String.format("Keyspace %s already at schema version %d", session.getLoggedKeyspace(), currentVersion));
+      if (migrationRequired.get() == true) {
+        int currentVersion = database.getVersion();
+        MigrationRepository migrationRepo = new MigrationRepository("db/cassandra");
+        if (currentVersion < migrationRepo.getLatestVersion()) {
+          if (isInSidecarMode && !concurrentMigrationsAreEnabled) {
+            throw new ReaperException(
+                "Running in sidecar mode and concurrent migrations are not enabled.");
```
michaelsembwever: trivial: i doubt this line was over 120 characters and needed to be wrapped? ide configured wrong?

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
```diff
@@ -109,6 +109,7 @@
 public final class CassandraStorage implements IStorage, IDistributedStorage {
 
   private static final int LEAD_DURATION = 600;
+  private static final UUID SCHEMA_MIGRATION_LOCK_ID = UUIDs.startOf(1);
```
michaelsembwever: a problem i've got with this commit is the way it intrudes a temporary fix onto this class.

i think we agree that it's a temporary fix, that such a distributed migration approach would be better as a feature to the `patka/cassandra-migration` library.
but this patch is going to become difficult/clumsy to undo as it's touches a few different places and methods throughout this class.

would it be possible to encapsulate most of it into an inner class, so that it will be easy to remove when a "proper fix" comes about?

src/server/src/main/java/io/cassandrareaper/storage/cassandra/SchemaMigrationLock.java
```diff
@@ -0,0 +1,172 @@
...
+import com.datastax.driver.core.Session;
+import com.datastax.driver.core.VersionNumber;
+import com.datastax.driver.core.utils.UUIDs;
+import jersey.repackaged.com.google.common.base.Preconditions;
+import org.cognitor.cassandra.migration.Database;
+import org.cognitor.cassandra.migration.MigrationRepository;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public final class SchemaMigrationLock {
```
michaelsembwever: i'm guessing this need not be public (ie can be package-protected)?

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
```diff
@@ -213,92 +212,104 @@ public CassandraStorage(
...
-
-        migrate(database.getVersion(), migrationRepo, session);
-        // some migration steps depend on the Cassandra version, so must be rerun every startup
-        Migration016.migrate(session, keyspace);
-        // Switch metrics table to TWCS if possible, this is intentionally executed every startup
-        Migration021.migrate(session, keyspace);
-      } else {
-        LOG.info(
-            String.format("Keyspace %s already at schema version %d", session.getLoggedKeyspace(), currentVersion));
+        schemaMigrationLock.unlockSchemaMigration();
```
michaelsembwever: shouldn't this be inside a `} finally {` block?

src/server/src/main/java/io/cassandrareaper/storage/cassandra/SchemaMigrationLock.java
```diff
@@ -0,0 +1,172 @@
...
+            // We own the lock and a migration is required
+            migrationRequired = Optional.of(true);
+          } else {
+            // Migration is required but another reaper instance owns the lock
+            LOG.info("Cannot grab the lock on schema migration. Waiting for it to be released...");
+            try {
+              Thread.sleep(10000);
+            } catch (InterruptedException e) {
+              // TODO Auto-generated catch block
+              throw new ReaperException(e);
```
michaelsembwever: swallow or `IllegalStateException`

src/server/src/main/java/io/cassandrareaper/storage/cassandra/SchemaMigrationLock.java
```diff
@@ -0,0 +1,172 @@
...
+          + " (parent_id, started_at)"
+          + " values(?, dateOf(now())) IF NOT EXISTS USING TTL ?");
+      unlockSchemaMigrationPrepStmt = session.prepare("DELETE FROM system_distributed.parent_repair_history"
+          + " WHERE parent_id = ?");
+      renewLockSchemaMigrationPrepStmt = session.prepare("update system_distributed.parent_repair_history USING TTL ?"
+          + " set started_at = dateOf(now()) where parent_id = ?");
+    }
+  }
+
+  public synchronized boolean isSchemaMigrationRequired(com.datastax.driver.core.Cluster cassandra)
```
michaelsembwever: it's not clear why this method is inside `SchemaMigrationLock`, it looks like logic around schema migration that's only trying to optimise when to grab the lock. but that complicates/breaks things.

can't the sleep retry lock loop be moved inside the `lockSchemaMigration()`  method.
that would mean the lock is taken and released every reaper startup, but i think we can live with that as this is intended as a temporary hack in the code.

michaelsembwever: this is what i mean: https://github.com/thelastpickle/cassandra-reaper/commit/eb7a757ba38be468f84c191197be2f9ba915ebfe

src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
adejanovski: Actually, after giving it a try, it involves a lot of additional code to mock stuff. Instantiating a RepairRunner triggers a lot of calls to dependencies.

src/packaging/resource/cassandra-reaper-cassandra-sidecar.yaml
adejanovski: right, done 👍

src/packaging/resource/cassandra-reaper-cassandra-sidecar.yaml
adejanovski: yes, done 👍

src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
adejanovski: ok, done.
Will appear in my next force push.

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
adejanovski: yes, putting it in a `finally` block

src/server/src/main/java/io/cassandrareaper/ReaperApplicationConfiguration.java
michaelsembwever: Null is never exposed, it's only an internal state (which indicates the default value).
Because it is never expose I don't think using and exposing Optional is correct.

(The `@Default` annotation is just for docs is my understanding, and it is up to us to implement it to match the same default value.)

src/server/src/main/java/io/cassandrareaper/storage/cassandra/SchemaMigrationLock.java
michaelsembwever: see what i've done in https://github.com/thelastpickle/cassandra-reaper/commit/eb7a757ba38be468f84c191197be2f9ba915ebfe

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
michaelsembwever: i went further and used try-resources, see what i did in https://github.com/thelastpickle/cassandra-reaper/commit/eb7a757ba38be468f84c191197be2f9ba915ebfe

PR #736, opened 2019-08-05
Small fixes (locking, error handling, memory) to integration tests (2019-07)

michaelsembwever:


src/server/src/main/java/io/cassandrareaper/service/RepairManager.java
```diff
@@ -234,65 +234,63 @@ private void abortSegmentsWithNoLeader(RepairRun repairRun, Collection<RepairSeg
...
+      }
+      return context.storage.getRepairSegment(runId, segmentId).get();
+    } catch (AssertionError error) {
+      throw new ReaperException("lead is already taken on " + runId + ":" + segmentId, new Exception(error));
     }
-
-    return context.storage.getRepairSegment(repairRunId, segmentId).get();
   }
 
   void abortSegments(Collection<RepairSegment> runningSegments, RepairRun repairRun) {
```
adejanovski, line 269: You removed the forced abort feature it seems, which is valuable in the case we have stuck segments in the UI (prevents the requirement to have to modify the storage directly through SQL/CQL).
What's the motivation here?

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
```diff
@@ -290,10 +290,11 @@ private boolean runRepair() {
       }
 
       try (Timer.Context cxt1 = context.metricRegistry.timer(metricNameForRepairing(segment)).time()) {
+        boolean segmentsLocked = false;
         try {
           LOG.debug("Enter synchronized section with segment ID {}", segmentId);
           synchronized (condition) {
-            if (!lockSegmentRunners()) {
+            if (!(segmentsLocked = lockSegmentRunners())) {
```
adejanovski, line 296: This will conflict with a fix I have in #732 : https://github.com/thelastpickle/cassandra-reaper/pull/732/commits/29291f15c9ac11b3995e1f5995cca7d39e03b1b6
Also, this patch is incomplete as we're still releasing the lock twice when a segment is successfully triggered.

I tried to do it the way you did, but abandoned because the `segmentsLocked` boolean could get set to false by a concurrent thread since it's outside the synchronized section.

For what it's worth, I've tested my fix on several attempts with up to 9 nodes, after getting into stuck repairs. They all passed nicely after that and I had no more errors.

adejanovski (approved):
I can handle the locking improvements while rebasing #732
You can merge this PR.

src/server/src/main/java/io/cassandrareaper/service/RepairManager.java
michaelsembwever: it was only used by one caller, and could be solved with the `tookLead` boolean and by simplifying how  that one called `abortSegments(..)` was forcibly first taking the lead: https://github.com/thelastpickle/cassandra-reaper/blob/master/src/server/src/main/java/io/cassandrareaper/service/RepairManager.java#L254

(it was not like anywhere in the code was actually aborting the segment without holding the lock to begin with. but now the taking and releasing of the lock is better paired and controlled.)

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
michaelsembwever: > I tried to do it the way you did, but abandoned because the segmentsLocked boolean could get set to false by a concurrent thread since it's outside the synchronized section.

No it can't. It's a variable to the method, a separate thread will have its own `segmentsLocked` variable.

> Also, this patch is incomplete as we're still releasing the lock twice when a segment is successfully triggered.

Fixed 👍 
Added the line `segmentsLocked = false;` in
```
            if (0 != repairNo) {
              releaseSegmentRunners();
              segmentsLocked = false;
              processTriggeredSegment(segment, coordinator, repairNo);
            } else {
```

PR #737, opened 2019-08-08
Handle unreachable clusters

michaelsembwever:
Add three states to clusters: `UNKNOWN`, `UNREACHABLE`, and `ACTIVE`; along with a day-timestamp of when the last successful jmx connection was made.

`UNKNOWN` is only used when the `Cluster` object is created on-the-fly to pass through jmx connection details. It is never a state that is persisted in the storage backend.

`ACTIVE` is the default state. It indicates there have been successful jmx connections made against the cluster in the past seven days.

`UNREACHABLE` is the state a cluster is transitioned to when it has not had any successful jmx connections over the past 7 days. This period is configurable via the (undocumented) `ReaperApplicationConfiguration.clusterTimeoutInDays` yaml option.

---
Left to do
- [x] unit tests

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/ReaperApplication.java
+++ b/src/server/src/main/java/io/cassandrareaper/ReaperApplication.java
@@ -492,6 +492,7 @@ public final class ReaperApplication extends Application<ReaperApplicationConfig
           .storage
           .getClusters()
           .parallelStream()
+          .sorted()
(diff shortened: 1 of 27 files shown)


adejanovski (approved):
You'll need a rebase and have a conflict to handle, but otherwise I'm good with merging this.

PR #738, opened 2019-08-09
Expect the right SSTable name in ProgresInfo based on stream direction

rzvoncek:


rzvoncek:
This got in via #740.

PR #740, opened 2019-08-12
Fix how streams are tracked

rzvoncek:
This patch changes how streams are tracked.

The first major change is what get's stored into the `metric` table. Previously, we were putting `Set<CompositeData>` in there. That was not correct because it caused no actual information being stored in the DB. Now we put `List<StreamSession>`.

The second major change has to do with how these stream metrics were read back from the DB. The deserialisation was failing because the target classes didn't have constructors/builders available for Jackson. So I've added the annotations allowing Jackson deserialise the classes properly.

While I was at it, I also moved a whole bunch of parsing code from StreamService to StreamProxy. This way it mirrors how CompactionService and CompactionProxy are built.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/core/Stream.java
+++ b/src/server/src/main/java/io/cassandrareaper/core/Stream.java
@@ -19,6 +19,7 @@ package io.cassandrareaper.core;
 
 import java.util.List;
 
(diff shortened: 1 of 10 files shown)


adejanovski (approved):
LGTM, thanks @rzvoncek

PR #741, opened 2019-08-14
Small fixes: metrics collection and NPE on last contact

adejanovski:
2 small fixes:

- The cluster name was lost in the metrics we store in the backend (`node_metrics_v2`) after the Cluster.Builder PR.
- last contact is null when upgrading from a previous version of Reaper, which will trigger an NPE.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
+++ b/src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
@@ -44,6 +44,7 @@ import io.cassandrareaper.storage.cassandra.SchemaMigrationLock;
 
 import java.io.IOException;
 import java.math.BigInteger;
+import java.time.LocalDate;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Date;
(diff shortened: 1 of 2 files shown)


src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
```diff
@@ -577,14 +578,18 @@ private Cluster parseCluster(Row row) throws IOException {
           ? objectMapper.readValue(row.getString("properties"), ClusterProperties.class)
           : ClusterProperties.builder().withJmxPort(Cluster.DEFAULT_JMX_PORT).build();
 
+    LocalDate lastContact = row.getTimestamp("last_contact") == null
+        ? new java.sql.Date(new Date().getTime()).toLocalDate()
```
michaelsembwever: when null it should be `lastContact = LocalDate.MIN`

(subsequent jmx calls will update it correctly accordingly…)

michaelsembwever (approved):
just the change to `LocalDate.MIN` needed. LGTM otherwise and good to merge.

adejanovski:
> just the change to `LocalDate.MIN` needed. LGTM otherwise and good to merge.

ok, done.
Thanks

PR #746, opened 2019-08-25
Fixed CSS / bootstrap layout issues for cluster-list page

shollingsworth:
- Set layout to be row based for clusters with many DC's / racks in each DC
- Adjusted the cluster JMX info bar to float to the right
    (couldn't figure out how to have it float left :\)
- Could not test against multiple cluster layout, but simulated it in UI only

# Previous Layout:

# Fixed Layout

### DIFF
--- a/src/ui/app/jsx/cluster-list.jsx
+++ b/src/ui/app/jsx/cluster-list.jsx
@@ -142,25 +142,26 @@ const Cluster = React.createClass({
       <div className="panel panel-default" style={clusterDisplayStyle}>
         <div className="panel-body">
           <div className="row">
-            <div className="col-lg-2">
+            <div className="col-lg-auto">
               <div className="row">
+                <button
+                  type="button"
+                  className="pull-right cluster-info-button btn btn-md glyphicon glyphicon-info-sign"
+                  data-toggle="modal"
+                  data-target="#clusterInfoModal">
+                </button>
                 <div className="col-lg-8">
                   <a href={'repair.html?currentCluster=' + this.props.name}><h4>{this.props.name}</h4></a>
-                </div>
-                <div className="col-lg-1">
-                  <button
-                    type="button"
-                    className="cluster-info-button btn btn-lg glyphicon glyphicon-info-sign"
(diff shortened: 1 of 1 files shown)


michaelsembwever:
This is really appreciated @shollingsworth !

My quick and incredibly poor fix on Friday shouldn't have been delayed to see this come in instead :-)
We can give this a test on a number of large multi-DC clusters. Will get back to you.

shollingsworth:
> This is really appreciated @shollingsworth !
> 
> My quick and incredibly poor fix on Friday shouldn't have been delayed to see this come in instead :-)
> We can give this a test on a number of large multi-DC clusters. Will get back to you.

Happy to help. I really dig how the project is setup. Despite being new-ish to Java projects, I was able to run build the jar and test / deploy relatively easily.

Not important, but it would have sped things up quite a bit to have been able to pull a cluster config into the UI when running webpack? Couldn't figure that one out so I had to run a loop on the dev instance and refresh the UI when I ran builds :-P

Cheers :beers:

michaelsembwever:
> Not important, but it would have sped things up quite a bit to have been able to pull a cluster config into the UI when running webpack? Couldn't figure that one out so I had to run a loop on the dev instance and refresh the UI when I ran builds :-P

i think @ossarga can answer that…?

adejanovski:
Hi @shollingsworth,

I've tested your PR and switching the divs to `col-lg-auto` was generating some spacing issues and didn't work nicely in all situations.
we've just merged #753 which should hopefully fix your display issues.
It cleanly limits the number of datacenters per row to 3, puts the racks on distinct rows if the cluster has more than a given number of nodes, uses the same fixed size for all nodes (changing opacity to reflect the data load), and a few other refinements.
The fix is available in `1.4` and in `master`.
Let us know if the changes works for you.

For testing purposes, I've generated a [Wiremock](http://wiremock.org/) profile that implements just the endpoints I needed to display the cluster view.
I need to double check the content and make sure it's all obfuscated correctly and will probably push it to `test/resources` soon.

PR #750, opened 2019-09-10
The setting `enableConcurrentMigrations` should imply that migrations are allowed to be run concurrently

michaelsembwever:
Currently it's implemented as the opposite: it enables locking on schema migrations, which ensures they run sequentially. Such behaviour should be invoked with the use of `enableConcurrentMigrations=false`.


adejanovski (approved):
yes, good point 👍

Issue #770, opened 2019-10-17
1.4.7 on Ubuntu 14.04 issues a lengthy `find` command


rborer:
Hi,

Ubuntu 14.04, I just upgraded from reaper 1.4.4 to 1.4.7 and now it takes a few minutes to start, obviously due to the system-wide `find` command from there: https://github.com/thelastpickle/cassandra-reaper/blob/b891afb7003df0530de666f37dd045b801a30e21/src/packaging/bin/cassandra-reaper#L17

Coming from #731 .

`cassandra-reaper.log` is full of lines like the following:

```
find: `./home/docker': Permission denied
find: `./home/ansible/.ansible': Permission denied
```

Which obviously means that the find command is issued from `/`.

I couldn't understand from the PR what the purpose of this PR is. As a quick fix I can think of adding `maxdepth 4` to it so it doesn't scan the whole disk (imagine a server connected to a slow NFS disk, it'll take ages to start).

What do you think?

adejanovski:
I think it makes sense to do so. Would you be willing to send a PR for this? It would be much appreciated.

PR #779, opened 2019-10-27
Avoid length `find` commands when starting reaper from a higher/top- level directory by limiting find's depth to 4 levels

michaelsembwever:
ref: https://github.com/thelastpickle/cassandra-reaper/issues/770

### DIFF
--- a/src/packaging/bin/cassandra-reaper
+++ b/src/packaging/bin/cassandra-reaper
@@ -14,7 +14,7 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 
-REAPER_JAR=$(find . -regex '.*/cassandra-reaper-.*[0-9rT]\.jar')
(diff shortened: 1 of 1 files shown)


michaelsembwever:
is this what you meant @rborer ?

src/packaging/bin/cassandra-reaper
```diff
@@ -23,7 +23,7 @@ fi
 
 if [ -z "$CLASS_PATH" ]; then
   echo "Looking for reaper in /usr/share/cassandra-reaper/"
```
rborer, line 25: Btw I just noticed that the path that is echoed here doesn't match where it is looking (line just below) :wink:

rborer (approved):
@michaelsembwever :+1: yes exactly what I meant.

adejanovski (approved):

PR #782, opened 2019-10-31
Make the docker image compatible with non-root container environments

mimarpe:
The goal of this PR is to make the docker image compatible with environments where you cannot run the container as root, which is in any case a good practice. My use case is running cassandra-reaper in OpenShift.

- The dependency of su-exec has been removed. The right permissions are given to the files and folders involved before executing the entrypoint.
-  The key point about the permissions rely on the group permissions, since the container user is by default part of the root(0) group. Being in the root group doesn't give you the privileges of the root user.
- The USER 1001 (or other numeric uid) is kind of an artifact needed by OpenShift. OpenShift assings pseudo-random uids and groups to the container users. Example in my case:

$ id
uid=1000210000 gid=0(root) groups=1000210000
$ ps -ef
PID   USER     TIME  COMMAND
    1 10002100  1:24 java -jar /usr/local/lib/cassandra-reaper.jar server /etc/cassandra-reaper.yml

while running the same image in docker vanilla:

/ $ id
uid=1001 gid=0(root)
/ $ ps -ef
PID   USER     TIME  COMMAND
    1 1001      0:08 java -jar /usr/local/lib/cassandra-reaper.jar server /etc/cassandra-reaper.yml

The USER field could be a parameter if running as 1001 (or other numeric uid) instead of "reaper" is bothersome.

More information: https://docs.openshift.com/container-platform/3.3/creating_images/guidelines.html

I have built and pushed the image to dockerhub: **mimarpe/reaper-test**
I see is heavier than the normal cassandra-reaper image, perhaps because of the way I've included the cassandra-reaper.jar...

Thanks a lot for making a review.


mimarpe:
Merging from master instead of 1.4, my mistake. Will re-create

PR #793, opened 2019-11-13
Add "force" flag to cluster delete REST endpoint

michaelsembwever:
Add "force" flag to cluster delete REST endpoint, allowing registered clusters with schedules and repairs runs to be forcibly deleted.

Forcibly deleting a cluster entails first deleting those schedules, repair runs, and repair units belonging to the cluster, before also deleting the cluster.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
+++ b/src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
@@ -20,6 +20,7 @@ package io.cassandrareaper.resources;
 import io.cassandrareaper.AppContext;
 import io.cassandrareaper.ReaperException;
 import io.cassandrareaper.core.Cluster;
+import io.cassandrareaper.core.RepairRun;
 import io.cassandrareaper.jmx.ClusterFacade;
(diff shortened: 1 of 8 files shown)


src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
```diff
@@ -307,28 +308,42 @@ private Cluster updateClusterSeeds(Cluster cluster, String seedHosts) throws Rea
...
+      @PathParam("cluster_name") String clusterName,
+      @QueryParam("force") Optional<Boolean> force) {
+
     LOG.info("delete cluster {}", clusterName);
     try {
-      if (!context.storage.getRepairSchedulesForCluster(clusterName).isEmpty()) {
-        return Response.status(Response.Status.CONFLICT)
-            .entity("cluster \"" + clusterName + "\" cannot be deleted, as it has repair schedules")
-            .build();
+      if (!force.isPresent() || !force.get()) {
```
adejanovski: You could compact this into `if (!force.orElse(false)) {`

adejanovski (approved):
Just a minor nit, otherwise LGTM

PR #803, opened 2019-11-28
Remove leader election in Reaper for schema migrations

adejanovski:
Relying now on the cassandra-migration lib built in consensus that shipped with 2.3.0.
Also compacted migrations 1 to 16 into a single file which breaks upgrade from versions before 1.2.2, but allows much faster and reliable migrations from scratch.


src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
```diff
@@ -239,11 +238,14 @@ private static void initializeAndUpgradeSchema(
...
-    try (
-        SchemaMigrationLock schemaMigrationLock = new SchemaMigrationLock(version, session, config);
-        Database database = new Database(cassandra, config.getCassandraFactory().getKeyspace())) {
+    try (Database database = new Database(cassandra, config.getCassandraFactory().getKeyspace())) {
 
       int currentVersion = database.getVersion();
+      Preconditions.checkState(
+          currentVersion == 0 || currentVersion >= 15,
+          "You need to upgrade from Reaper 1.2.2 at least in order to run this version. "
+          + "Please upgrade to 1.2.2 before performing this upgrade.");
```
michaelsembwever: `"Please upgrade to 1.2.2, or greater, before performing this upgrade."`

src/server/src/main/resources/db/cassandra/016_disable_dclocal_read_repair_chance.cql
```diff
@@ -1,5 +1,5 @@
 --
```
michaelsembwever: Can we change the name of the file?? (eg it would be great to change it to `016_base.cql`, but idk if that's possible)

michaelsembwever:
A few of the `Migration0XX.java` classes also need to be removed.

And the conditional on line https://github.com/thelastpickle/cassandra-reaper/blob/master/src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java#L251 can be removed (the conditional's block just always run, since we're alway past `4` now)

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
adejanovski: done

src/server/src/main/resources/db/cassandra/016_disable_dclocal_read_repair_chance.cql
adejanovski: I thought it would be a problem, but it turns out it's not.

adejanovski:
> A few of the `Migration0XX.java` classes also need to be removed.

Right 👍 

> 
> And the conditional on line https://github.com/thelastpickle/cassandra-reaper/blob/master/src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java#L251 can be removed (the conditional's block just always run, since we're alway past `4` now)

Actually not, because we don't want to run what's in the conditional block if we're initializing the db. I've changed the test to make it against v15 instead of v4.

PR #804, opened 2019-11-28
Add the name of metrics in the detail panel of the repairs

adejanovski:
Metrics in Reaper were previously exposed by repair run. This didn't allow to track repairs across runs, and monitor cyclic repairs on the same keyspace (or set of tables).
Now we provide metrics by repair unit instead, which will allow to associate consecutive repair runs under the same metric.
The Repair unit it (or the metric name) should be provided in the UI to ease up its tracking in monitoring.

Also 2 minor fixes:
- Fixed NPE when `last_contact` is null in postgres/h2
- Fixed the schema in migrations 17&18 for h2, which didn't match the postgres schema


michaelsembwever (approved):
👍

PR #808, opened 2019-12-09
ignore src/packaging/data for apache rat checks

jsanda:
docker-compose stores data files in src/packaging/data, and it was causing the rat checks to fail for me when building locally.


michaelsembwever:
I think this would be better solved by sticking to placing generated files under an existing build directories, instead of in the source tree. (and each src module having only the one "build" directory.)

Can we move the data directory to be under `src/packaging/build/` ? 
(we'd also be able to remove the data line from `.gitignore`)

PR #815, opened 2019-12-17
Allow to renew all seed nodes for a cluster

### LINKED ISSUE #813: Cannot update cluster seed hosts list when previous IPs are unavailable
I manage some Cassandra clusters that are on Kubernetes and nodes often come and go.
New nodes will have different IP addresses.
In some cases, the whole cluster can be recycled with new nodes, which means the seed hosts list for this cluster becomes outdated.
I have a job that periodically updates the seed hosts ([`PUT /cluster/$cluster_name?seedHost=$ip`](https://github.com/thelastpickle/cassandra-reaper/blob/master/src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java#L155-L168)) for all our clusters but in the previous case it won't work since [Reaper tries to get the list of all previous nodes (using the outdated seed hosts)](https://github.com/thelastpickle/cassandra-reaper/blob/master/src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java#L284) and fails (throws a "no host could be reached through JMX" error)

adejanovski:
Fixes #813 

The PR allows to renew all seed nodes for a cluster if they've all been replaced (quite common in k8s deployments).
During an update of the seed nodes, Reaper will first try to get the list of live nodes in the cluster, which triggers an exception if all seed hosts have changed.
The changes here will instead catch the exception and return an empty set for the live nodes. The precondition that follows will succeed if the list of previous nodes is empty and the check was removed from the storage implementation classes as they wouldn't allow this.
I've had to rework the `ClusterResource` class a bit to allow injecting a mock of the `ClusterFacade` class and enable proper testing for that case.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
+++ b/src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
@@ -57,8 +57,10 @@ import javax.ws.rs.core.Response;
 import javax.ws.rs.core.UriInfo;
 
 import com.codahale.metrics.InstrumentedExecutorService;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Splitter;
+import com.google.common.base.Supplier;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
@@ -77,11 +79,22 @@ public final class ClusterResource {
   private final ClusterRepairScheduler clusterRepairScheduler;
   private final ClusterFacade clusterFacade;
 
-  public ClusterResource(AppContext context, ExecutorService executor) {
+  private ClusterResource(AppContext context, ExecutorService executor, Supplier<ClusterFacade> clusterFacadeSupplier)
+      throws ReaperException {
     this.context = context;
     this.executor = new InstrumentedExecutorService(executor, context.metricRegistry);
     this.clusterRepairScheduler = new ClusterRepairScheduler(context);
-    this.clusterFacade = ClusterFacade.create(context);
+    this.clusterFacade = clusterFacadeSupplier.get();
+  }
+
+  @VisibleForTesting
+  static ClusterResource create(AppContext context, ExecutorService executor, Supplier<ClusterFacade> supplier)
+      throws ReaperException {
+    return new ClusterResource(context, executor, supplier);
+  }
+
+  public static ClusterResource create(AppContext context, ExecutorService executor) throws ReaperException {
+    return new ClusterResource(context, executor, () -> ClusterFacade.create(context));
   }
 
   @GET
@@ -281,11 +294,16 @@ public final class ClusterResource {
   private Cluster updateClusterSeeds(Cluster cluster, String seedHosts) throws ReaperException {
     Set<String> newSeeds = parseSeedHosts(seedHosts);
     try {
-      Set<String> previousNodes = ImmutableSet.copyOf(clusterFacade.getLiveNodes(cluster));
+      Set<String> previousNodes = Collections.emptySet();
+      try {
+        previousNodes = ImmutableSet.copyOf(clusterFacade.getLiveNodes(cluster));
+      } catch (ReaperException expected) {
+        // The previous node could all be gone, and that's fine.
+      }
       Set<String> liveNodes = ImmutableSet.copyOf(clusterFacade.getLiveNodes(cluster, newSeeds));
 
       Preconditions.checkArgument(
-          !Collections.disjoint(previousNodes, liveNodes),
+          !Collections.disjoint(previousNodes, liveNodes) || previousNodes.isEmpty(),
           "Trying to update a different cluster using the same name: %s. No nodes overlap between %s and %s",
           cluster.getName(), StringUtils.join(previousNodes, ','), StringUtils.join(liveNodes, ','));
 
--- a/src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
+++ b/src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
@@ -559,20 +559,6 @@ public final class CassandraStorage implements IStorage, IDistributedStorage {
         "Cluster should not be persisted with UNKNOWN state");
 
     Preconditions.checkState(cluster.getPartitioner().isPresent(), "Cannot store cluster with no partitioner.");
-    // assert we're not overwriting a cluster with the same name but different node list
-    Set<String> previousNodes;
-    try {
-      previousNodes = getCluster(cluster.getName()).getSeedHosts();
-    } catch (IllegalArgumentException ignore) {
-      // there is no previous cluster with same name
-      previousNodes = cluster.getSeedHosts();
-    }
-    Set<String> addedNodes = cluster.getSeedHosts();
-
-    Preconditions.checkArgument(
-        !Collections.disjoint(previousNodes, addedNodes),
-        "Trying to add/update cluster using an existing name: %s. No nodes overlap between %s and %s",
-        cluster.getName(), StringUtils.join(previousNodes, ','), StringUtils.join(addedNodes, ','));
 
     return true;
   }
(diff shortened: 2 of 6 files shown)


src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
```diff
@@ -559,20 +559,6 @@ private boolean addClusterAssertions(Cluster cluster) {
...
-    } catch (IllegalArgumentException ignore) {
-      // there is no previous cluster with same name
-      previousNodes = cluster.getSeedHosts();
-    }
-    Set<String> addedNodes = cluster.getSeedHosts();
-
-    Preconditions.checkArgument(
-        !Collections.disjoint(previousNodes, addedNodes),
-        "Trying to add/update cluster using an existing name: %s. No nodes overlap between %s and %s",
-        cluster.getName(), StringUtils.join(previousNodes, ','), StringUtils.join(addedNodes, ','));
```
michaelsembwever, line 575: This checks that the cluster being added doesn't already exist. 

Removing the check is dangerous. Accidentally adding a new cluster that re-uses another clusters name will wipe it.
I rather require the user to delete (unregister) the previous cluster first, especially now that force delete is available.

The check might be bypassed if the cluster is in unreachable state.

src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
```diff
@@ -281,11 +294,16 @@ private Response addOrUpdateCluster(
   private Cluster updateClusterSeeds(Cluster cluster, String seedHosts) throws ReaperException {
     Set<String> newSeeds = parseSeedHosts(seedHosts);
     try {
-      Set<String> previousNodes = ImmutableSet.copyOf(clusterFacade.getLiveNodes(cluster));
+      Set<String> previousNodes = Collections.emptySet();
+      try {
+        previousNodes = ImmutableSet.copyOf(clusterFacade.getLiveNodes(cluster));
+      } catch (ReaperException expected) {
```
michaelsembwever, line 300: What ReaperExceptions would be thrown and caught here. Can't say I'm comfortable with this approach, it should at a minimum check the cluster state.

michaelsembwever:
This removes the check that someone accidentally adds a new cluster over the top of an existing cluster with the same name. That check should remain in place, it does prevent the user from shooting themselves in the foot.

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
adejanovski: but if we keep the check, it's impossible to renew the seed list of a cluster that changed all of its ips, which is what we're trying to do.
How would you change the check to allow this?

Unregistering the cluster would remove the schedules and the history, which we're trying to avoid.

michaelsembwever: > How would you change the check to allow this?

We don't.
It's important that it remains impossible to replace an existing cluster with a new cluster where none of the ipaddresses overlap.

If an existing cluster changes the ipaddress of all its nodes, in a timeframe short enough that Reaper did not notice, then the cluster needs be unregistered before it can be added to Reaper again. We know from experience that relying upon the cluster-name as an identifier is dangerous. The set of ipaddresses is a better identifier to the cluster. 

There is a caveat here: when the cluster has become stale and moves into INACTIVE state. Then it might be an option to make it possible, but i'm still uncomfortable with making this too easy. I would suggest a yaml setting that permits to overriding INACTIVE clusters, which defaults to the safe value of false.

atallahade:
@michaelsembwever: There are two things that we could possibly do:
1. Add a force parameter to bypass the checks
1. Ping all seeds nodes (through JMX I guess) and replace if none of them respond

Problem with 2. is that you might have a case where a cluster is down and another one with the same name gets added.

What do you think?

adejanovski:
Hi @atallahade,

we discussed this with @jsanda and he said we could probably fix your issue without modifying Reaper much:

```
when you set up a cluster in reaper and have to specify a seed node, specifying the cassandra headless service will return all the cassandra endpoints.
reaper will just take the first address in the list which is fine because it can resolve the rest of the addresses from the subsequent jmx calls it makes.
```
 
In order for this to work, you need to set `enableDynamicSeedList: false` in configuration.
@jsanda, after checking, there's one slight change needed in the code to avoid the list of seeds to be refreshed before each repair run : https://github.com/thelastpickle/cassandra-reaper/blob/master/src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java#L297-L305

@atallahade, let us know if that would work for you.

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
jsanda: >If an existing cluster changes the ipaddress of all its nodes, in a timeframe short enough that Reaper did not notice, then the cluster needs be unregistered before it can be added to Reaper again. We know from experience that relying upon the cluster-name as an identifier is dangerous. The set of ipaddresses is a better identifier to the cluster.

I would like to suggest avoid relying on IP addresses as a cluster identifier. It will not work clusters running in Kubernetes. If I have a cluster in Kubernetes and do a rolling restart of the cluster, then every  Cassandra node will wind up with a new IP address.

What about having a UID for each cluster? It could either provided by users/clients or generated by Reaper.

jsanda (commented):
I need to add the check for `enableDynamicSeedList` in `RepairRunner` and do some testing.

Let me to take a step back provide a bit more context for the Kubernetes use cases. In Kubernetes Cassandra is deployed as a [StatefulSet](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/). Pods in a StatefulSet have stable network ids, e.g., `node-0`, `node-1`, etc. Let's say we need to restart the Cassandra node running in pod `node-0`. A restart is accomplished by deleting and recreating the pod. The new pod will have a different ip address; however it will still have the hostname, `node-0`.

A StatefulSet requires a [headless Service](https://kubernetes.io/docs/concepts/services-networking/service/#headless-services). DNS lookups for that Service object will return a list of all ip addresses of the Cassandra pods. That list is automatically updated as pods come and go.

Discovery is one of the primary functions of a k8s service. For Kubernetes deployments I think it would be ideal if Reaper could rely on that headless service as the seed list. I think some changes would be necessary though. The first in `RepairRunning.updateClusterNodeList` is

```java
Set<String> liveNodes  = ImmutableSet.copyOf(clusterFacade.getLiveNodes(cluster));
```
If  you trace the call path of `clusterFacade.getLiveNodes` it leads to:

```java
private JmxProxy connectImpl(Cluster cluster, Collection<String> endpoints) throws ReaperException {
    try {
      JmxProxy proxy = context.jmxConnectionFactory.connectAny(
          endpoints
              .stream()
              .map(host -> Node.builder().withCluster(cluster).withHostname(host).build())
              .collect(Collectors.toList()));

      Async.markClusterActive(cluster, context);
      return proxy;
    } catch (ReaperException ex) {
      Async.markClusterUnreachable(cluster, context);
      throw ex;
    }
  }
```

If we are relying on the headless service, then I think `connectImp` should be changed to something like:

```java
// Maybe it would be better if the k8s service was provided as an explicit argument. Alternatively, I 
// suppose we could just iterate over endpoints, call InetAddress.getAllByName and build up the list.
String k8sService = ...
JmxProxy proxy = context.jmxConnectionFactory.connectAny(
          endpoints
              .stream()
              .map(host -> Node.builder().withCluster(cluster).withHostname(host).build())
              .collect(Collectors.toList()));
```

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
michaelsembwever: Such an UUID would have to be provided by Cassandra itself. We could rely on the Host-IDs instead of the ip-addresses. I believe the Host-IDs stay the same even when you do a `"-Dcassandra.replace_address=…"`

jsanda:
Given that @atallahade is not using a k8s service, my comments  are not entirely relevant to this discussion and PR. 

In Kubernetes Cassandra should be run with StatefulSets which require a headless Service. I now understand that is not the case with this issue; so, I think it makes more sense for me to move the discussions about changes for Kubernetes to a separate issue.

jsanda:
@adejanovski / @michaelsembwever 

#870 should be sufficient for the Kubernetes use case. This PR is not necessary for Kubernetes. If these changes were intended primarily to address that use case, then this can be closed.

adejanovski:
Base issue dealt with in #870

Issue #809, opened 2019-12-09
Pluggable JMX credentials backend


denniskline:
Provide a mechanism for allowing reaper users to define where/how the JMX credentials are stored.  

This was briefly discussed as part of issue 143: https://github.com/thelastpickle/cassandra-reaper/issues/143

When reaper is used to orchestrate repairs across many clusters where clusters are added and removed often and each cluster has different JMX credentials, it is cumbersome to make modifications to the config.yaml on all the running instances of reaper and then restart each instance to pick up the change.  

It would be great if there were options for users to be able to choose where reaper pulled JMX credentials.  Reaper would also need the ability to be able to pull in new JmxCredentials as they are added.  

Along these lines (and perhaps this should be split into a different issue), it would also be nice if these JMX credentials could be encrypted as leaving them as plain text raises red flags with audit it some companies.

If this is something that the community would like, I'd be more than happy to contribute to this issue

adejanovski:
Hi @denniskline, 

we sure need something like this because what we have now is faaaaar from agile (and not secured much).
We've evolved the format of the `cluster` table to allow storing more cluster specific data in the `properties` column (right now just the JMX port). It would be a great place to store the encrypted credentials.
I think a single issue should cover both as storing unencrypted credentials is probably not something we want to do.

We would love to have a contribution on this 😍 
Let us know if we can help you in any way.

denniskline:
@adejanovski - That's great news!  I'll take a stab at it and keep this issue thread updated with questions along the way.  Thank you very much.

Issue #802, opened 2019-11-28
Provide a way to encrypt password in authProvider


Payal147:
Provide a way to encrypt password in authProvider in Cassandra section while using Cassandra as storageType.

"storageType: cassandra
cassandra:
  clusterName: "test"
  contactPoints: ["127.0.0.1"]
  keyspace: reaper_db
  authProvider:
    type: plainText
    username: cassandra
    password: cassandra"

There should be a way to provide an encrypted password.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-412) by [Unito](https://www.unito.io)
┆Issue Number: K8SSAND-412

denniskline:
@Payal147 I have a workaround for this that I can contribute (at least to help this discussion along) that is based on Spring's implementation of encrypted credentials.

It boils down to a custom dropwizard AuthProviderFactory and a change to the config.yaml to reference it.  The authProvider then looks for a system property whose key is defined in systemPropertySecret for the value to decipher the password.

```
  authProvider:
    type: encryptedText
    username: "myusername"
    password: "{cipher}e3a055c901e71c12f0a2f762f30a561bfd01c97d5ecbb430698b99025560144a"
    systemPropertySecret: SYS_PROPERTY_ENCRYPTION_KEY
```

Payal147:
@denniskline Can you elaborate a little bit on how can we define the keys in systemPropertySecret and what system property we need to define. I am not much into these things.

sedulam:
@denniskline is this something that can be configured in Reaper as-is, or is it a new contribution to the project?

denniskline:
Correct @sedulam.  I need to open a PR to address this.  What I pasted above is from local testing.  I'll site this issue in the PR so that the two are linked.

adejanovski:
This was shipped a while ago.
Closing this ticket.

PR #817, opened 2019-12-20
Issue #809 : Allow for JMX Credentials to be stored alongside JMX Port in the backend storage system

denniskline:
Enhance the API endpoints to accept JMX Credentials so that Clusters can be added without needing to edit the reaper yaml and restart the application.  

The JMX Auth and JMX Credentials section are left untouched in the reaper yaml to support backwards compatibility and user preference.  Adding JMX credentials is optional.

The UI has not been enhanced in this PR.  I am planning on making that change in a subsequent PR.  Please let me know if the preferences is to make both changes in a single PR.

I'm unsure how sidecar should manage JMX Credentials since he does a bit of auto discovery.  Suggestions here would be greatly appreciated.


adejanovski:
Thanks @denniskline! 
We'll take a close look at this ASAP.

For the sidecar mode, it works with the assumption that JMX is local only, which is when it's not password protected so I think we're fine. In the worst case, the self registration can fail without preventing the cluster to be added manually afterwards.

A quick glance at the code suggests that the JMX password is stored in plain text, is it the case?
If so we'll need to add encryption because we probably don't want to store plain text passwords in the db.

denniskline:
Appreciate the quick feedback @adejanovski!

Correct, it's plain text at the moment.  I was hoping to introduce encryption with issue #802 and then use the same mechanism for the JMX password.  However, I probably should have either done #802 first or introduced encryption with this PR and then used that for #802
 
Would you like to have encrypted JMX password in this PR?

adejanovski:
You can totally do this in the order you see fit, and we can start reviewing this PR before it gets the encryption. But yes, it would be great to have encryption for the JMX password before we merge it.

Thanks!

denniskline:
Will do @adejanovski, thanks!

src/server/src/main/java/io/cassandrareaper/core/Cluster.java
```diff
@@ -179,6 +194,15 @@ public Builder withJmxPort(int jmxPort) {
       return this;
     }
 
+    public Builder withJmxCredentials(JmxCredentials jmxCredentials) {
+      Preconditions.checkNotNull(jmxCredentials);
+      Preconditions.checkNotNull(jmxCredentials.getUsername());
+      Preconditions.checkNotNull(jmxCredentials.getUsername());
```
michaelsembwever: was this line meant to be 
```suggestion
      Preconditions.checkNotNull(jmxCredentials.getPassword());
```

src/server/src/main/java/io/cassandrareaper/core/ClusterProperties.java
```diff
@@ -23,22 +23,36 @@
 @JsonDeserialize(builder = ClusterProperties.Builder.class)
 public final class ClusterProperties {
   private final int jmxPort;
+  private String jmxUsername;
+  private String jmxPassword;
```
michaelsembwever: both can be `final`

src/server/src/main/java/io/cassandrareaper/jmx/JmxConnectionFactory.java
```diff
@@ -171,37 +165,37 @@ public final HostConnectionCounters getHostConnectionCounters() {
...
+        nodeCredentials = jmxCredentials.get(Cluster.toSymbolicName(node.getClusterName()));
+      }
 
-    // As clusters get stored in the database with their "symbolic name" we have to look for that too
-    if (jmxCredentials != null && jmxCredentials.containsKey(Cluster.toSymbolicName(clusterName))) {
-      jmxCreds = Optional.of(jmxCredentials.get(Cluster.toSymbolicName(clusterName)));
     }
 
-    return jmxCreds;
+    return nodeCredentials == null ? JmxCredentials.builder().build() : nodeCredentials;
```
michaelsembwever: can we use a dedicated instance for no credentials instead please. eg `JmxCredentials.NO_CREDENTIALS`

(( Such a class might then be able to `throw UnsupportedOperation()` in its getters, rather than us having to support nullable getters. ))

src/server/src/main/java/io/cassandrareaper/jmx/JmxConnectionFactory.java
```diff
@@ -171,37 +165,37 @@ public final HostConnectionCounters getHostConnectionCounters() {
...
 
-  public Optional<JmxCredentials> getJmxCredentialsForCluster(String clusterName) {
-    Optional<JmxCredentials> jmxCreds = Optional.ofNullable(jmxAuth);
-    if (jmxCredentials != null && jmxCredentials.containsKey(clusterName)) {
-      jmxCreds = Optional.of(jmxCredentials.get(clusterName));
-    }
+  public JmxCredentials getJmxCredentialsForCluster(Node node) {
+    JmxCredentials nodeCredentials = node.getJmxCredentials().orElse(jmxAuth);
+
+    if (nodeCredentials == null && jmxCredentials != null) {
```
michaelsembwever: if i got this right, this will preference jmx credentials in the following order:
1. backend stored in the backend
2. default credentials stored in the yaml
3. per-cluster credentials stored in the yaml

I would have thought that `per-cluster credentials, even if they are now deprecated, came second in the list?

(Can we also add unit tests for these types of preference checks between the various variant combinations)

denniskline:
Thank you @michaelsembwever, I appreciate the advice/feedback.  I'll get this PR updated.

src/server/src/main/java/io/cassandrareaper/jmx/JmxConnectionFactory.java
denniskline: I took a slightly different path here and switched getJmxCredentialsForCluster(Node) to return an Optional<JmxCredentials>.  Doing this made it easier for the caller to do an .isPresent() check to see if there were credentials or not.  Please let me know if this approach works for you.

denniskline:
The comments in the PR have been addressed.  

Symmetric encryption has been added to encrypt the JMX password as it enters storage and to decrypt it right before it is used to establish a connection.  I added hooks so that asymmetric encryption can be added later (if the community wants more options).  I started with symmetric mainly because it is easier to implement, understand, and configure.

The cypher that is created is prefixed with {cipher}.  I borrowed this approach from Spring and like how it allows for the cryptograph to be used on clear and encrypted text.  This let's people slowly convert to encrypted without needing to introduce workarounds or relying on exceptions and assumptions.

The cypher is encoded as HEX instead of Base64 mainly because of aesthetics.   It just looks cleaner than Base64 and since we are only encrypting a single word, Base64's shorter length does not buy us too much.

I took a stab at where to put these classes and opted for a new package "crypto".  I'm completely open to suggestions here if you want it moved/renamed/etc.  

Currently the cryptograph is being created and stored in the AppContext.  I'm hoping to use this to help in other places: storage password and https/keystore password.

Please let me know thoughts/concerns.  Happy to make changes.  Thanks you.

src/server/src/main/java/io/cassandrareaper/ReaperApplication.java
```diff
@@ -198,8 +202,7 @@ public void run(ReaperApplicationConfiguration config, Environment environment)
       context.jmxConnectionFactory.setJmxCredentials(jmxCredentials);
     }
 
-    context.repairManager = RepairManager.create(
-        context,
+    context.repairManager = RepairManager.create(context,
```
michaelsembwever: nit-pick: each argument on a separate line please, including the first. (this happens in a few places)

src/server/src/main/java/io/cassandrareaper/AppContext.java
```diff
@@ -46,6 +47,7 @@
   public SchedulingManager schedulingManager;
   public JmxConnectionFactory jmxConnectionFactory;
   public ReaperApplicationConfiguration config;
+  public Cryptograph cryptograph;
```
michaelsembwever: nit-pick: can we do inversion-of-control (dependency-injection through the constructors) rather than service lookup via the context. I believe the instance created in ReaperApplication can be passed in to the needed classes via new constructor parameters.

src/server/src/main/java/io/cassandrareaper/jmx/JmxConnectionFactory.java
michaelsembwever: thanks for addressing this @denniskline 👍

src/server/src/main/java/io/cassandrareaper/core/ClusterProperties.java
michaelsembwever: thanks for addressing this 👍

src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java
```diff
@@ -238,14 +256,16 @@ private Response addOrUpdateCluster(
     return Response.created(location).build();
   }
 
-  public Optional<Cluster> findClusterWithSeedHost(String seedHost, Optional<Integer> jmxPort) {
+  public Optional<Cluster> findClusterWithSeedHost(String seedHost, Optional<Integer> jmxPort,
+                                                   Optional<JmxCredentials> jmxCredentials) {
```
michaelsembwever, line 263: nit-pick: if wrapped, one parameter per line pls.

src/server/src/main/java/io/cassandrareaper/resources/CryptoResource.java
```diff
@@ -0,0 +1,56 @@
...
+  public CryptoResource(AppContext context) {
+    this.cryptograph = context.cryptograph;
+  }
+
+  @GET
+  @Path("/encrypt/{text}")
+  public Response encrypt(@PathParam("text") String text) {
+    try {
+      return Response.ok().entity(cryptograph.encrypt(text)).build();
+    } catch (RuntimeException e) {
```
michaelsembwever: question: can this catch be more specific? and the response more specific?

src/server/src/main/java/io/cassandrareaper/core/ClusterProperties.java
```diff
@@ -23,22 +23,36 @@
 @JsonDeserialize(builder = ClusterProperties.Builder.class)
 public final class ClusterProperties {
   private final int jmxPort;
+  private final String jmxUsername;
+  private final String jmxPassword;
```
michaelsembwever: why not just store the `jmxCredentials` here instead? (then we wouldn't have to build it everytime in the cluster's getter method)

src/server/src/main/java/io/cassandrareaper/jmx/JmxConnectionFactory.java
```diff
@@ -171,47 +168,52 @@ public final HostConnectionCounters getHostConnectionCounters() {
     return accessibleDatacenters;
   }
 
-  public Optional<JmxCredentials> getJmxCredentialsForCluster(String clusterName) {
```
michaelsembwever, line 174: can we keep the `getJmxCredentialsForCluster(String clusterName)` method as well? it is the inner block to the `getJmxCredentialsForCluster(Node node)` method, and is used by the `ClusterResource.getCluster(..)` rest endpoint. (It would be nice if that endpoint method didn't have to pass in a blank node as a hack.)

michaelsembwever (changes requested):
The design of the PR is approved. There's lots of fantastic stuff in here! Really really like the approach of pluggable crypto. 

I've thrown at you a ton of little nit-picking-ish comments, that hopefully you're ok with and are painless to do.

src/server/src/main/java/io/cassandrareaper/ReaperApplication.java
denniskline: yep ... this is me being lazy.  checkstyle would only allow 150 line methods and when I added the cryptograph it bumped it up to 153.  i "shortened" the method a bit to make checkstyle happy.  I'll refactor something in there into its own method aand then put these other lines back to what they were.

src/server/src/main/java/io/cassandrareaper/resources/CryptoResource.java
denniskline: Like a ReaperException?  Or are you thinking more along the lines of a new exception type like: CipherException or something?

src/server/src/main/java/io/cassandrareaper/core/ClusterProperties.java
denniskline: ahh, good idea.  absolutely.

denniskline:
Really appreciate the feedback @michaelsembwever!  Happy to make those changes.

src/server/src/main/java/io/cassandrareaper/jmx/JmxConnectionFactory.java
denniskline: It's not entirely a blank node.  It's a node with a cluster but without a hostname.  The cluster is needed because of the potential JmxCredentials that have been set.  I can change this to allow clusterName again but it would need to then also look up the Cluster object.

src/server/src/main/java/io/cassandrareaper/resources/CryptoResource.java
michaelsembwever: I was thinking of keeping it unchecked. But come to think about it doesn't the codebase catch have handle runtime exceptions already for us? (ie you don't need a catch here at all…?)

src/server/src/main/java/io/cassandrareaper/jmx/JmxConnectionFactory.java
michaelsembwever: I think it's ok to look up the cluster object. Here it's happening at a high-level, just once in the request, ie from a call from the `ClusterResource` class.

src/server/src/main/java/io/cassandrareaper/resources/CryptoResource.java
denniskline: oh that is interesting.  i didn't realize that dropwizard had a default Exception mapper.  Yep, I will definitely remove this try/catch.

src/server/src/main/java/io/cassandrareaper/jmx/JmxConnectionFactory.java
denniskline: looking closer, the ClusterResource really does not need to pass the Node into getJmxCredentialsForCluster.  All he really needs is the Cluster object.  Knowing the hostname does not effect the credentials.  How about getJmxCredentialsForCluster accepts an @Nullable Cluster instead of a Node?

michaelsembwever: `Optional<Cluster>` please :)

`@Nullable` shouldn't be used for public method signatures, except those json serialisable.

src/server/src/main/java/io/cassandrareaper/crypto/SymmetricCryptograph.java
```diff
@@ -0,0 +1,259 @@
...
+    String trimmedText = StringUtils.trimToNull(plainText);
+    Preconditions.checkNotNull(trimmedText);
+
+    try {
+      String passphrase = fetchPassphrase();
+      return PREFIX + encryptText(passphrase, trimmedText);
+    } catch (NoSuchAlgorithmException | NoSuchPaddingException | InvalidKeySpecException
+            | InvalidAlgorithmParameterException | InvalidKeyException | BadPaddingException
+            | IllegalBlockSizeException e) {
+      throw new RuntimeException("Unable to encrypt text", e);
```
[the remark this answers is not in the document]
denniskline: I really like the idea of trapping for errors in the builder but ended up just switching to an IllegalStateException.  The builder values can depend on each other and would result in a lot of conditional checks (AES vs DES for instance).  As I played around with validations, it grew more and more complicated so I reverted.

denniskline:
I believe all the changes are now committed.  Please let me know if something else needs addressed.  Thanks!

michaelsembwever (approved):
LGTM!

could you squash the commits into one nice commit please, and remove the `merge from master` commits (rebasing the branch off master instead).

denniskline:
Please decline this PR.  I had a bit of a snafu squashing and recreated the PR under:
https://github.com/thelastpickle/cassandra-reaper/pull/837

Sorry about that

PR #818, opened 2019-12-23
Add repair unit id to cluster status

adejanovski:
Also adds a lost commit on allowing to use `force` in cluster deletion and a small fix on the spreaper `list_runs()` command which didn't use the right names for query params.


michaelsembwever (approved):
:shipit:

PR #837, opened 2020-01-15
Issue #809 : Allow for JMX Credentials to be strored alongside JMX Po…

denniskline:
…rt in the backend storage system.

This is a fresh PR spun off of:
https://github.com/thelastpickle/cassandra-reaper/pull/817
I had some issues squashing that PR so I cherry-picked into this PR and squashed.

Thank you

### DIFF
--- a/src/docs/content/docs/api.md
+++ b/src/docs/content/docs/api.md
@@ -58,14 +58,32 @@ This operation expects that a call was previously made to `/login` and that the 
   * Expected query parameters:
       * *seedHost*: Host name or IP address of the added Cassandra
         clusters seed host.
+      * *jmxPort*: A custom port that reaper will use to JMX into the Cluster.  Defaults to 7199.  (Optional)
+  * Adds a new cluster to the service, and returns the newly added cluster object, if the operation was successful. If the cluster is already registered, the list of host will be updated to match the current topology.
+
+* **POST    /cluster/auth**
+  * Expected query parameters:
+      * *seedHost*: Host name or IP address of the added Cassandra
+        clusters seed host.
+      * *jmxPort*: A custom port that reaper will use to JMX into the Cluster.  Defaults to 7199.  (Optional)
+      * *jmxUsername*: JMX Username specific to the Cluster.  Defaults to what is defined in [configuration](http://cassandra-reaper.io/docs/configuration/reaper_specific/).  (Optional)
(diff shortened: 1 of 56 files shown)


adejanovski:
Hi @denniskline,

I'll shortly run some tests on this PR and merge it if it's all good.

adejanovski (changes requested):
Hi @denniskline,

this is an awesome addition, thanks a lot!
I have 3 small pain points that I think we should address to enforce security:

- By default, encryption is disabled. I think we should never store passwords in plain text in the database, and should fail with an appropriate message if ever someone tried to add a cluster with JMX authentication with encryption off. 
- the sample config files do not include any sample of how to enable encryption (under `src/packaging/resources`)
- Reaper logs all requests and shows the password in plain text : 
`INFO   [2020-01-24 11:18:01,729] [dw-54 - POST /cluster?seedHost=127.0.0.1&jmxPort=7100&jmxUsername=cassandra&jmxPassword=cassandrapassword] i.c.r.ClusterResource - creating new cluster based on given seed host: c3.0.15 `. 
I think we would be fine if the password was sent as post body instead of a query param. It would then not show up in the logs.

Other from these small changes, the PR looks fantastic to me!

denniskline:
@adejanovski great feedback.  I really appreciate you and @michaelsembwever dedicating the time to help hash out this feature.  

New commit has been squashed.  Please let me know if there are any additional changes you would like to see.

Thank you very much

adejanovski:
Hi @denniskline, 

Thanks for making those changes, it's all safe now 👍 
A few more issues:
- Changing the cluster resource as we did there (using form data with the post/put operations) will break backwards compatibility
- spreaper wasn't updated to register clusters the "new" way
- Integration tests are broken because they try to use the old POST operation with query params
- We had no integration tests for various setups arounds jmx credentials (I tested registering with credentials, without credentials, and with credentials but no encryption system env variable)

Here's a commit that fixes all these issues which you can cherry pick : https://github.com/thelastpickle/cassandra-reaper/commit/9d344409243c014dd2000fa14895322cc13e050e

I've updated the API docs to show both operations for PUT/POST in the cluster resource. The version that has JMX credentials was pushed under `/cluster/auth` to distinguish both endpoints.

I tried to push it to your branch directly but you didn't allow committers to push there :) 
Let me know if that works for you.

denniskline:
@adejanovski cherry-pick'd and squashed.  very sorry for breaking backwards compatibility.  thank you again for fixing!!

PR #839, opened 2020-01-16
Including spreaper cli to docker image

### LINKED ISSUE #838: Include spreaper on docker image
It would be very useful to include `spreaper` on the docker image
https://github.com/thelastpickle/cassandra-reaper/blob/master/src/packaging/bin/spreaper

That way, I could use `spreaper` to automatically include my cassandra cluster inside reaper.

velo:
Fixes #838

### DIFF
--- a/src/server/src/main/docker/Dockerfile
+++ b/src/server/src/main/docker/Dockerfile
@@ -88,6 +88,11 @@ ADD configure-webui-authentication.sh /usr/local/bin/configure-webui-authenticat
 ADD ${SHADED_JAR} /usr/local/lib/cassandra-reaper.jar
 ADD spreaper /usr/local/bin/spreaper
 
+
+# get around `/usr/local/bin/configure-persistence.sh: line 65: can't create /etc/cassandra-reaper.yml: Interrupted system call` unknown error
+RUN touch /etc/cassandra-reaper.yml
+
+
 RUN addgroup -g 1001 -S reaper && adduser -G reaper -S -u 1001 reaper && \
     mkdir -p /var/lib/cassandra-reaper && \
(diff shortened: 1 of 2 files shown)


velo:
I went a bit out of scope and also changed the image to include a `reaper` user, with all files owned by it.

src/server/src/main/docker/Dockerfile
```diff
@@ -71,45 +71,51 @@ ENV REAPER_SEGMENT_COUNT=200 \
...
 ADD configure-webui-authentication.sh /usr/local/bin/configure-webui-authentication.sh
 ADD ${SHADED_JAR} /usr/local/lib/cassandra-reaper.jar
+ADD spreaper /usr/local/bin/spreaper
 
-# get around `/usr/local/bin/configure-persistence.sh: line 65: can't create /etc/cassandra-reaper.yml: Interrupted system call` unknown error
-RUN touch /etc/cassandra-reaper.yml
-
-RUN mkdir -p /var/lib/cassandra-reaper && \
+RUN addgroup -g 1001 -S reaper && adduser -G reaper -S -u 1001 reaper && \
+    mkdir -p /var/lib/cassandra-reaper && \
```
adejanovski, line 97: @velo, could we keep the `touch` on the yaml file?
We've seen that this could prevent the docker image from working properly under some circumstances, especially when you set it up with a storage backend or enable shiro authentication.
Any specific reason why you removed it?

adejanovski (changes requested):
Happy overall with the changes, but I'd like to have the `touch` back on the yaml file. Removing it backfired last time we did it.

adejanovski (approved):

Issue #376, opened 2018-03-22
segment stuck and always failing


vrischmann:
Hi,

I noticed today there's a segment stuck since approximately 5 days, the fail count is at 661 now.

The only exceptions I see in the log are these:

    DEBUG  [2018-03-22 16:39:26,393] [pool-9-thread-7] i.c.j.JmxProxy - No repair has run yet on the node. Ignoring exception.
    javax.management.InstanceNotFoundException: org.apache.cassandra.internal:type=AntiEntropySessions
            at com.sun.jmx.interceptor.DefaultMBeanServerInterceptor.getMBean(DefaultMBeanServerInterceptor.java:1095)
            at com.sun.jmx.interceptor.DefaultMBeanServerInterceptor.getAttribute(DefaultMBeanServerInterceptor.java:643)
            at com.sun.jmx.mbeanserver.JmxMBeanServer.getAttribute(JmxMBeanServer.java:678)
            at javax.management.remote.rmi.RMIConnectionImpl.doOperation(RMIConnectionImpl.java:1445)
            at javax.management.remote.rmi.RMIConnectionImpl.access$300(RMIConnectionImpl.java:76)
            at javax.management.remote.rmi.RMIConnectionImpl$PrivilegedOperation.run(RMIConnectionImpl.java:1309)
            at javax.management.remote.rmi.RMIConnectionImpl.doPrivilegedOperation(RMIConnectionImpl.java:1401)
            at javax.management.remote.rmi.RMIConnectionImpl.getAttribute(RMIConnectionImpl.java:639)
            at sun.reflect.GeneratedMethodAccessor351.invoke(Unknown Source)
            at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	    ...

there are quite a lot of them but from what I remember these exceptions should be harmless so I'm not sure if it's relevant.

Otherwise I noticed a couple of things which seem weird to me. Here it looks like the segment has finished but yet is detected has having timed out ?

    DEBUG  [2018-03-22 15:00:54,424] [bep_cluster] i.c.j.JmxProxy - Received notification: javax.management.Notification[source=repair:10385][type=progress][message=Repair command #10385 finished in 3 minutes 25 seconds] with type progress
[...]

PR #847, opened 2020-02-12
Avoid aborting repair segments too early

smarsching:
There was a bug which would sometimes cause repair segments to be aborted before the timeout was reached, resulting in the repair process to get stuck.

We experienced this problem with Cassandra Reaper 2.0.2 (and older versions) running on Ubuntu 18.04 LTS with OpenJDK 1.8.0.131.

The problem was in `SegmentRunner.processTriggeredSegment`. Inside the loop, the code would check whether at least 60 seconds have passed since the last iteration. However, with the version of the JRE specified above, `condition.await(…)` would often return a bit early (e.g. after 59997 ms instead of 60000 ms). In this case, the segment would be aborted even though the repair was still running.

This PR fixes this problem by checking the return value of `condition.await(…)` instead of checking the time that has passed since the last iteration. `condition.await(…)` returns `true` if and only if `condition.signalAll()` has been called. This means that the loop will now only be quit if the repair has finished (state `DONE`) or the repair function has returned for some other reason. The timeout is still handled correctly by the loop condition.

I have deployed Cassandra Reaper 2.0.2 with the patch applied on one of our clusters and it seems to reliably resolve the problem for us.

BTW: I believe this PR might fix the issue reported as #376.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
+++ b/src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
@@ -396,21 +396,17 @@ final class SegmentRunner implements RepairStatusHandler, Runnable {
       final long startTime = System.currentTimeMillis();
       final long maxTime = startTime + timeoutMillis;
       final long waitTime = Math.min(timeoutMillis, 60000);
-      long lastLoopTime = startTime;
 
       while (System.currentTimeMillis() < maxTime) {
-        condition.await(waitTime, TimeUnit.MILLISECONDS);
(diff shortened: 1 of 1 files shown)


adejanovski (approved):
Very good catch! It's much cleaner this way indeed.
This it's a bug fix, I'd like to merge this PR on the `2.0` branch instead.
Could you rebase on `2.0` and push your branch again? Then we'll edit the base branch for the PR and I'll be able to merge.
Thanks!

smarsching:
I just rebased the PR against the 2.0 branch.

adejanovski:
Awesome, thanks @smarsching !

PR #856, opened 2020-02-26
Fix large partitions in the metrics v2 table for the sidecar

adejanovski:
Introduce a new metrics table to reduce disk usage and partition sizes


src/server/src/main/java/io/cassandrareaper/jmx/ClusterFacade.java
```diff
@@ -536,7 +536,7 @@ public boolean nodeIsAccessibleThroughJmx(String nodeDc, String node) {
                 Optional.of(node.getHostname()),
                 "org.apache.cassandra.metrics",
                 "ClientRequest",
-                DateTime.now().minusMinutes(1).getMillis()));
+                DateTime.now().minusMinutes(11).getMillis()));
```
michaelsembwever: `11` comes across as a magic number here.
Can you explain why 11 minutes, eg the relationship to the schema (or whatever)

adejanovski: The table is partitioned in chunks of 10 minutes. This allows to include the previous partition in the call if the current one has no metric yet.

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
```diff
@@ -1830,11 +1819,12 @@ public void storeMetric(GenericMetric metric) {
...
             metric.getMetricDomain(),
             metric.getMetricType(),
-            metric.getTs().toString(HOURLY_FORMATTER),
+            metric.getTs().toString(TIME_BUCKET_FORMATTER).substring(0, 11) + "0",
             metric.getHost(),
             metric.getMetricScope(),
             metric.getMetricName(),
-            metric.getTs().toDate(),
+            METRIC_TS_FORMATTER.parseDateTime(
+                metric.getTs().toString(METRIC_TS_FORMATTER).substring(0, 15) + "0").toDate(),
```
michaelsembwever: nit-pick: flooring the timestamp to the minute (or the ten-minute above) can be done in smarter ways using javax.time,  instead of text formatting.

src/server/src/main/java/io/cassandrareaper/storage/cassandra/Migration021.java
```diff
@@ -56,21 +56,23 @@ public static void migrate(Session session, String keyspace) {
...
-                      + "'unchecked_tombstone_compaction': 'true', "
-                      + "'compaction_window_size': '1', "
-                      + "'compaction_window_unit': 'HOURS'}");
+          if (session.getCluster().getMetadata().getKeyspace(keyspace).getTable(METRICS_V2_TABLE) != null) {
+            LOG.info("Altering {} to use TWCS...", METRICS_V2_TABLE);
+            session.execute(
+                    "ALTER TABLE " + METRICS_V2_TABLE + " WITH compaction = {'class': 'TimeWindowCompactionStrategy', "
+                        + "'unchecked_tombstone_compaction': 'true', "
+                        + "'compaction_window_size': '10', "
+                        + "'compaction_window_unit': 'MINUTES'}");
```
michaelsembwever: why not just remove this?
ie, why do you need to alter METRICS_V2_TABLE to TWCS, in a codebase where it is now METRICS_V3_TABLE ?

src/server/src/main/resources/db/cassandra/024_node_metrics_v2_partitioning.cql
```diff
@@ -0,0 +1,35 @@
+--
```
michaelsembwever: shouldn't this file `024_node_metrics_v2_partitioning.cql` be called instead `024_node_metrics_v3_partitioning.cql`

michaelsembwever:
there are a number of new files here who's start copyright year is wrong. can you fix that?
(might as well update the end copyright year in the files you touch as well at the same time)

src/server/src/main/java/io/cassandrareaper/jmx/ClusterFacade.java
michaelsembwever: can you put that in as a comment to the code. or better yet, put the number into constant that has a name that explains/hints what you just^ said.

michaelsembwever (approved):

src/server/src/main/java/io/cassandrareaper/jmx/ClusterFacade.java
adejanovski: sure thing

PR #870, opened 2020-03-25
check is enableDynamicSeedList is set before updating seeds

jsanda:
This PR is for #813.

In #813 we discussed adding special flag for enabling this behavior. I do not think an additional flag is necessary.

There is already similar logic in `ClusterResource` when adding/updating a cluster. I think that this behavior just makes the behavior consistent with respect to dynamic seed behavior.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
+++ b/src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
@@ -297,7 +297,10 @@ final class RepairRunner implements Runnable {
   private void updateClusterNodeList() throws ReaperException {
     Set<String> liveNodes  = ImmutableSet.copyOf(clusterFacade.getLiveNodes(cluster));
     Cluster cluster = context.storage.getCluster(clusterName);
-    if (!cluster.getSeedHosts().equals(liveNodes) && !liveNodes.isEmpty()) {
+    // Note that the seed hosts only get updated if enableDynamicSeedList is true. This is
+    // consistent with the logic in ClusterResource.findClusterWithSeedHost.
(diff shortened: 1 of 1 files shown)


src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
```diff
@@ -297,7 +297,8 @@ private void start() throws ReaperException, InterruptedException {
   private void updateClusterNodeList() throws ReaperException {
     Set<String> liveNodes  = ImmutableSet.copyOf(clusterFacade.getLiveNodes(cluster));
     Cluster cluster = context.storage.getCluster(clusterName);
-    if (!cluster.getSeedHosts().equals(liveNodes) && !liveNodes.isEmpty()) {
+    if (context.config.getEnableDynamicSeedList() && !cluster.getSeedHosts().equals(liveNodes)
```
michaelsembwever, line 302: can you add a comment referencing the `ClusterResource` logic that we're keeping in sync with here…?

puneetloya:
@jsanda thanks for the fix, I tried it out and worked for me. It was able to handle restarting of kubernetes pods

adejanovski (approved):
LGTM 👍

michaelsembwever (approved):
:shipit:

adejanovski:
Hi @jsanda,

is this PR mergeable or do you plan on pushing more changes?

jsanda:
@adejanovski I just rebased on master. Assuming CI passes, then this is ready for merge on my side.

adejanovski:
Postgres GHA checks seem to fail almost consistently lately, nothing to do with the changes in that PR.

puneetloya:
can I request for a release?

PR #876, opened 2020-04-02
Add integration tests featuring Elassandra

rzvoncek:
This adds a separate job that will run Reaper's tests with Elassandra for both Reaper's storage backend and nodes undergoing a repair.
* We install Elassandra from a release tarball.
* We use ecm (Elassandra flavour of ccm) to create a ccm cluster using the above tarball.
* We don't run 4.0 and sidecar tests against Elassandra.

I'm sorry for the white space changes.

### DIFF
--- a/.github/scripts/configure-ccm.sh
+++ b/.github/scripts/configure-ccm.sh
@@ -46,7 +46,7 @@ case "${TEST_TYPE}" in
         echo "ERROR: Environment variable TEST_TYPE is unspecified."
         exit 1
         ;;
-    "ccm"|"upgrade")
+    "ccm"|"upgrade"|"elassandra")
         mkdir -p ~/.local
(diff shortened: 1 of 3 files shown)


adejanovski (changes requested):
its-elassandra seems to be a distributed test (2 <= grim <= 4), which puts it in the `its-distributed` and `its-flapping` test phase.
Instead of `needs: [build]` it should have the following `needs: [its-ccm-local, its-pgsql, its-ccm-cass]` so it starts later. Also, `needs: [its-distributed, its-flapping, its-sidecar]` should become `needs: [its-distributed, its-flapping, its-sidecar, its-elassandra]`.
Aside from this, LGTM.

adejanovski (approved):
LGTM!

adejanovski:
@rzvoncek, GHA CI passed ✅ 

:shipit:

PR #851, opened 2020-02-15
Performance optimisations for `storage.getSegmentsWithStates(id, states…)`

smarsching:
On startup (or periodically in case of the Cassandra storage), the `RepairManager` aborts segments that do not have a leader or that belong to paused repair runs.

However, this code would only abort segments in the `RUNNING` state. This is a problem because segments might also be left in the `STARTED` state when a Cassandra Reaper instance is restarted. Such segments would block other segments from running (because when checking for busy hosts, `STARTED` segments are treated like `RUNNING` segments), but would never finish or be restarted, thus effectively blocking the repair process.

This PR fixes this problem by treating `RUNNING` and `STARTED` segments in the same way when aborting segments that do not have a leader or that belong to paused repair runs.

PR #883, opened 2020-04-09
Quick fix for stuck segments at STARTED state

adejanovski:
This PR will prevent repairs from getting stuck while we work on a cleaner fix in #851.


adejanovski:
replaced by #884

PR #884, opened 2020-04-09
Treat started segments as running when aborting.

adejanovski:
Quick fix for stuck STARTED segments while we work on a more performant version for the next minor/major release.


michaelsembwever (approved):

michaelsembwever:
Not too sure what the second commit is doing there though…?

adejanovski:
> Not too sure what the second commit is doing there though…?

I had to cherry pick a commit from `master` which fixes some issues in integration tests (otherwise they're not passing at all).

PR #885, opened 2020-04-09
Enable sidecar tests when using Elassandra

rzvoncek:


adejanovski (changes requested):
You also need to remove `--tags ~@sidecar` in `its-elassandra`:
https://github.com/thelastpickle/cassandra-reaper/pull/885/files#diff-60fe268f71ae7f8409aff40d0a3e51d8R333

rzvoncek:
Removed the other flag and amended the commit.

adejanovski (approved):

adejanovski:
:shipit:

PR #902, opened 2020-04-20
Fix file handling + posting forms in spreaper

rzvoncek:
* The file handling is a relict from the python2-3 upgrade.
* The form handling comes from dunno where, but using spreaper add-cluster with Reaper auth makes it looks like the login didn't happen when add-cluster goes.


michaelsembwever (approved):
:shipit:

michaelsembwever:
does this need to go to `2.0` branch as well ?

PR #911, opened 2020-05-04
Show pending compactions in /compactions + show them in UI. Fixes #877

### LINKED ISSUE #877: Feature Request: Show total number of compactions pending
Current UI shows the compactions running and their progress, it would be great if it could also show any that are pending - even the number pending similar to that shown in `nodetool compactionstats` would be helpful, but maybe showing them greyed out of with a "yet to start" message in the same table as the current ones running would be nice!

rzvoncek:
Before this changes, we only shown the active compactions, despite pending compactions being a prominent information inside Reaper (used for backoff):
Screenshot 2020-04-30 at 14 38 26

This change takes the pending compactions Reaper uses internally and exposes it in the /compactions endpoint and in the UI.
Screenshot 2020-04-30 at 15 29 57

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/jmx/ClusterFacade.java
+++ b/src/server/src/main/java/io/cassandrareaper/jmx/ClusterFacade.java
@@ -22,6 +22,7 @@ import io.cassandrareaper.ReaperApplicationConfiguration.DatacenterAvailability;
 import io.cassandrareaper.ReaperException;
 import io.cassandrareaper.core.Cluster;
 import io.cassandrareaper.core.Compaction;
+import io.cassandrareaper.core.CompactionStats;
 import io.cassandrareaper.core.DroppedMessages;
 import io.cassandrareaper.core.GenericMetric;
 import io.cassandrareaper.core.JmxStat;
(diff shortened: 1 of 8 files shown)


src/server/src/main/java/io/cassandrareaper/jmx/ClusterFacade.java
```diff
@@ -439,45 +439,49 @@ public String getLocalEndpoint(Node node) throws ReaperException, InterruptedExc
...
+      return listCompactionStatsDirect(node);
     } else {
       // We don't have access to the node through JMX, so we'll get data from the database
       LOG.info("Node {} in DC {} is not accessible through JMX", node.getHostname(), nodeDc);
 
       String compactionsJson = ((IDistributedStorage)context.storage)
           .listOperations(node.getClusterName(), OpType.OP_COMPACTION, node.getHostname());
 
-      return parseJson(compactionsJson, new TypeReference<List<Compaction>>(){});
+      return parseJson(compactionsJson, new TypeReference<CompactionStats>(){});
```
adejanovski: That could blow up during an upgrade. It's not backwards compatible with the previous storage format, so data in the database could come with the `Compaction` type instead of `CompactionStats`.
We'd need to gracefully ignore that, right?

src/server/src/main/java/io/cassandrareaper/jmx/CompactionProxy.java
```diff
@@ -93,4 +94,14 @@ public void forceCompaction(String keyspaceName, String... tableNames) {
     }
     return activeCompactions;
   }
+
+  public Integer getPendingCompactions() {
+    try {
+      return proxy.getPendingCompactions();
+    } catch (JMException e) {
+      LOG.warn(String.format("Could not fetch pending compactions from %s", proxy.getHost()));
```
adejanovski: Could you add the exception as argument to `LOG.warn` so we get the stack trace in the logs?
This should do it (you don't need `String.format()` either):

```
LOG.warn("Could not fetch pending compactions from {}", proxy.getHost(), e);
```

src/ui/app/jsx/active-compactions.jsx
```diff
@@ -48,9 +48,12 @@ const ActiveCompactions = CreateReactClass({
...
         complete: function(data) {
           try {
-            this.component.setState({activeCompactions: data.responseJSON});
+            this.component.setState({
+                pendingCompactions: data.responseJSON.pendingCompactions,
+                activeCompactions: data.responseJSON.activeCompactions
+            });
           } catch(error) {
-            this.component.setState({activeCompactions: []});
+            this.component.setState({pendingCompactions:-1, activeCompactions: []});
```
adejanovski, line 56: It might be good to add some tracing here: `console.error('Error occurred while retrieving compaction stats:', error);`
wdyt?

adejanovski (changes requested):
The only blocking(ish) problem I see is that exception might trigger upon upgrade. We could silently ignore this case and provide a better backwards compatibility.
You could also parse the old `Compaction` format and build a `CompactionStats` object out of it, but that could be overkill for a 2 minutes transition window :)

src/ui/app/jsx/active-compactions.jsx
rzvoncek: Sure thing, added the trace.

src/server/src/main/java/io/cassandrareaper/jmx/CompactionProxy.java
rzvoncek: 👍

src/server/src/main/java/io/cassandrareaper/jmx/ClusterFacade.java
rzvoncek: Yes, that's a good catch. I've just pushed a commit fixing this.

adejanovski (approved):
LGTM @rzvoncek

rzvoncek:
I've squashed the commits.

PR #919, opened 2020-05-12
Correct default keyspace value in documentation

simonfontana:
Update the default value for keyspace configuration in documentation.
Previous default was cassandra-reaper but it is now reaper_db.


adejanovski:
The change should be made under the `src/docs...` folder as `docs/doc` contains hugo generated pages.
I've ninja fixed this in commit [a8ba415](https://github.com/thelastpickle/cassandra-reaper/commit/a8ba415471ffd49940482b2d889b812c4712f178)

Issue #896, opened 2020-04-15
Performance profile Reaper 2.0+ (Cassandra storage)


michaelsembwever:
**Analyse Reaper trunk to identify performance bottlenecks and weaknesses. Focusing only on the Cassandra backend.**

This should involve:
- stress testing
- snapshots of CPU and GC, Memory, Garbage Collection, Threads and Classes profiling graphs
- Method, Object, and Thread, call graphs
- CQL call graphs
- flame graphs

The outcome of this ticket should be observations that validate and warrant further performance optimisation in the codebase. Subsequent issues/PRs will be opened for each identified bottleneck/weakness. For example https://github.com/thelastpickle/cassandra-reaper/pull/851

Only performance optimisations that are simple patches will be applied to `2.0`. Otherwise work will be done in `master`, slated for `2.1.0`

**Anyone is free to run performance tests and contribute additional findings to this issue.**

michaelsembwever:
The following profile runs were done on a 6 node single-dc ccm cluster holding ~20GB of inconsistent data and one reaper instance using `ALL` mode with C* storage. Reaper was built using SHA `d6b96e2`.

# Observations 

First is shown hotspots on various operations. Only methods under `io.cassandrareaper.` are shown.

Hotspots during a normal boot sequence:
![reaper-d6b96e2-normal-boot](https://user-images.githubusercontent.com/559444/82220852-7e497280-991f-11ea-9cf2-bb1df05cad0f.png)

Hotspots during the initial boot where schema creation occurs:
![reaper-d6b96e2-initialisation-boot](https://user-images.githubusercontent.com/559444/82221262-0e87b780-9920-11ea-81c1-c9edcf5cb131.png)

The difference between the two:
![reaper-d6b96e2-boot-diff](https://user-images.githubusercontent.com/559444/82221495-5dcde800-9920-11ea-8b71-4ddeed99f8eb.png)

Hotspots during creating a cluster:
![reaper-d6b96e2-add-cluster](https://user-images.githubusercontent.com/559444/82222154-2f9cd800-9921-11ea-9e0d-a3b3e6abc4de.png)

Hotspots during creating a repair run:
![reaper-d6b96e2-add-repair](https://user-images.githubusercontent.com/559444/82222272-58bd6880-9921-11ea-8265-354137f09465.png)

Hotspots during a repair of the 20GB using 122 segments:
[...]

PR #927, opened 2020-05-19
When there are no broadcasters and no known subscriptions, only check for updated events once per minute

michaelsembwever:
This will reduce table scans on diagnostic_event_subscription when reaper is distributed and diagnostics not otherwise used/subscribed to.
Based off findings from https://github.com/thelastpickle/cassandra-reaper/issues/896


michaelsembwever:
ready for review @adejanovski

adejanovski (approved):

Issue #568, opened 2018-10-04
reaper_db.repair_run_by_cluster schema ASC for DESC


thetdh:
While it would take some time for this to show, the fact that the schema for reaper_db.repair_run_by_cluster has "WITH CLUSTERING ORDER BY (id ASC)" instead of "WITH CLUSTERING ORDER BY (id DESC)" means that eventually, the 1000-result limit in CassandraStorage.getRepairRunsForCluster() will fail to include the most-recent runs.  (Ten keyspaces repaired at seven-day intervals would take nearly two years to show this problem.)

A possible transitional workaround would be to alter the query to specify "ORDER BY id DESC".

I wonder also if fetching an unlimited number of results, limited only by code fetching the results, will eventually cause a problem.  Last I looked, LIMIT couldn't be parameterized in a prepared statement, but it should be possible to chunk results with a fixed LIMIT if also using "WHERE id < ?".

adejanovski:
This was fixed in the latest versions.

PR #937, opened 2020-07-13
Limit the number of queries sent to C* backend

adejanovski:
Reaper checks the running and paused repair runs every 10 to 30 seconds to perform maintenance operations.
When the repair history starts growing, Reaper will scan them all to look for running/paused repairs because the repair_run_by_cluster table doesn't contain the repair run state.
This commit adds this column, which allows to filter runs early in the process instead of scanning them all in the repair_run table.
With hundreds or thousands of repair runs in history, this can end up adding a lot of unnecessary load on the Cassandra backend.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
+++ b/src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
@@ -42,6 +42,7 @@ import io.cassandrareaper.storage.cassandra.DateTimeCodec;
 import io.cassandrareaper.storage.cassandra.Migration016;
 import io.cassandrareaper.storage.cassandra.Migration021;
 import io.cassandrareaper.storage.cassandra.Migration024;
+import io.cassandrareaper.storage.cassandra.Migration025;
 
 import java.io.IOException;
 import java.math.BigInteger;
@@ -116,6 +117,7 @@ public final class CassandraStorage implements IStorage, IDistributedStorage {
 
   private static final int METRICS_PARTITIONING_TIME_MINS = 10;
   private static final int LEAD_DURATION = 600;
+  private static final int MAX_RETURNED_REPAIR_RUNS = 1000;
   /* Simple stmts */
   private static final String SELECT_CLUSTER = "SELECT * FROM cluster";
   private static final String SELECT_REPAIR_SCHEDULE = "SELECT * FROM repair_schedule_v1";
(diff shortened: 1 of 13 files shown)


src/server/src/main/resources/db/cassandra/025_lighten_load_repair_run_scans.cql
```diff
@@ -0,0 +1,27 @@
...
+--  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+--  See the License for the specific language governing permissions and
+--  limitations under the License.
+--
+-- Upgrade to handle diagnostic event subscriptions
+
+CREATE TABLE IF NOT EXISTS repair_run_by_cluster_v2 (
+    cluster_name text,
+    id timeuuid,
+    repair_run_state text,
```
michaelsembwever, line 21: why not just `alter table add column` on repair_run_by_cluster ?

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
```diff
@@ -744,7 +751,20 @@ public RepairRun addRepairRun(Builder repairRun, Collection<RepairSegment.Builde
 
   @Override
   public boolean updateRepairRun(RepairRun repairRun) {
-    session.execute(
+    return updateRepairRun(repairRun, Optional.of(true));
+  }
+
+  @Override
+  public boolean updateRepairRun(RepairRun repairRun, Optional<Boolean> updateRepairState) {
```
michaelsembwever, line 767: curious: how is this not still implicitly a read-before-write anti-pattern?

i.e. how will the write know for certain that it is not updating state?

if it is because this method is only ever called by the "leader elect" then an `assert` statement would be nice here to ensure this really is the case of all scenarios…

src/server/src/main/resources/db/cassandra/025_lighten_load_repair_run_scans.cql
adejanovski: Because I'm also changing the clustering order, to fix #568

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
adejanovski: I made it so that we either update status in both tables or in none of them.

michaelsembwever: Yes, and I see now the separate insert statement that doesn't touch the `state` column.

Very nice. It's only to look at the call hierarchy to `RepairRunner.updateLastEvent(..)` to see the positive impact this will have!

michaelsembwever (approved):

PR #938, opened 2020-07-18
Store replicas in the segments

### LINKED ISSUE #947: Change the segment/node locking system
Change the lock system to enable/bring:
- faster repairs (less contention)
- simpler code
- concurrent repairs on different keyspaces

adejanovski:
This will allow to get more visibility on which nodes are involved in segments, reduce the number of JMX calls to perform when starting a segment.
It is also a first step to build a new segment scheduler that will not rely on JMX metrics for limiting the number of repair per node but rather use backend storage with locks.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/core/RepairSegment.java
+++ b/src/server/src/main/java/io/cassandrareaper/core/RepairSegment.java
@@ -18,6 +18,7 @@
 package io.cassandrareaper.core;
 
 import java.math.BigInteger;
+import java.util.Map;
 import java.util.UUID;
 
 import javax.annotation.Nullable;
@@ -26,6 +27,7 @@ import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
 import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableMap;
 import org.joda.time.DateTime;
 
 @JsonDeserialize(builder = RepairSegment.Builder.class)
@@ -42,6 +44,7 @@ public final class RepairSegment {
   private final String coordinatorHost;
   private final DateTime startTime;
   private final DateTime endTime;
+  private final Map<String, String> replicas;
 
   private RepairSegment(Builder builder, @Nullable UUID id) {
     this.id = id;
@@ -53,6 +56,9 @@ public final class RepairSegment {
     this.coordinatorHost = builder.coordinatorHost;
     this.startTime = builder.startTime;
     this.endTime = builder.endTime;
+    this.replicas = builder.replicas != null
+        ? ImmutableMap.copyOf(builder.replicas)
+        : null;
   }
 
   public static Builder builder(Segment tokenRange, UUID repairUnitId) {
@@ -108,6 +114,10 @@ public final class RepairSegment {
     return endTime;
   }
 
+  public Map<String, String> getReplicas() {
+    return replicas;
+  }
+
   public Builder with() {
     return new Builder(this);
   }
@@ -149,6 +159,7 @@ public final class RepairSegment {
     private String coordinatorHost;
     private DateTime startTime;
     private DateTime endTime;
+    private Map<String, String> replicas;
 
     private Builder() {}
 
@@ -159,6 +170,7 @@ public final class RepairSegment {
       this.tokenRange = tokenRange;
       this.failCount = 0;
       this.state = State.NOT_STARTED;
+      this.replicas = tokenRange.getReplicas();
     }
 
     private Builder(RepairSegment original) {
@@ -171,6 +183,7 @@ public final class RepairSegment {
       coordinatorHost = original.coordinatorHost;
       startTime = original.startTime;
       endTime = original.endTime;
+      replicas = original.replicas;
     }
 
     public Builder withRunId(UUID runId) {
@@ -222,6 +235,11 @@ public final class RepairSegment {
 
     public Builder withId(@Nullable UUID segmentId) {
       this.id = segmentId;
+      return this;
+    }
+
+    public Builder withReplicas(Map<String, String> replicas) {
+      this.replicas = replicas;
       return this;
     }
 
--- a/src/server/src/main/java/io/cassandrareaper/core/Segment.java
+++ b/src/server/src/main/java/io/cassandrareaper/core/Segment.java
@@ -23,9 +23,11 @@ import java.math.BigInteger;
 import java.util.Arrays;
 import java.util.Comparator;
 import java.util.List;
+import java.util.Map;
 
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
 import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
+import com.google.common.collect.ImmutableMap;
 
 @JsonDeserialize(builder = Segment.Builder.class)
 public final class Segment {
@@ -36,6 +38,7 @@ public final class Segment {
 
   RingRange baseRange;
   List<RingRange> tokenRanges;
+  Map<String, String> replicas;
 
   private Segment(Builder builder) {
     this.tokenRanges = builder.tokenRanges;
@@ -43,6 +46,9 @@ public final class Segment {
     if (builder.baseRange != null) {
       this.baseRange = builder.baseRange;
     }
+    this.replicas = builder.replicas != null
+        ? ImmutableMap.copyOf(builder.replicas)
+        : null;
   }
 
   public RingRange getBaseRange() {
@@ -62,6 +68,10 @@ public final class Segment {
     return tokens;
   }
 
+  public Map<String, String> getReplicas() {
+    return this.replicas;
+  }
+
   public static Builder builder() {
     return new Builder();
   }
@@ -70,6 +80,7 @@ public final class Segment {
   public static final class Builder {
     private List<RingRange> tokenRanges;
     private RingRange baseRange;
+    private Map<String, String> replicas;
 
     private Builder() {}
 
@@ -85,6 +96,11 @@ public final class Segment {
 
     public Builder withBaseRange(RingRange baseRange) {
       this.baseRange = baseRange;
+      return this;
+    }
+
+    public Builder withReplicas(Map<String, String> replicas) {
+      this.replicas = replicas;
       return this;
     }
 
--- a/src/server/src/main/java/io/cassandrareaper/service/RepairRunService.java
+++ b/src/server/src/main/java/io/cassandrareaper/service/RepairRunService.java
@@ -26,9 +26,12 @@ import io.cassandrareaper.core.RepairUnit;
 import io.cassandrareaper.core.Segment;
 import io.cassandrareaper.core.Table;
 import io.cassandrareaper.jmx.ClusterFacade;
+import io.cassandrareaper.jmx.EndpointSnitchInfoProxy;
+import io.cassandrareaper.jmx.JmxProxy;
 
 import java.math.BigInteger;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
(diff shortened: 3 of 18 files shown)


src/server/src/main/java/io/cassandrareaper/core/RepairSegment.java
```diff
@@ -108,6 +111,11 @@ public DateTime getEndTime() {
     return endTime;
   }
 
+  @Nullable
```
michaelsembwever: what does the null state represent, versus an empty collection?

src/server/src/main/java/io/cassandrareaper/service/RepairRunService.java
```diff
@@ -185,7 +196,32 @@ public RepairRun registerRepairRun(
...
+                 .withBaseRange(segment.getBaseRange())
+                 .withTokenRanges(segment.getTokenRanges())
+                 .withReplicas(getDCsByNodeForRepairSegment(targetCluster, segment, repairUnit.getKeyspaceName()))
+                 .build());
+    }
+
+    return segmentsWithReplicas;
+  }
+
+  private Map<String, String> getDCsByNodeForRepairSegment(
```
michaelsembwever, line 214: my opinion is that this method should be in the `SegmentRunner` class.

and, the existing `getDCsByNodeForRepairSegment` there should be removed as no one is using it now.
https://github.com/thelastpickle/cassandra-reaper/blob/99ff8fc1743106cc3905a916ee44641d0a7ea6b0/src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java#L750

src/server/src/main/java/io/cassandrareaper/service/RepairRunService.java
```diff
@@ -185,7 +196,32 @@ public RepairRun registerRepairRun(
...
+      Cluster cluster,
+      Segment segment,
+      String keyspace) throws ReaperException {
+
+    JmxProxy jmxConnection = clusterFacade.connect(cluster);
+    // when hosts are coming up or going down, this method can throw an UndeclaredThrowableException
+    Collection<String> nodes = clusterFacade.tokenRangeToEndpoint(cluster, keyspace, segment);
```
michaelsembwever: How do we deal with failure here? Previously we'd just move on to another segment? Now the whole repair is jammed/failed because initialisation fails?

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
```diff
@@ -503,7 +503,7 @@ boolean canRepair(
 
     if (RepairSegment.State.NOT_STARTED == segment.getState()) {
       try {
-        Map<String, String> dcByNode = getDCsByNodeForRepairSegment(coordinator, cluster, segment, keyspace);
+        Map<String, String> dcByNode = segment.getReplicas();
```
michaelsembwever, line 506: nice that this is reducing the complexity (time and network calls) involved when determining whether to start a segment!

src/server/src/main/java/io/cassandrareaper/service/RepairRunService.java
adejanovski: It would fail indeed.
I can add a retry mechanism to at least give it a few attempts before failing the repair init.
wdyt?

src/server/src/main/java/io/cassandrareaper/service/RepairRunService.java
adejanovski: I agree that the existing one in the `SegmentRunner` class should be removed, but I don't see a reason why the one from `RepairRunService` should move to `SegmentRunner`. It's used in `RepairRunService.generateSegments()` only and we don't have a SegmentRunner object instanciated at that stage.

src/server/src/main/java/io/cassandrareaper/service/RepairRunService.java
adejanovski: Added a retry mechanism

adejanovski:
@michaelsembwever, I've addressed your comments and removed draft state.

src/server/src/main/java/io/cassandrareaper/service/RepairRunService.java
michaelsembwever: > I don't see a reason why the one from RepairRunService should move to SegmentRunner. It's used in RepairRunService.generateSegments() only and we don't have a SegmentRunner object instanciated at that stage.

Agreed.

src/server/src/main/java/io/cassandrareaper/core/RepairSegment.java
```diff
@@ -53,6 +55,7 @@ private RepairSegment(Builder builder, @Nullable UUID id) {
     this.coordinatorHost = builder.coordinatorHost;
     this.startTime = builder.startTime;
     this.endTime = builder.endTime;
+    this.replicas = builder.replicas;
```
michaelsembwever:
```suggestion
    this.replicas = ImmutableMap.copyOf(builder.replicas);
```

src/server/src/main/java/io/cassandrareaper/core/Segment.java
```diff
@@ -36,13 +37,15 @@
...
   List<RingRange> tokenRanges;
+  Map<String, String> replicas;
 
   private Segment(Builder builder) {
     this.tokenRanges = builder.tokenRanges;
     this.baseRange = builder.tokenRanges.get(0);
     if (builder.baseRange != null) {
       this.baseRange = builder.baseRange;
     }
+    this.replicas = builder.replicas;
```
michaelsembwever:
```suggestion
    this.replicas = ImmutableMap.copyOf(builder.replicas);
```

src/server/src/main/java/io/cassandrareaper/service/RepairRunService.java
```diff
@@ -292,6 +341,7 @@ static int computeGlobalSegmentCount(
                                 .withTokenRanges(Arrays.asList(range.getValue()))
                                 .build(),
                             repairUnit.getId())
+                        .withReplicas(Maps.newHashMap())
```
michaelsembwever:
```suggestion
                        .withReplicas(Collections.EMPTY_MAP)
```
?

src/server/src/main/java/io/cassandrareaper/service/RepairRunService.java
```diff
@@ -185,7 +196,45 @@ public RepairRun registerRepairRun(
...
+
+    return segmentsWithReplicas;
+  }
+
+  private Map<String, String> getDCsByNodeForRepairSegment(
+      Cluster cluster,
+      Segment segment,
+      String keyspace) throws ReaperException {
+
+    final int maxAttempts = 5;
```
michaelsembwever: why `5`? In what time-frame can the 5 subsequent calls fail within?

src/server/src/main/java/io/cassandrareaper/service/RepairRunService.java
michaelsembwever: Is this retry mechanism at repair-run initialisation enough? 

What happens when there are range movements during an ongoing repair-run? How do we detect that? How does it fail? Why (should) it fail? Do the docs need to be updated to describe these types of situations?

src/server/src/main/java/io/cassandrareaper/service/RepairRunService.java
adejanovski: It's totally arbitrary. What's your take on this?

src/server/src/main/java/io/cassandrareaper/service/RepairRunService.java
adejanovski: My take on this is that repair should fail if there's a range movement. Such operations will probably break the guarantees that repair brings.

adejanovski: I made it so that repair fails in case of topology change with an appropriate error message.
Anyway, adding/removing nodes would probably lead to the same result as token ranges that were computed for the segments aren't valid anymore.

wdyt @michaelsembwever ?

src/server/src/main/java/io/cassandrareaper/service/RepairRunService.java
michaelsembwever: IMHO it shouldn't be a hardcoded (nor magic) constant, and it should have a sleep in the loop.

Looking at the code I would think that when a node is down it is going to fail 5 times in a row reasonably quickly.

How long can/should this loop run for? How long should we be waiting for a node to become available again? What is this relative to, e.g. other settings in reaper?

src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
```diff
@@ -465,33 +465,65 @@ private boolean repairSegment(final int rangeIndex, final UUID segmentId, Segmen
...
+      } else if (
+          !Sets.difference(
+            segmentReplicas,
+            potentialCoordinators.stream().collect(Collectors.toSet())).isEmpty()) {
+        LOG.warn(
+            "Segment #{} is faulty, replica set changed since repair was started: [{}, {}]",
+            segmentId,
+            segment.getStartToken(), segment.getEndToken());
+        // This segment has a faulty token range. Abort the entire repair run.
+        synchronized (this) {
```
michaelsembwever, line 512: why `synchronized` ?

src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
```diff
@@ -465,33 +465,65 @@ private boolean repairSegment(final int rangeIndex, final UUID segmentId, Segmen
...
+          context.storage.updateRepairRun(
+              context.storage.getRepairRun(repairRunId).get()
+                  .with()
+                  .runState(RepairRun.RunState.ERROR)
+                  .lastEvent(String.format("Replica set changed for segment %s on range [%s, %s]",
+                      segmentId,
+                      segment.getStartToken(),
+                      segment.getEndToken()))
+                  .endTime(DateTime.now())
+                  .build(repairRunId));
```
michaelsembwever, line 522: This kills/aborts the whole repair run correct?  

What about other segments in it currently running? 

What about other repair runs registered but not yet started?

michaelsembwever:
Couple of questions posted.

Plus,   tests are needed for handling failures around token range movements, as that is a big part of the code change.

src/server/src/main/java/io/cassandrareaper/service/RepairRunService.java
michaelsembwever: 👍 

I agree that the whole repair run should fail (quickly and cleanly) on any token range movement.

src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
adejanovski: Repair run state updates are always protected by synchronized blocks. That's especially useful to deal with JMX notifications (if we decided to fail the repair run, we don't want a JMX notification to short circuit that in any way).

src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
adejanovski: yes, the whole repair will fail.
Running segments won't be impacted by this.

Other repair runs registered but not started will fail as well as soon as they hit a segment whose replica set is no longer valid.

src/server/src/main/java/io/cassandrareaper/service/RepairRunService.java
adejanovski: The idea here is rather to handle transient networking issues rather than handling nodes going down.
The way we implement this is that we'll go through `ClusterFacade.connect()` which will pick the fastest node answering in the cluster and then invoke `getRangeToEndpointMap()` on that node.
If we manage connecting to the node, and then it goes down before `getRangeToEndpointMap()`, then the loop will trigger the operation again, picking another node and trying to get the info from it. We won't end up trying to get the information from the same node throughout the retries.
That should be fairly resilient already I think and failing to create the repair run is ok as well (it'll get rescheduled by Reaper or the user will get some feedback and click the button again), so it's probably not worth waiting for too long there.

I can move the max number of attempts to a constant at the top, but I don't think we should make it configurable (probably no need to complexify the config file for this).
Looking at my above description of how things go, I think we could just perform a single retry to cover edge cases and fail fast instead of retrying too much.
wdyt?

adejanovski:
@michaelsembwever, 

I've pushed a few commits adding a unit test for topology changes and reducing the attempts on getting the list of replicas to a single one when generating segments.

Should be good unless you want to further discuss the retry stuff.

michaelsembwever (approved):
LGTM. just needs the commits squashed and commit msg written up…

PR #951, opened 2020-08-28
updates for SSL configuration

zencircle:


src/packaging/docker-services/generate-ssl-stores/docker-entrypoint.sh
```diff
@@ -35,7 +35,7 @@ set -ex
 
 CASSANDRA_KEYSTORE_PASSWORD=$1
 CASSANDRA_TRUSTSTORE_PASSWORD=$2
-
+WORKDIR=$PWD
```
ossarga, line 38: We can probably remove this line.

This is already defined as an environment variable in the [Dockerfile](https://github.com/thelastpickle/cassandra-reaper/blob/master/src/packaging/docker-services/generate-ssl-stores/Dockerfile#L18). It will already be set when the container is built. Hence, it will already be assigned a value when the _docker-entrypoint.sh_ executes. Only override this value if you really really needed to.

src/packaging/docker-services/generate-ssl-stores/docker-entrypoint.sh
```diff
@@ -49,14 +49,21 @@ GENERIC_TRUSTSTORE=${WORKDIR}/ssl-stores/generic-server-truststore.jks
 
 CA_CERT_CONFIG=${WORKDIR}/ca_cert.conf
 ROOT_CA_CERT=${WORKDIR}/ssl-stores/ca-cert
+CA_KEY=${WORKDIR}/ca-key
```
ossarga, line 52: We should remove this.

Originally when we wrote this container we debated about whether we would have a variable for the _ca-key_.

We decided against defining a variable for it because the key is only used as part of the process. It is never used as an external resource. For example:
* The keystore and truststore that are generated are used by Cassandra, tence they can be used as an external resource.
* The ca-certificate can be imported into the browser, hence it is an external resource.

The ca-key is never used externally, hence we kept it private and never bothered to make an environment variable for it.

The reason the container is set up like this is because, it is designed to serve purely as an example.

src/packaging/docker-services/generate-ssl-stores/docker-entrypoint.sh
```diff
@@ -49,14 +49,21 @@ GENERIC_TRUSTSTORE=${WORKDIR}/ssl-stores/generic-server-truststore.jks
...
+
+if [ -f "$CA_CERT_CONFIG" ] ; then
```
ossarga, line 54: Remove the error checking here.

Similar to my previous comment about the container being an example. The idea is to put the minimal amount of code to show the steps on how to produce the artefacts.

src/packaging/docker-services/generate-ssl-stores/docker-entrypoint.sh
```diff
@@ -49,14 +49,21 @@ GENERIC_TRUSTSTORE=${WORKDIR}/ssl-stores/generic-server-truststore.jks
...
+fi
 
 # Create the directory where the stores will go into if required.
 mkdir -p ${WORKDIR}/ssl-stores/
 
 # Check if there are any of the SSL stores exists and if so, prompt the user to delete them or exit
 set +x
 if [[ $(ls ${WORKDIR}/ssl-stores/*.jks | wc -l) -gt 0 ]]
-then
+ then
```
ossarga, line 66: Remove the added space.

src/packaging/docker-services/generate-ssl-stores/docker-entrypoint.sh
```diff
@@ -73,24 +80,33 @@ then
         esac
     done
 
-fi
+ fi
```
ossarga, line 83: Remove the added space.

src/packaging/docker-services/generate-ssl-stores/docker-entrypoint.sh
```diff
@@ -73,24 +80,33 @@ then
...
+else
+   openssl genrsa -out ${CA_KEY} 2048
+fi
+
+if [ -f "$ROOT_CA_CERT" ] ; then
+    echo "$ROOT_CA_CERT" exists, will not create
+else
+# Create the Root Certificate Authority (Root CA) from the Certificate Authority Configuration and verify contents.
+   openssl req -x509 -sha256 -new -nodes -key ca-key -out ${ROOT_CA_CERT} -config ${CA_CERT_CONFIG}
+fi
```
ossarga, line 102: Similar to my previous comment about the container being an example, remove the checks to see if the `ca-key` exists. Originally the `ca-key` was generated further down as part of the Root CA creation. Hence, there was never any need to check for it existing.

I do like the idea of showing the ca-key generation:
```
openssl genrsa -out ca-key 2048
```

I also like the Root CA generation command:
```
openssl req -x509 -sha256 -new -nodes -key ca-key -out ${ROOT_CA_CERT} -config ${CA_CERT_CONFIG}
```

I think it would be better to add these lines after we `echo "Generic Certificate Authority configuration"`. Maybe replace [line 92](https://github.com/thelastpickle/cassandra-reaper/pull/951/files#diff-ac79225b81efb4bc9081fc4c103f1cd7L92) with the above two lines. So the final blow should look like this:

```
# Create Certificate Authority Key, then create the Root Certificate Authority (Root CA) from the Certificate Authority Configuration and verify contents.
openssl genrsa -out ca-key 2048
openssl req -x509 -sha256 -new -nodes -key ca-key -out ${ROOT_CA_CERT} -config ${CA_CERT_CONFIG}
openssl x509 -in ${ROOT_CA_CERT} -text -noout
```

src/packaging/docker-services/generate-ssl-stores/docker-entrypoint.sh
```diff
@@ -73,24 +80,33 @@ then
...
+
 set +x
 echo
 echo "Generic Certificate Authority configuration"
 cat ${CA_CERT_CONFIG}
-echo
```
ossarga, line 88: Leave this new line in here. It makes the output on the terminal a bit easier to follow.

src/packaging/docker-services/generate-ssl-stores/docker-entrypoint.sh
```diff
@@ -73,24 +80,33 @@ then
...
 set -x
 
-# Create the Root Certificate Authority (Root CA) from the Certificate Authority Configuration and verify contents.
```
ossarga, line 91: Leave this line in OR if we are generating the `ca-key` as mentioned above, perhaps add that to the comment.

src/packaging/docker-services/generate-ssl-stores/docker-entrypoint.sh
```diff
@@ -73,24 +80,33 @@ then
...
-openssl req -config ${CA_CERT_CONFIG} -new -x509 -keyout ca-key -out ${ROOT_CA_CERT}
```
ossarga, line 92: Leave this line in OR replace with the line that generates the `ca-key` and the on which generates the Root CA.

src/packaging/docker-services/generate-ssl-stores/docker-entrypoint.sh
```diff
@@ -73,24 +80,33 @@ then
...
-openssl x509 -in ${ROOT_CA_CERT} -text -noout
```
ossarga, line 93: This is a good example of how to verify the certificate. Please leave it in.

ossarga (changes requested):
@zencircle Thank you for taking the time to submit this patch.

I have left a number of changes that we need to make before we merge this. The main reason is the docker containers for this project were set up to purely examples. There is no error checking deliberately as we wanted to show the basic steps to get something running. In a production environment, I would recommend adding error checking. However, for these examples, it is unnecessary. They were never designed to generate actual SSL stores for users.

zencircle:
$ cd src/packaging
$ docker-compose run generate-ssl-stores

Commands works

Issue #956, opened 2020-09-10
Improve sidecar mode for Kubernetes


jsanda:
# Improve sidecar mode for Kubernetes
The [docs](http://cassandra-reaper.io/docs/usage/sidecar_mode/) states that sidecar mode is _Kubernetes Friendly_. The purpose of this ticket is to highlight and address some problems with Reaper sidecar in Kubernetes.

I added initial support for Reaper sidecar mode in [Cass Operator](https://github.com/datastax/cass-operator), a Kubernetes operator for Cassandra. The initial integration does not work when Cassandra authentication is enabled (see this [issue](https://github.com/datastax/cass-operator/issues/211) for details). While working on the fix, I have encountered problems in the following areas:

* health probe
* startup initialization code
* schema updates

In Kubernetes, Cassandra is deployed in a StatefulSet. A StatefulSet consists of one or more pods. A pod can have multiple application containers. When Cass Operator deploys Reaper, each Cassandra pod includes a `cassandra` container as well as a `reaper` container. Containers can have readiness and liveness probes. Kubernetes uses the readiness probe to determine when a container is ready to start serving requests. It uses the liveness probe to determine whether the container needs to be restarted (see [here](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/) for more details about Kubernetes probes).

I am using Reaper's `healthcheck` endpoint for both the liveness and readiness probes. This endpoint is handled by the class `ReaperHealthCheck`. `ReaperHealthCheck` first checks that it is connected to the backend storage and then queries it.

Here is where things get interesting. The startup order of application containers in a pod is effectively non-deterministic. Cassandra can start before Reaper and vice versa. A pod is considered ready when all of its application containers report ready. For the Cassandra pod this means that both the `cassandra` and the `reaper` containers need to report ready in order for the pod to be considered ready. 

Cass Operator create roles in Cassandra, including one for Reaper, but they do not get created until the C* cluster is ready. The `reaper` container cannot reach the ready state because it does not have credentials with which to connect to Cassandra. And the pods cannot reach the ready state since the `reaper` containers cannot reach the ready state.

Initially I updated Cass Operator so that it doe the following:

* Deploy Cassandra without Reaper
* Wait for Cassandra to be ready
* Create the roles
* Update StatefulSet definition to include Reaper

[...]

PR #957, opened 2020-09-10
[WIP] Improved k8s sidecar support 

jsanda:
This is for #956

### DIFF
--- a/src/server/src/main/docker/entrypoint.sh
+++ b/src/server/src/main/docker/entrypoint.sh
@@ -71,6 +71,11 @@ if [ "$1" = 'register-clusters' ]; then
   echo "admin" > ~/.reaper/credentials
 
   if [ "SIDECAR" == "$REAPER_DATACENTER_AVAILABILITY" -a "true" == "$IS_K8S"]; then
+    # In this type of deployment we assume that Reaper is running in the same pod as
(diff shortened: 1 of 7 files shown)


michaelsembwever:
As mentioned in https://github.com/thelastpickle/cassandra-reaper/issues/956#issuecomment-694114459

I'm not in favour of such a large change, which essentially introduces a new modified sidecar mode just for kubernetes.

I would… 
a) implement Cass Operator to use either LOCAL/EACH, and
b) implement a dropwizard `db` command, instead of the `"SCHEMA_ONLY"` flag approach in #957 

For (b), references are 
- https://www.dropwizard.io/en/latest/getting-started.html#running-your-application
- https://github.com/dropwizard/dropwizard/blob/master/dropwizard-core/src/main/java/io/dropwizard/Application.java#L102-L105

This would implement the desired changes in a far simpler (correctly designed IMHO) manner.
You would not need the `"IS_K8S"` flag, and the addition k8_sidecar mode introduced. The `Initializer` would be replaced with the `db` dropwizard command (that would just call the storage initialisation methods already available in `ReaperApplication`). The healthcheck doesn't need to be undermined (it can remain true, i.e. _the service is actually initialised and running_.)

adejanovski:
Closing this PR as we took a different path for k8s.

PR #960, opened 2020-09-14
Fix race conditions when receiving COMPLETE/SUCCESS notifications

adejanovski:
if database updates are slow, COMPLETE notifications might miss SUCCESS flag as it is being set after database update.

Also, Astra won't allow CL.ONE, thus retries are now set to use the same CL

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
+++ b/src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
@@ -1824,7 +1824,7 @@ public final class CassandraStorage implements IStorage, IDistributedStorage {
       Preconditions.checkState(WriteType.CAS != type ||  ConsistencyLevel.SERIAL == cl);
 
       return null != stmt && !Objects.equals(Boolean.FALSE, stmt.isIdempotent())
-          ? WriteType.CAS == type ? RetryDecision.retry(ConsistencyLevel.ONE) : RetryDecision.retry(cl)
+          ? RetryDecision.retry(cl)
           : DefaultRetryPolicy.INSTANCE.onWriteTimeout(stmt, cl, type, required, received, retry);
     }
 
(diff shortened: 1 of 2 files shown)


src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
```diff
@@ -1824,7 +1824,7 @@ public RetryDecision onWriteTimeout(
       Preconditions.checkState(WriteType.CAS != type ||  ConsistencyLevel.SERIAL == cl);
 
       return null != stmt && !Objects.equals(Boolean.FALSE, stmt.isIdempotent())
-          ? WriteType.CAS == type ? RetryDecision.retry(ConsistencyLevel.ONE) : RetryDecision.retry(cl)
+          ? RetryDecision.retry(cl)
```
michaelsembwever, line 1827: How does Astra get around all the `CL.LOCAL_ONE` writes otherwise in this class ?

adejanovski: We have no other explicit `LOCAL_ONE` writes. I can't remember if the default CL is set at the driver or server level.
I've seen no sign of failed writes aside from failed retries, so I guess Astra sets the default CL server side:
Capture d’écran 2020-09-15 à 14 57 43

michaelsembwever: Weird. I was certain that the driver defaults to `CL.LOCAL_ONE`

michaelsembwever: `CassandraStorage` is designed so that the default is `CL.LOCAL_ONE`.

michaelsembwever: references
- https://docs.datastax.com/en/dsbulk/doc/dsbulk/reference/driverOptions.html#driverOptions__datastaxJavaDriverBasicRequestConsistency
- https://docs.datastax.com/en/developer/java-driver/4.9/manual/cloud/

michaelsembwever:
I'm good on this, so long all (flapping) tests on circleci pass.

adejanovski:
Thanks @michaelsembwever, flapping tests are green so I'll merge the PR.

PR #973, opened 2020-10-20
Expose max pending compactions configuration as environment variable

### LINKED ISSUE #983: Docker image: Configure max pending compactions configuration using environment variable
I need to modify the maxPendingCompactions configuration, but was unable to do so while using the docker image.

My current workaround is to building a custom cassandra-reaper docker image with a modified configuration file, but I would prefer to see this configuration option supported by default.

PR: https://github.com/thelastpickle/cassandra-reaper/pull/973

@adejanovski

jvv-trackunit:
I need to modify the `maxPendingCompactions` configuration, but was unable to do so while using the docker image. 

My current workaround is to building a custom cassandra-reaper docker image with a modified configuration file, but I would prefer to see this configuration option supported by default.


adejanovski (approved):
LGTM, nice addition 👍

PR #974, opened 2020-10-20
Added "millis since last repair"-metric for repair schedules.

### LINKED ISSUE #982: Add "millis since last repair"-metric for repair schedules.
### Use case
We have two schedules per cluster. One for incremental repairs and one for non-incremental repairs.

We wish to obtain metrics regarding the repairs triggered by each schedule, so that we can alert if one of the schedules has not run successfully for a while.

We use Prometheus for monitoring. The existing metrics provided details regarding ongoing repairs, but the metrics did not reveal which schedule they came from.

### Proposed enhancement
This PR adds a "millisSinceLastRepairForSchedule" for each repair schedule. For schedules that have never successfully run, the metric returns the number of milliseconds passed since epoch.

PR: https://github.com/thelastpickle/cassandra-reaper/pull/974

(This PR also implements a mapping of Dropwizard metrics to Prometheus metrics with nice labels. This means that users do not need to create custom relabeling rules as described in the documentation. I have added a MapperConfig for some of the metrics, but not all.)

adejanovski

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-262) by [Unito](https://www.unito.io)
┆Issue Number: K8SSAND-262

jvv-trackunit:
Fixes #982

### Use case
We have two schedules per cluster. One for incremental repairs and one for non-incremental repairs.

We wish to obtain metrics regarding the repairs triggered by each schedule, so that we can alert if one of the schedules has not run successfully for a while.

We use Prometheus for monitoring. The existing metrics provided details regarding ongoing repairs, but the metrics did not reveal which schedule they came from. 

### Proposed enhancement

This PR adds a "millisSinceLastRepairForSchedule" for each repair schedule. For schedules that have never successfully run, the metric returns the number of milliseconds passed since epoch.

This PR also implements a mapping of Dropwizard metrics to Prometheus metrics with nice labels. This means that users do not need to create custom relabeling rules as described in the documentation. I have added a MapperConfig for some of the metrics, but not all.


adejanovski:
@rzvoncek, you've been working on metrics and labels not so long ago. Would you mind reviewing this PR?

Thanks!

adejanovski:
@rzvoncek, gentle bump :)

docs/docs/metrics/prometheus/index.html
```diff
@@ -336,54 +336,6 @@ <h2 id="prometheus">Prometheus</h2>
 </code></pre></div><ul>
 <li>The <code>host.docker.internal</code> tells a Prometheus instance running inside a docker container to connect to the host&rsquo;s <code>8081</code> port where Raper runs from a JAR.</li>
 </ul>
-<h3 id="metric-relabelling">Metric Relabelling</h3>
```
rzvoncek-ds, line 339: I'm okay with removing documentation, especially when we're automating the actions elsewhere.

However, I'm afraid I'd need to ask you for a few things.

First, it'd be nice if you make the automatic re-labeling cover all metrics, or at least the ones mentioned in the document. The `repairProgress`is one (and perhaps only one) that I'm missing.

Second, we have a bit obscure setup with how the docs are generated. The HTML file here is just a render. Its source lives in [src/docs/content/docs/metrics/prometheus.md](https://github.com/thelastpickle/cassandra-reaper/blob/master/src/docs/content/docs/metrics/prometheus.md). Can you please change that too?

rzvoncek-ds (changes requested):
Hello and applogies for taking so long to have a look.

It's been a while since I've been looking into the metrics, but the changes you propose look good and make sense. 

There's just a few tweaks that I think would help.

src/server/src/main/java/io/cassandrareaper/service/RepairScheduleService.java
```diff
@@ -94,6 +100,71 @@ public RepairSchedule storeNewRepairSchedule(
...
+
+      Long lastRepairDate = schedule.orElseThrow(() -> new IllegalArgumentException("Repair schedule not found"))
+          .getRunHistory().stream()
+          .map(uuid -> context.storage.getRepairRun(uuid))
+          .filter(Optional::isPresent)
+          .map(Optional::get)
+          .map(RepairRun::getEndTime)
+          .filter(Objects::nonNull)
+          .max(DateTime::compareTo)
+          .map(dateTime -> DateTime.now().getMillis() - dateTime.getMillis())
```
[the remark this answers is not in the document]
jvv-trackunit: You are right - I used the same reasoning:

When a schedule has never been successfully repaired, no data about a repair is present, so what value should the metric have? 
The value 0 would indicate that it was last repaired 0 milliseconds ago, which is not the case. 
The value epoch, would indicate that a repair happened at the beginning of time, which is also not the case, but fits alert rules better.

The (Prometheus) alert expression I use looks like this:
```
io_cassandrareaper_service_RepairScheduleService_millisSinceLastRepairForSchedule > 1000 * 60 * 60 * 24 * 2
```
It triggers if there has not been a repair in the last 2 days. If there has never been a repair, it would trigger as well.

docs/docs/metrics/prometheus/index.html
jvv-trackunit: Sure. I'll try to find time for it this week.

adejanovski:
Hi @rzvoncek-ds,

what's the follow up on this PR? Anything we should do here before we can merge?

adejanovski:
@rzvoncek @van-vliet @ujo-trackunit,

any chance we could move forward with this PR?
What would be nice is to rebase this work on top of 2.3 and point the PR to the 2.3 branch.
Also, it seems like there were more metrics that needed the same changes based on the latest comment. @van-vliet, let us know if you don't have the bandwidth to handle this and we'll see if we can pick it up ourselves.

Thanks

adejanovski:
We'd need to rebase this work on top of `master` and try to get it merged.

jvv-trackunit:
Hi @adejanovski 

I completly forgot about this one. I'll attempt the merge and address the original feedback.

adejanovski:
thanks @van-vliet, that would be much appreciated 🙏

jvv-trackunit:
Converted to draft. Still working on it.

jvv-trackunit:
PR is ready for review.

The changes to metric names/labels are contained in the following file: PrometheusMetricsConfiguration.java
Here are some examples of the updated metric names/labels:
```
io_cassandrareaper_service_SegmentRunner_runRepair_count{cluster="tm2clusterdev",keyspace="tm2",host="10x0x0x2",} 1.0
io_cassandrareaper_service_SegmentRunner_postpone{cluster="tm2clusterdev",keyspace="tm2",host="10x0x0x2",} 2.0
io_cassandrareaper_service_RepairRunner_segmentsTotal{cluster="tm2clusterdev",keyspace="tm2",repairid="0143cdc0e88411ebafc94758af2aa92d",} 5.0
io_cassandrareaper_service_RepairScheduleService_millisSinceLastRepairForSchedule{cluster="tm2clusterdev",keyspace="tm2",scheduleid="ea24f870996911ec8b8d73d062fcc75d",} 2.2572745E7

```

src/server/src/main/java/io/cassandrareaper/service/RepairScheduleService.java
```diff
@@ -117,6 +123,71 @@ public RepairSchedule storeNewRepairSchedule(
...
+    if (context.metricRegistry.getMetrics().containsKey(metricName)) {
+      context.metricRegistry.remove(metricName);
+    }
+  }
+
+  private Gauge<Long> getMillisSinceLastRepairForSchedule(UUID repairSchedule) {
+    return () -> {
+      Optional<RepairSchedule> schedule = context.storage.getRepairSchedule(repairSchedule);
+
+      Long lastRepairDate = schedule.orElseThrow(() -> new IllegalArgumentException("Repair schedule not found"))
```
jvv-trackunit: The elseThrow-clause here is causing a lot of noise in the logs when cluster are deleted with the force option.

See an example [here](https://pipelines.actions.githubusercontent.com/serviceHosts/ae4c892b-b01f-48c5-a10b-ad35cb1983f3/_apis/pipelines/1/runs/1393/signedlogcontent/273?urlExpires=2022-04-21T09%3A31%3A14.6595577Z&urlSigningMethod=HMACV1&urlSignature=Ft%2FCZEa11qtvFl588tT%2ByRVlJGy8i3xF8wTo8Grb090%3D) (search for `java.lang.IllegalArgumentException: Repair schedule not found`):

In that case, the metric name is not unregistered properly and the system fails to retrieve the repair schedule.

I see 3 ways to fix this:
1. Make the delete cluster endpoint check for existing schedules and let it attempt to delete them (which will trigger the metric to unregister). This is a change of behaviour and I am unsure if this has other consequences.
2. Change this method so it returns a default value in case the repair schedule cannot be retrieved. The default could be `DateTime.now().getMillis()`  which translates to "the scheduled repair was last run just now. The reason for choosing this default value is the same as discussed in [this thread](https://github.com/thelastpickle/cassandra-reaper/pull/974/files#r566073632). This solution would ensure the error is never thrown, but the reaper would keep exposing a metric for a schedule that is not present anymore.
3. Change this method so it unregisters the metricname if the schedule is not found. The danger of this approach is that it can potentially unregister the metric if the getRepairSchedule method fails due to network issues but later recovers. I'm also unsure if this will trigger a ConcurrentModificationException in the underlying metric collection. Maybe the DropWizard metrics registry has safeguards against that - otherwise the unregister would have to be delayed until all metrics have been scraped.

I prefer solution 1, but I would like to hear your thoughts.

adejanovski:
Thanks @van-vliet,

I'll pick up the review and answer your questions ASAP.

src/server/src/main/java/io/cassandrareaper/service/RepairScheduleService.java
adejanovski: I'm ok with 1, keeping in sight that we will still delete the cluster even if we fail deleting the repair schedules.
This can be done as a separate follow up ticket if you want, which would allow to move forward with this PR.

jvv-trackunit: I have implemented suggestion 1. 
I haven't added a try/catch around the repair schedule deletion, since I don't expect that to be necessary.

Should the storage backend fail to delete the repair schedule, I imagine the entire storage backend is in a bad state and deleting the cluster will also fail.

Let me know if the current implementation is acceptable or not.

src/server/src/main/java/io/cassandrareaper/service/RepairScheduleService.java
```diff
@@ -117,6 +123,71 @@ public RepairSchedule storeNewRepairSchedule(
...
+          .getRunHistory().stream()
```
adejanovski: `runHistory` is no longer updated in the repair schedule table because it was unused and could create very large collections over time.
You can use `lastRun` instead which gives the id of the last repair run for this schedule.

src/server/src/main/java/io/cassandrareaper/service/RepairScheduleService.java
```diff
@@ -117,6 +123,71 @@ public RepairSchedule storeNewRepairSchedule(
...
+          .map(uuid -> context.storage.getRepairRun(uuid))
+          .filter(Optional::isPresent)
+          .map(Optional::get)
+          .map(RepairRun::getEndTime)
+          .filter(Objects::nonNull)
+          .max(DateTime::compareTo)
+          .map(dateTime -> DateTime.now().getMillis() - dateTime.getMillis())
+          .orElse(DateTime.now().getMillis()); // Return epoch if no repairs from this schedule ever completed
+      return lastRepairDate;
```
adejanovski: Shouldn't `lastRepairDate` be renamed to `millisSinceLastRepair` for example?

src/server/src/main/java/io/cassandrareaper/service/RepairScheduleService.java
adejanovski: LGTM

adejanovski (changes requested):
Thanks for these new additions @van-vliet.
I've requested a couple changes that should be easy to make.
It would be great if we could have some tests added, may they be unit or integration to make sure we secure those changes in the future.

src/server/src/main/java/io/cassandrareaper/service/RepairScheduleService.java
jvv-trackunit: Yes, I'll rename the variable.

src/server/src/main/java/io/cassandrareaper/service/RepairScheduleService.java
jvv-trackunit: Is there a reason why the runHistory is still present in the code if it is no longer kept up to date?

Looking at the code, it looks like the lastRun is updated immediately upon creating a new repair run, before it has completed. I suspect this will cause the metric to get assigned the default value when a new repair for the schedule starts and until the repair is completed. 

I'll write a couple of tests to confirm this (as mentioned in the other comments).

adejanovski: > Is there a reason why the runHistory is still present in the code if it is no longer kept up to date?

Not that I know of, we should've removed it.
We should then change the code so that `lastRun` gets updated at the end of the repair instead of at its beginning. This way it won't mess the metrics.

jvv-trackunit:
Looking at the current test setup, I think the best place to add these tests is as part of the cucumber test suite. This allows reusing the existing setup to add/removes schedules and verify that the metrics are correct.

I'm having a hard time running the cucumber tests locally. Ideally, I would be able to run them directly from my IDE (Intellij), but executing them by running a maven command is also acceptable.

I'm trying to start the cucumber tests by running src/test/java/io/cassandrareaper/acceptance/ReaperCassandraIT.java, but it seems to depend on cassandra clusters that have been setup by the CI pipeline.

@adejanovski How do you run the cucumber tests locally?

adejanovski:
@van-vliet, sounds good 👍 

Here's a script that will create a `ccm` cluster which is fit for Reaper's integration tests: https://gist.github.com/adejanovski/00d1885eb591bbb53a2034f501d36de9

Invoke it like this: `./create-reaper-test-cluster.sh --version=4.0.3`
You can then run the IT with maven:

```
mvn -B surefire:test -DsurefireArgLine="-Xmx384m" -Dtest=ReaperCassandraIT
```

If you create your cluster with a version of Cassandra that's < 4.0.0, you'll need to skip the tests that are specific to Cassandra 4.0+:

```
mvn -B surefire:test -DsurefireArgLine="-Xmx384m" -Dtest=ReaperCassandraIT -Dcucumber.options="--tags ~@cassandra_4_0_onwards"
```

src/server/src/main/java/io/cassandrareaper/service/RepairScheduleService.java
jvv-trackunit: I have rewritten the code to use lastRun and changed when lastRun is updated.

adejanovski: Awesome! I'll review the PR today.

adejanovski:
@van-vliet, changes look great!

Thanks for creating the additional integration tests. The last thing we need to do is have them run in CI.
Could you add the following command after line 67 in `.github/workflows/ci.yaml`?

```
mvn -B org.jacoco:jacoco-maven-plugin:${{ env.JACOCO_VERSION }}:prepare-agent surefire:test org.jacoco:jacoco-maven-plugin:${{ env.JACOCO_VERSION }}:report -Dtest=ReaperMetricsIT
```

That'll do it.
Other than that, it's all looking good for a merge.

van-vliet:
> Thanks for creating the additional integration tests. The last thing we need to do is have them run in CI. Could you add the following command after line 67 in `.github/workflows/ci.yaml`?
> 
> ```
> mvn -B org.jacoco:jacoco-maven-plugin:${{ env.JACOCO_VERSION }}:prepare-agent surefire:test org.jacoco:jacoco-maven-plugin:${{ env.JACOCO_VERSION }}:report -Dtest=ReaperMetricsIT
> ```

The ReaperMetricsIT class requires a running Cassandra cluster. The `ci.yaml` does not set that up. 
I have moved the test invocation to `.github/scripts/run-tests.sh` instead to make it run.

jvv-trackunit:
> Thanks for creating the additional integration tests. The last thing we need to do is have them run in CI. Could you add the following command after line 67 in `.github/workflows/ci.yaml`?
> 
> ```
> mvn -B org.jacoco:jacoco-maven-plugin:${{ env.JACOCO_VERSION }}:prepare-agent surefire:test org.jacoco:jacoco-maven-plugin:${{ env.JACOCO_VERSION }}:report -Dtest=ReaperMetricsIT
> ```

The ReaperMetricsIT class requires a running Cassandra cluster. The `ci.yaml` does not set that up. 
Instead, I have moved the test invocation to `.github/scripts/run-tests.sh` to make it run.

Let me know if any other changes are needed @adejanovski .

jvv-trackunit:
> > Thanks for creating the additional integration tests. The last thing we need to do is have them run in CI. Could you add the following command after line 67 in `.github/workflows/ci.yaml`?
> > ```
> > mvn -B org.jacoco:jacoco-maven-plugin:${{ env.JACOCO_VERSION }}:prepare-agent surefire:test org.jacoco:jacoco-maven-plugin:${{ env.JACOCO_VERSION }}:report -Dtest=ReaperMetricsIT
> > ```
> 
> The ReaperMetricsIT class requires a running Cassandra cluster. The `ci.yaml` does not set that up. Instead, I have moved the test invocation to `.github/scripts/run-tests.sh` to make it run.
> 
> Let me know if any other changes are needed @adejanovski .

Any news on this @adejanovski ?

adejanovski:
Hi @van-vliet , yes sorry, I was out for the past two weeks and just got back to your PR today.
It seems like memory backends IT aren't compatible with the ReaperMetricsIT suite.
My recommendation here would be to add tags to the ReaperMetricsIT scenario outlines such as `@metrics` and exclude them from the memory IT tests by adding the following to cucumber options for memory tests under the `its-ccm-local` job in `ci.yaml`:

```
        include:
          # don't run against the following C* versions when using cassandra storage type
          - cassandra-version: "binary:1.2.19"
            cucumber-options: '--tags ~@cassandra_2_1_onwards --tags ~@cassandra_4_0_onwards --tags ~@cassandra_3_11_onwards --tags ~@metrics'
            experimental: false
            jdk-version: '8'
          - cassandra-version: "binary:2.0.17"
            cucumber-options: '--tags ~@cassandra_2_1_onwards --tags ~@cassandra_4_0_onwards --tags ~@cassandra_3_11_onwards --tags ~@metrics'
            experimental: false
            jdk-version: '8'
          - cassandra-version: 'binary:2.1.22'
            cucumber-options: '--tags ~@cassandra_4_0_onwards --tags ~@cassandra_3_11_onwards --tags ~@metrics'
            experimental: false
            jdk-version: '8'
          - cassandra-version: 'binary:2.2.19'
            cucumber-options: '--tags ~@cassandra_4_0_onwards --tags ~@cassandra_3_11_onwards --tags ~@metrics'
            experimental: false
            jdk-version: '8'
          - cassandra-version: 'binary:3.0.25'
            cucumber-options: '--tags ~@cassandra_4_0_onwards --tags ~@cassandra_3_11_onwards --tags ~@metrics'
            experimental: false
            jdk-version: '8'
          - cassandra-version: 'binary:3.11.11'
            cucumber-options: '--tags ~@cassandra_4_0_onwards --tags ~@metrics'
            experimental: false
            jdk-version: '8'
          - cassandra-version: 'binary:4.0.1'
            cucumber-options: '--tags ~@metrics'
            experimental: false
            jdk-version: '11'
          - cassandra-version: 'github:apache/trunk'
            cucumber-options: '--tags ~@metrics'
            experimental: true
            jdk-version: '11'
```

The docker IT failure is unrelated to the changes from this PR.

Once we get a green run (except Docker IT), we'll be good to merge!

adejanovski:
Something weird is happening with the integration tests 🤔 I'll investigate to understand why they aren't all running.

.github/workflows/ci.yaml
```diff
@@ -129,33 +129,33 @@ jobs:
...
+            cucumber-options: '--tags ~@cassandra_4_0_onwards --tags ~@cassandra_3_11_onwards --tags ~@metrics'
             experimental: false
             jdk-version: '8'
           - cassandra-version: 'binary:3.11.11'
-            cucumber-options: '--tags ~@cassandra_4_0_onwards'
+            cucumber-options: '--tags ~@cassandra_4_0_onwards --tags ~@metrics'
             experimental: false
             jdk-version: '8'
-          - cassandra-version: 'binary:4.0.1'
+          - cassandra-version: 'binary:4.0.1 --tags ~@metrics'
```
adejanovski: I made a mistake in my example.
The `--tags`  argument should be in the `cucumber-options` matrix variable, not the `cassandra-version` one.
Hence, it should be the following:

```
          - cassandra-version: 'binary:4.0.1'
            cucumber-options: '--tags ~@metrics'
            experimental: false
            jdk-version: '11'
          - cassandra-version: 'github:apache/trunk'
            cucumber-options: '--tags ~@metrics'
            experimental: true
            jdk-version: '11'
```

adejanovski (changes requested):

.github/workflows/ci.yaml
adejanovski: @van-vliet, you can cherry pick [this commit](https://github.com/thelastpickle/cassandra-reaper/pull/1199/commits/3db2aa794e94c8956da9f3bb42a9b70837606176) if you want to apply the fix.

adejanovski: @van-vliet, I finally understand what's wrong. Trying to run multiple features in the same `mvn` call doesn't work because some resources don't seem to be cleaned up correctly.

Here's a new commit you can cherry pick which separates features in distinct calls and removes the metrics cucumber tag which isn't necessary after all: https://github.com/thelastpickle/cassandra-reaper/pull/1199/commits/b9a384c5e4093df81262b9403a4a6e81b085e73d

jvv-trackunit: Thanks @adejanovski for figuring out the CI parts. I've just pushed the cherrypicked commit.

adejanovski (approved):
We're good to go, at last!
Thanks for going through with this @van-vliet, greatly appreciated 🙏

jvv-trackunit:
Thanks for all the help you provided along the way @adejanovski 
I'm glad we finally got this merged. Looking forward to the next official release.

PR #977, opened 2020-10-21
Fix Astra migrations with LWTs not being properly allowed.

adejanovski:
Astra rejects LWTs when the statements don't have a (non serial) CL set to ONE or LOCAL_ONE.
This PR enforces using LOCAL_QUORUM as default CL when using Astra as storage backend.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
+++ b/src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
@@ -524,12 +524,15 @@ public final class CassandraStorage implements IStorage, IDistributedStorage {
     takeLeadPrepStmt = session
         .prepare(
             "INSERT INTO leader(leader_id, reaper_instance_id, reaper_instance_host, last_heartbeat)"
-                + "VALUES(?, ?, ?, " + timeUdf + "(now())) IF NOT EXISTS USING TTL ?");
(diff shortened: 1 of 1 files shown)


src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
```diff
@@ -524,12 +524,15 @@ private void prepareLeaderElectionStatements(final String timeUdf) {
     takeLeadPrepStmt = session
         .prepare(
             "INSERT INTO leader(leader_id, reaper_instance_id, reaper_instance_host, last_heartbeat)"
-                + "VALUES(?, ?, ?, " + timeUdf + "(now())) IF NOT EXISTS USING TTL ?");
+                + "VALUES(?, ?, ?, " + timeUdf + "(now())) IF NOT EXISTS USING TTL ?")
+        .setConsistencyLevel(ConsistencyLevel.QUORUM);
```
michaelsembwever, line 528: is any of the leader election usages cross-dc (and need to be QUORUM over LOCAL_QUORUM) ?

adejanovski: Yes, sadly we need cross DC consistency on leader election otherwise two Reaper instances in different DCs might be able to take the lead on the same segment.

michaelsembwever (approved):

Issue #998, opened 2020-12-07
Cannot configure authentication due to "Interrupted System Call" error


fatihafizoglu:
While using version 2.1.2 of cassandar reaper in kubernetes cluster, I am getting the below error very frequently:

```/usr/local/bin/configure-webui-authentication.sh: line 44: can't `create` /etc/shiro.ini: Interrupted system call```

Because of this problem, it is required to delete the pod and wait for another reaper pod to be scheduled each time and expect that error won't repeat.

adejanovski:
I guess we could do the same thing as with the `cassandra-reaper.yml` [here](https://github.com/thelastpickle/cassandra-reaper/blob/master/src/server/src/main/docker/Dockerfile#L98-L99).

A workaround would be to mount the `/etc/cassandra-reaper/shiro` to a local directory containing a valid `shiro.ini` file and point your REAPER_SHIRO_INI variable to `/etc/cassandra-reaper/shiro.ini`.

fatihafizoglu:
In the workaround solution, `REAPER_AUTH_USER` and `REAPER_AUTH_PASSWORD` variables won't work then, right? I mean the content of the shiro.ini file should be prepared and mount to that directory without using those variables?

adejanovski:
yes, you got it right.

fatihafizoglu:
Do you know when a solution like [here](https://github.com/thelastpickle/cassandra-reaper/blob/master/src/server/src/main/docker/Dockerfile#L98-L99) will be available? I am asking because personally, I would choose using env variables instead of using the workaround. If it will be available soon, it would be worth waiting :)

adejanovski:
if you feel like it, you could push a PR to the 2.1 branch, adding a touch for the shiro.ini file in the Dockerfile.
We have another pressing issue to deal with in the Docker container, which we could possibly address and then ship 2.1.3 containing both fixes.

PR #1000, opened 2020-12-08
Fix interrupted system call error for shiro initialization

fatihafizoglu:
PR for the problem discussed in #998


adejanovski (approved):
Thanks @fatihafizoglu!

Successfully tested ✅

Issue #987, opened 2020-11-07
Upgrading to 2.1 gives me "Storage is not ready yet, trying again to connect shortly..."


nicholasamorim:
I'm using Docker version.

When I upgrade to tag `ad53ff3` (because I need this [fix](https://github.com/thelastpickle/cassandra-reaper/pull/972)), the message "Storage is not ready yet, trying again to connect shortly..." keeps appearing on the logs forever and the connection never really happens.

```
ERROR  [2020-11-07 13:11:16,885] [main] i.c.ReaperApplication - Storage is not ready yet, trying again to connect shortly...
java.lang.NullPointerException: null
	at systems.composable.dropwizard.cassandra.CassandraFactory.build(CassandraFactory.java:469)
	at systems.composable.dropwizard.cassandra.CassandraFactory.build(CassandraFactory.java:447)
	at io.cassandrareaper.storage.CassandraStorage.<init>(CassandraStorage.java:225)
	at io.cassandrareaper.ReaperApplication.initializeStorage(ReaperApplication.java:477)
	at io.cassandrareaper.ReaperApplication.tryInitializeStorage(ReaperApplication.java:300)
	at io.cassandrareaper.ReaperApplication.run(ReaperApplication.java:179)
	at io.cassandrareaper.ReaperApplication.run(ReaperApplication.java:98)
	at io.dropwizard.cli.EnvironmentCommand.run(EnvironmentCommand.java:43)
	at io.dropwizard.cli.ConfiguredCommand.run(ConfiguredCommand.java:87)
	at io.dropwizard.cli.Cli.run(Cli.java:78)
	at io.dropwizard.Application.run(Application.java:93)
	at io.cassandrareaper.ReaperApplication.main(ReaperApplication.java:117)
```

At first I thought it could be a migration issue so I dropped and recreated the `reaper_db` keyspace. The issue is still there.

These are my envvars:

```yaml
REAPER_CASS_CLUSTER_NAME: "myclustername"
REAPER_CASS_CONTACT_POINTS: "[bunch,of,ips]"
REAPER_CASS_LOCAL_DC: "datacenter1"
REAPER_STORAGE_TYPE: cassandra
REAPER_HANGING_REPAIR_TIMEOUT_MINS: "120"
REAPER_BLACKLIST_TWCS_TABLES: "0"
REAPER_REPAIR_MANAGER_SCHEDULING_INTERVAL_SECONDS: "10"
[...]

PR #1001, opened 2020-12-18
Fix bower build, docker build and one k8ssandra bug

burmanm:
* Fixing #987 by installing bash to the docker image and also updated the Alpine image to a maintained version.
* Fix the bower.json build by taking care of the dependency issue (conflict with the font-awesome versions)
* k8ssandra could in a certain edge case (cluster goes down, user deletes it from reaper and then restarts the cluster) call the addCluster again for an existing cluster and that could cause the seeds to get corrupted inside Reaper. Only update the seeds in re-add of existing cluster in addCluster if dynamicSeeds is enabled.


adejanovski (approved):
Great much needed fixes! Thanks @burmanm 👍

adejanovski:
@burmanm, could we point the PR to the `2.1` branch instead of `master`?
These are all bug fixes which will be released quicker in 2.1 rather than waiting for our next minor.

Thanks!

michaelsembwever:
I can't make heads or tails of this PR, with its 16 commits.
Could this please be cleaned up so it's possible to review…

burmanm:
It's the change of base with those merge commits that messes it up (the original branch is against master). The real commits are the last four.

adejanovski:
Merged as part of #1002

PR #1002, opened 2020-12-18
PR #1001 against 2.1

burmanm:


src/server/src/main/docker/configure-persistence.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
michaelsembwever, line 1: It's unclear why this change is needed.

Does it fix something? Or is it incase someone in the future would like to use bash?

If the latter, then it should be removed from the PR.

If there's a use-case, outside of the Reaper codebase, that needs it could you elaborate-on/demonstrate it a bit please.

burmanm: The configure-persistence.sh uses a bashism and doesn't work with some `sh` implementations, such as the Alpine one used by the Dockerfile. #987 has the error message.

adejanovski (approved):

adejanovski:
Thanks again @burmanm!
Merging the PR.

PR #1016, opened 2021-02-12
Add information about Reaper operator to the "Sidecar mode" 

Miles-Garnsey:
Feel free to reject if this one isn't helpful, just thought it might be good to mention the work you folks are doing on the Reaper operator.

I've also tried to clarify how the UI works with sidecar mode - @adejanovski do I have this right? Is it the case that all the sidecar Reaper instances in a cluster are also running a UI instance, and any one of them can control all the others? 

I wasn't entirely clear on this when I glanced over the docs, so hoping I've understood correctly.


Miles-Garnsey:
Re-examining this, I don't think my commentary on the k8ssandra Reaper operator is in the right place, because the new operator doesn't actually run in sidecar mode. I will go away and think about how to incorporate more information on this stuff.

PR #1026, opened 2021-02-25
Fix 404

johnsmartco:
authentication link


adejanovski (approved):
Good catch @johnsmartco 👍

PR #1027, opened 2021-02-26
Limit repair runners per cluster instead of globally

adejanovski:
2.2 introduced parallel repairs with a limit on the number of repair runners to manage overall repair pressure.
This was done globally instead of per cluster, which will be very limiting for Reapers managing several clusters.

This PR applies the limit per cluster.


jsanda (commented):
@adejanovski I have a couple questions:

* Where is the max parallel repairs stored?
* Is it a global setting, or can it be configured per cluster?

adejanovski:
> @adejanovski I have a couple questions:
> 
> * Where is the max parallel repairs stored?

In the configuration yaml file.

> * Is it a global setting, or can it be configured per cluster?

It's a global setting that cannot be configured per cluster. All clusters will get the same limit. We can revisit this and make it a per cluster setting in the future if the need is expressed.

jsanda (approved):
LGTM and thanks for the explanation!

PR #1029, opened 2021-03-04
Fix EACH and LOCAL modes

### LINKED ISSUE #1028: datacenterAvailability: EACH might be broken with useAddressTranslator: true
I have inherited a cluster and am trying to get Reaper 2.2.1 working, and it seems to be completely unable to start any repairs. Depending on how it's started, either the repair job never appears as running, or every segment is stuck in NOT_STARTED forever.

According to the log, segments keep getting postponed because metrics are unavailable. Also according to the logs, it keeps trying (and failing) to connect to JMX endpoints that are in other datacenters, which doesn't work. I think it shouldn't attempt to do that with `datacenterAvailability: EACH`, but maybe `useAddressTranslator: true` is confusing the reachability check?

Starting a new repair job (that isn't already existing and in running state) also fails with a JMX error, from trying to connect to a node it shouldn't connect to.

It's entirely possible I'm using it wrong, I have no previous experience with the reaper, but I feel like I've checked all the obvious things and at this point it seems like a bug.

adejanovski:
The metrics request queue was removed in 2.2 to rely on metrics regularly extracted to the backend during heartbeats.
The code was missing support for such extractions in EACH and LOCAL modes, which made it impossible to perform repairs in EACH mode especially.
There was also a some dead code which only served the obsolete metrics request queue system.
The failed/successful connection counters to optimize scenarios where cross DC JMX is forbidden were fixed as well as they were sticking at their initial value all the time. They now count correctly successful connections and avoid spamming the logs with failed connection attempts.

Fixes #1028


adejanovski:
@maikzumstrull, I have a PR ready which fixes the issue with EACH and adds integration tests for it.
Do you think you could test it in your environment if I provide you a packaged version in the format of your choice?

That would be very helpful!

maikzumstrull:
Yes, we use Docker to deploy so if you can get it on Docker Hub I can have our deployment system pull that particular image tag or hash and run it in the dev environment.

adejanovski:
> Yes, we use Docker to deploy so if you can get it on Docker Hub I can have our deployment system pull that particular image tag or hash and run it in the dev environment.

Awesome, building it now...

adejanovski:
@maikzumstrull, I just pushed the [2.2-eachfix](https://hub.docker.com/layers/thelastpickle/cassandra-reaper/2.2-eachfix/images/sha256-03122e40d20ccfc2effbaf2bf0cd9647f63a2e925f284e1120b899c2ac5fa5f1?context=repo) tag to Docker Hub for you to pull 🙏

maikzumstrull:
Thanks, percolating through our deploy process now.

maikzumstrull:
Got it started, it immediately promoted 7 repair jobs to "running" and they are actually working on and finishing segments. :+1: 

At least it's reporting successes, haven't dug down into Cassandra internals to confirm.

maikzumstrull:
Was running overnight, still seems good. The web interface seems a little wobbly – sometimes the page goes blank and needs a refresh to render again. That may not be a new issue though, and the actual repairs happening is definitely more important.

adejanovski:
Hi @maikzumstrull,
thanks for the update, it's very valuable.
I've seen the UI going blank in this distributed mode when clicking on nodes to access their metrics. I think we'd need to handle some error cases a bit better in the frontend.

jsanda (approved):
I tested the changes as follows:

* Deploy a cluster with two DCs via cccm
* Build Reaper from the 2.2.0 tag
* Start a Reaper instance with `datacenterAvailability: EACH` configured for DC1
* Start a Reaper instance with `datacenterAvailability: EACH` configured for DC2
* Observed that segments are not processed as metrics could not be obtained from all necessary nodes
* Shut down both Reaper instances
* Rebuild Reaper from the PR branch
* Restart both Reaper instances
* Verify that segments are processed and repair completes

adejanovski:
Thanks @jsanda!

Merging

PR #1032, opened 2021-03-09
Batch metric inserts by partition to reduce Reaper write throughput

adejanovski:
Reaper stores extracts metrics from nodes during heartbeats and stores them in the backend when using collocated modes (SIDECAR, EACH, LOCAL).
Such metrics can be then be viewed in the UI.
This commit optimizes the inserts of metrics that belong to the same partition, which are now batched together.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/storage/PostgresStorage.java
+++ b/src/server/src/main/java/io/cassandrareaper/storage/PostgresStorage.java
@@ -841,52 +841,54 @@ public class PostgresStorage implements IStorage, IDistributedStorage {
   }
 
   @Override
-  public void storeMetric(GenericMetric metric) {
+  public void storeMetrics(List<GenericMetric> metrics) {
     if (null != jdbi) {
       try (Handle h = jdbi.open()) {
-        Instant metricTs = Instant.ofEpochMilli(metric.getTs().getMillis());
-        IStoragePostgreSql storage = getPostgresStorage(h);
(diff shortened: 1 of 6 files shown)


src/server/src/main/java/io/cassandrareaper/storage/PostgresStorage.java
```diff
@@ -841,52 +841,54 @@ private boolean segmentIsWithinRanges(RepairSegment seg, List<RingRange> ranges)
...
       try (Handle h = jdbi.open()) {
-        Instant metricTs = Instant.ofEpochMilli(metric.getTs().getMillis());
-        IStoragePostgreSql storage = getPostgresStorage(h);
-        int rowsUpdated = storage.updateMetricSourceNodeTimestamp(
-            metric.getClusterName(),
-            metric.getHost(),
-            metricTs
-        );
-        if (rowsUpdated == 0) {
+        for (GenericMetric metric:metrics) {
```
jsanda, line 847: From the description I thought this PR just involved changes with writes to the Cassandra backend. Can you explain these changes? (mostly because I don't understand 😄)

adejanovski: Yes, it's the case but in order to achieve this, I had to modify the signature of a method from an interface that is also used by the postgres backend class.
Instead of storing metrics one by one, I know provide a list of metrics which will be batched in Cassandra but inserted one by one in Postgres.

jsanda (approved):
LGTM and thanks for the explanation!

PR #1034, opened 2021-03-09
K8ssandra-483 rebrand

jeffbanks:
A phased approach, which includes rebranding for general docs and comments in support of [k8ssandra-483](https://github.com/k8ssandra/k8ssandra/issues/483).

Change of naming per requirement to replace `Cassandra Reaper` with the rebranded wording `Reaper for Apache Cassandra`.

### DIFF
--- a/docs/docs/backends/index.html
+++ b/docs/docs/backends/index.html
@@ -321,7 +321,7 @@
                     <div class="col-md-9" id="blog-listing-medium">
 
                         <h1 id="backends">Backends</h1>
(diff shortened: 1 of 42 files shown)


adejanovski (approved):

michaelsembwever:
Hey @jeffbanks, in this project please don't merge master into PR branches. It is a practice we have avoided so far, for the sake of version control hygiene.

The hygiene is based on the practice that merging must not happen in two directions in a git repository, if we are to avoid noise like this:
Screenshot 2021-03-10 at 20 11 44

That is, in the commit in our `master` branch a commit that says `Merge branch 'master' into k8ssandra-483-rebrand` is but noise, and makes life more difficult for us the maintainers.

The correct approach is to rebase the PR branch, rather than merge it, when you need to bring it up to date. (When it is then merged into master it should become a fast-forward merge.) 

Check out the following presentations  http://www.youtube.com/watch?v=AJ-CpGsCpM0 and https://www.perforce.com/video-tutorials/mainline-model-tofu-scale on some of the related topics:
- why fixes should be merged forwards from more-stable branches, rather than back-ported from trunk
- "don't drive through hedges"
- "always accept stabilising changes. never impose destabilising changes"
- "merge down, copy up"
- "tofu scale: firm above, soft below"

The "merge down, copy up" would be the most relevant part, but it all is relevant really…


Otherwise, thanks for fixing this important legal breach!

PR #1041, opened 2021-03-25
Fix segment count per node not being correctly accounted for in schedules

### LINKED ISSUE #1030: Schedule "segment per node" parameter is ignored
**Spotted on 2.2.1 (4be9bfc)** and also on 2.1.3

When creating a new **schedule** with "Segments per node" valorized, it is ignored and default 0 is persisted.

Creating a new **repair** run with "Segments per node" valorized works fine.

adejanovski:
The segment count per node was ignored in schedules as the UI didn't set the right query param:
The repair run resource was expecting `segmentCount` while the repair schedule resource was expecting `segmentCountPerNode`.

I've made both consistent to use `segmentCountPerNode` so that the setting is properly accounted for in both schedule and run creation.

Fixes #1030


emerkle826:
@adejanovski The changes look fine to me. I'll test locally ASAP and approve the request once I walk my way through it.

emerkle826 (approved):
Tested locally and the fix seems to work as designed!

Issue #968, opened 2020-10-02
Issues when repairing ScyllaDB


rmb938:
It looks like there is some mis-match between what Cassandra Reaper is  looking for and what ScyllaDB provides.

When I try and create a repair  on the `system_auth` keyspace the following shows up in the Cassandra Reaper logs

```
ERROR  [2020-10-02 21:08:37,227] [dw-55 - POST /repair_run?clusterName=sample&keyspace=system_traces&owner=test&cause=test&repairThreadCount=1&incrementalRepair=false] i.c.j.JmxProxy - Scylla API server HTTP GET to URL '/storage_service/range/system_traces' failed: Not found
WARN   [2020-10-02 21:08:37,228] [dw-55 - POST /repair_run?clusterName=sample&keyspace=system_traces&owner=test&cause=test&repairThreadCount=1&incrementalRepair=false] i.c.s.RepairRunService - couldn't connect to any host: [100.96.217.92, 100.96.37.79, 100.96.6.214], life sucks...
io.cassandrareaper.ReaperException: java.util.concurrent.ExecutionException: io.cassandrareaper.ReaperException: Scylla API server HTTP GET to URL '/storage_service/range/system_traces' failed: Not found
	at io.cassandrareaper.jmx.ClusterFacade.getRangeToEndpointMap(ClusterFacade.java:279)
	at io.cassandrareaper.service.RepairRunService.generateSegments(RepairRunService.java:159)
	at io.cassandrareaper.service.RepairRunService.registerRepairRun(RepairRunService.java:97)
	at io.cassandrareaper.resources.RepairRunResource.addRepairRun(RepairRunResource.java:237)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.glassfish.jersey.server.model.internal.ResourceMethodInvocationHandlerFactory$1.invoke(ResourceMethodInvocationHandlerFactory.java:81)
	at org.glassfish.jersey.server.model.internal.AbstractJavaResourceMethodDispatcher$1.run(AbstractJavaResourceMethodDispatcher.java:144)
[...]

PR #1045, opened 2021-04-09
Add basic support for scylla 4.4 repair

averzicco:
On scylla 4.4 release includes this [change](https://github.com/scylladb/scylla/commit/48c3c94aa6c4c301c02c0dddac0711eb6dc262c7) which can now be used from cassandra reaper to run repairs (see #968 for some context). Sadly the data returned from that method have a slightly different format, for example if cassandra return as example:
```
  public static Map<List<String>, List<String>> threeNodeClusterWithIps() {
    Map<List<String>, List<String>> map = new HashMap<List<String>, List<String>>();
    map = addRangeToMap(map, "0", "100", "127.0.0.1", "127.0.0.2", "127.0.0.3");
    map = addRangeToMap(map, "100", "200", "127.0.0.2", "127.0.0.3", "127.0.0.1");
    map = addRangeToMap(map, "200", "0", "127.0.0.3", "127.0.0.1", "127.0.0.2");
    return map;
  }
```
scylla example:

```
  public static Map<List<String>, List<String>> scyllaThreeNodeClusterWithIps() {
    Map<List<String>, List<String>> map = new HashMap<List<String>, List<String>>();
    map = addRangeToMap(map, "", "0", "127.0.0.3", "127.0.0.1", "127.0.0.2");
    map = addRangeToMap(map, "0", "100", "127.0.0.1", "127.0.0.2", "127.0.0.3");
    map = addRangeToMap(map, "100", "200", "127.0.0.2", "127.0.0.3", "127.0.0.1");
    map = addRangeToMap(map, "200", "", "127.0.0.3", "127.0.0.1", "127.0.0.2");
    return map;
  }
```

this PR is to handle correctly scylla format, plus some string cleanup/handling. It's not aimed at supporting scylla for all cassandra reaper features, but mostly to be able to run repair/scheduled repairs, in particular with changes included in this PR you would be able to:
- add a scylla cluster to cassandra reaper without causing exception/issues on the web UI
- create a repair/scheduled repair and execute it

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/jmx/ClusterFacade.java
+++ b/src/server/src/main/java/io/cassandrareaper/jmx/ClusterFacade.java
@@ -68,6 +68,7 @@ import com.google.common.base.Preconditions;
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.collect.Lists;
+import com.google.common.collect.Maps;
 import org.apache.commons.lang3.tuple.Pair;
 import org.joda.time.DateTime;
 import org.slf4j.Logger;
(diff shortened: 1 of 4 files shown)


averzicco:
hi, is there anything missing in the PR or should I just wait?

adejanovski:
Hi @averzicco, I didn't find time yet to review this PR but hang in tight, it's on my todo list ;)

src/server/src/main/java/io/cassandrareaper/jmx/ClusterFacade.java
```diff
@@ -772,8 +773,34 @@ public void clearSnapshot(String snapshotName, Node host) throws ReaperException
...
       Cluster cluster,
       String keyspace) throws ReaperException {
-
-    return connect(cluster).getRangeToEndpointMap(keyspace);
+    JmxProxy jmxConnection = connect(cluster);
+    Map<List<String>, List<String>> endpointMap = jmxConnection.getRangeToEndpointMap(keyspace);
+    return endpointCleanup(endpointMap);
+  }
+
+  protected static Map<List<String>, List<String>> endpointCleanup(Map<List<String>, List<String>> endpointMap) {
```
adejanovski: The name of the method is not enough here to understand what it's really doing and we don't have comments either.
Is this method converting Scylla endpoint lists to be compatible with Cassandra? Is it doing more?

adejanovski (changes requested):
I just have one comment about making it simpler to understand what the `endpointCleanup()`  method does when reading the code. It will make maintenance simpler and avoid breaking things.
Thanks!

src/server/src/main/java/io/cassandrareaper/jmx/ClusterFacade.java
averzicco: The method makes the Scylla endpoint list compatible with the Cassandra ones but gets called also while using cassandra (this is the reason while I didn't name it scyllaEndpointCleanup). Do you have any suggestion on how to make it more clear?

adejanovski: Adding what you just wrote in an element comment on the method would do it :)
You could also rename the method to `maybeCleanupEndpointFromScylla()` as the method may or may not do something (I understand it's a noop for Cassandra).

averzicco: Looks fair, thanks!
Applied changes as you suggested.

averzicco:
hi @adejanovski, is there anything missing on my side for this PR to be approved?

adejanovski (approved):

adejanovski:
Looks good @averzicco, sorry for the delay.
Thanks!

PR #1072, opened 2021-05-21
Fix percent repaired metrics purge issues

adejanovski:


emerkle826 (approved):
This LGTM. I did a sanity check and it compiles locally and the PG updates look good.

adejanovski:
Thanks @emerkle826 !

PR #1094, opened 2021-06-16
Fix endpoint state not returning in EACH mode due to seeds being shuffled

### LINKED ISSUE #1092: EACH datacenterAvailability can cause endpointStates to be empty list
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=13938552)

When you are using `datacenterAvailability: EACH`, the assumption is that Reaper will only have access to 1 DC (out of how ever many exist in the cluster). It's therefore possible that the call for getNodesStatus() will pick 3 nodes that it doesn't have access to, which causes the UI to fail in various ways (as well as make kicking off repairs impossible). On the clusters page, it actually blanks out the page. On the repairs page, it simply doesn't show any keyspaces available for the given cluster. 

https://github.com/thelastpickle/cassandra-reaper/blob/27a405cbe657268a231fcf39dc33a75223d9819a/src/server/src/main/java/io/cassandrareaper/resources/ClusterResource.java#L438-L457

Slack Conversation about Issue: https://the-asf.slack.com/archives/C010YBGLBF1/p1623484344111200

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/REAP-106) by [Unito](https://www.unito.io)
┆Issue Number: REAP-106

adejanovski:
Fixes #1092 

When getting the endpoint state through JMX, the implementation was shuffling nodes from the cluster and picking the first three without caring whether or not these nodes are accessible.
We now have methods in ClusterFacade that can connect to any node and invoke JMX operations while taking into account previous connection attempts to filter nodes.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/ReaperApplication.java
+++ b/src/server/src/main/java/io/cassandrareaper/ReaperApplication.java
@@ -213,10 +213,7 @@ public final class ReaperApplication extends Application<ReaperApplicationConfig
     final PingResource pingResource = new PingResource(healthCheck);
     environment.jersey().register(pingResource);
 
-    final ClusterResource addClusterResource = new ClusterResource(
-        context,
-        cryptograph,
-        environment.lifecycle().executorService("ClusterResource").minThreads(6).maxThreads(6).build());
+    final ClusterResource addClusterResource = ClusterResource.create(context, cryptograph);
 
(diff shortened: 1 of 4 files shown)


adejanovski:
This PR is totally outdated. I'll close it and we'll see if it can be ported over to the latest master branch.

PR #1095, opened 2021-06-16
Allow crypto options to be configured in Docker image

### LINKED ISSUE #1093: Missing cryptograph option in 2.2.5 image
Hi
First of all thank you for keep the reaper alive.
I just updated to version 2.2.5 in my k8s system and seems that this part is missing in `casssandra-reaper.yaml`
```
#cryptograph:
#  type: symmetric
#  systemPropertySecret: SOME_SYSTEM_PROPERTY
```
So yes, the WebUI keeps complaining about enable encryption before adding JMX Credential.
The workaround is to mount a pre-defined yaml file as ConfigMap but it's not ideal to do it.
Would be nice if you can make a HF update.
Thanks !

adejanovski:
Fixes #1093


src/server/src/main/docker/configure-jmx-credentials.sh
```diff
@@ -49,3 +49,10 @@ jmxAuth:
 EOT
 
 fi
+
+if [ ! -z "${CRYPTO_SYSTEM_PROPERTY_SECRET}" ]; then
+cat <<EOT >> /etc/cassandra-reaper.yml
+cryptograph:
+  type: symmetric
+  systemPropertySecret: ${CRYPTO_SYSTEM_PROPERTY_SECRET}
+EOT
```
ducnm0711, line 57: Hi @adejanovski 
Script can't run. You missed `fi` end of file.

PR #1096, opened 2021-06-28
Add schema migration startup mode

adejanovski:
This will help performing migrations pre deployment of Reaper when it makes sense. In the context of k8s, the main containers will have probes set to values that won't allow all migrations to go through, leaving the schema in a state that's incompatible with the migrations that would be subsequently replayed (such as `ALTER TABLE ADD COLUMN` statements).

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/ReaperApplication.java
+++ b/src/server/src/main/java/io/cassandrareaper/ReaperApplication.java
@@ -45,11 +45,8 @@ import io.cassandrareaper.service.Heart;
 import io.cassandrareaper.service.PurgeService;
 import io.cassandrareaper.service.RepairManager;
 import io.cassandrareaper.service.SchedulingManager;
-import io.cassandrareaper.storage.CassandraStorage;
 import io.cassandrareaper.storage.IDistributedStorage;
-import io.cassandrareaper.storage.IStorage;
-import io.cassandrareaper.storage.MemoryStorage;
-import io.cassandrareaper.storage.PostgresStorage;
+import io.cassandrareaper.storage.InitializeStorage;
 
 import java.util.EnumSet;
 import java.util.Map;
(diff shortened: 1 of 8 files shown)


src/server/src/main/java/io/cassandrareaper/ReaperApplication.java
```diff
@@ -179,6 +179,10 @@ public void run(ReaperApplicationConfiguration config, Environment environment)
         repairThreads, maxParallelRepairs);
 
     tryInitializeStorage(config, environment);
+    if (config.getSchemaMigrationMode()) {
+      LOG.info("Reaper is running in schema migration mode. Exiting now that the schema was created.");
```
burmanm: And if this fails, the normal stack trace is just thrown? Does it indicate enough to the user that the initialisation only failed? Or could we at least in the startup say "running in schema migration only" instead of only at the successful exit?

adejanovski:
> I think it would make more sense that a dropwizard command is implemented here rather than a configuration flag.
> ref: https://www.dropwizard.io/en/latest/manual/core.html#man-core-commands
> 
> ```
> java ${JVM_OPTS} -cp ${CLASS_PATH} io.cassandrareaper.ReaperApplication check ${CONFIG_PATH}
> java ${JVM_OPTS} -cp ${CLASS_PATH} io.cassandrareaper.ReaperApplication server ${CONFIG_PATH}
> java ${JVM_OPTS} -cp ${CLASS_PATH} io.cassandrareaper.ReaperApplication db ${CONFIG_PATH}
> ```
> 
> See also https://www.dropwizard.io/en/latest/manual/internals.html
> (Dropwizard already reserves the `db` command for usages just like this…)
> 
> (This was also suggested in [#957 (comment)](https://github.com/thelastpickle/cassandra-reaper/pull/957#issuecomment-694118402) )

You're absolutely right @michaelsembwever 👍 
The default `db` command is wired up with liquibase and cannot be overridden. I've created a new `schema-migrate` command which does the job.
Quick note: I've tried using an `EnvironmentCommand` and struggled for a while as it was running the actual `ReaperApplication.run()` code despite not using the `server` command (and we need the `environment` variable to set up the db connections). Turns out, for whatever reason, that's the intended behavior: https://github.com/dropwizard/dropwizard/blob/19051ff1a80d7cdffa268d21f26a9e882ef7c82b/dropwizard-core/src/main/java/io/dropwizard/cli/EnvironmentCommand.java#L43

I ended up using a `ConfiguredCommand` instead and recreating the `environment` variable.

src/server/src/main/java/io/cassandrareaper/ReaperApplication.java
```diff
@@ -570,4 +477,29 @@ private void initializeJmxSeedsForAllClusters() {
...
+        try {
+          context.storage = InitializeStorage.initializeStorage(
+            config, environment, context.reaperInstanceId).initializeStorageBackend();
+          break;
+        } catch (RuntimeException e) {
+          LOG.error("Storage is not ready yet, trying again to connect shortly...", e);
+          storageFailures++;
+          if (storageFailures > 60) {
+            LOG.error("Too many failures when trying to connect storage. Exiting :'(");
+            System.exit(1);
```
michaelsembwever: would exception handling at the command level be more appropriate place to do the `exit(..)` ? 
(that way you can easily list, and read, the different exit codes that exist)

src/server/src/main/java/io/cassandrareaper/ReaperDbMigrationCommand.java
```diff
@@ -0,0 +1,52 @@
...
+import io.cassandrareaper.storage.InitializeStorage;
+
+import javax.validation.Validation;
+
+import io.dropwizard.cli.ConfiguredCommand;
+import io.dropwizard.setup.Bootstrap;
+import io.dropwizard.setup.Environment;
+import net.sourceforge.argparse4j.inf.Namespace;
+
+public final class ReaperDbMigrationCommand extends ConfiguredCommand<ReaperApplicationConfiguration> {
```
michaelsembwever: can be package-protected class?

michaelsembwever (approved):

michaelsembwever:
> The default db command is wired up with liquibase and cannot be overridden. I've created a new schema-migrate command which does the job.

What a shame… poor design by dropwizard i say.

src/server/src/main/java/io/cassandrareaper/ReaperApplication.java
adejanovski: done 👍

src/server/src/main/java/io/cassandrareaper/ReaperDbMigrationCommand.java
adejanovski: done 👍

PR #1107, opened 2021-07-23
Add adaptive segment timeouts

### LINKED ISSUE #1059: Store timeout threshold on segments
In order to make repairs adaptive, we need to store the timeout on each individual segment instead of using a global fixed setting.
The timeout should be added to repair units so that a value can be set in schedules.
Each SegmentRunner should consider the segment's timeout as threshold as part of this change.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-209) by [Unito](https://www.unito.io)
┆Issue Number: K8SSAND-209

### LINKED ISSUE #1060: Double segment timeout on failure
If a segment reaches its timeout and gets canceled, the timeout should be doubled to raise the chances of success in the following attempt.

A new `FAILED` state should be created for segments that failed 4 times (that would make a total of ~8 hours already for a single segment).
At the end of a repair, if at least one segment was marked as `FAILED` then the whole repair should be marked as `FAILED`.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-207) by [Unito](https://www.unito.io)
┆Issue Number: K8SSAND-207

adejanovski:
Fixes #1059 
Fixes #1060 

As a first step towards adaptive repairs, this commit adds the segment timeout in the repair units using the configuration timeout as default.
This timeout can be set differently for each repair run or schedule.
Segments failing will get an extended timeout at each attempt to avoid blocking misconfigured repairs and heavy segments.

Unlike imagined in #1059 and #1060, timeouts don't need to be stored on the segments since we can compute the extended timeouts using the fail count. Also the timeout extensions do not double each time but rather apply the following formula: `timeout = segmentTimeout * failCount`

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
+++ b/src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
@@ -108,7 +108,8 @@ public final class RepairRunResource {
       @QueryParam("datacenters") Optional<String> datacentersToRepairParam,
       @QueryParam("blacklistedTables") Optional<String> blacklistedTableNamesParam,
       @QueryParam("repairThreadCount") Optional<Integer> repairThreadCountParam,
-      @QueryParam("force") Optional<String> forceParam) {
+      @QueryParam("force") Optional<String> forceParam,
+      @QueryParam("timeout") Optional<Integer> timeoutParam) {
 
     try {
       final Response possibleFailedResponse
@@ -126,7 +127,8 @@ public final class RepairRunResource {
           datacentersToRepairParam,
           blacklistedTableNamesParam,
           repairThreadCountParam,
-          forceParam);
+          forceParam,
+          timeoutParam);
 
       if (null != possibleFailedResponse) {
         return possibleFailedResponse;
@@ -198,6 +200,8 @@ public final class RepairRunResource {
         return Response.status(Response.Status.NOT_FOUND).entity(ex.getMessage()).build();
       }
 
+      int timeout = timeoutParam.orElse(context.config.getHangingRepairTimeoutMins());
+      boolean force = (forceParam.isPresent() ? Boolean.parseBoolean(forceParam.get()) : false);
 
       RepairUnit.Builder builder = RepairUnit.builder()
               .clusterName(cluster.getName())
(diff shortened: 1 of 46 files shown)


src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
```diff
@@ -1060,18 +1064,7 @@ public RepairUnit getRepairUnit(UUID id) {
...
-          && repairUnitRow.getSet("column_families", String.class).equals(params.columnFamilies)
-          && repairUnitRow.getBool("incremental_repair") == params.incrementalRepair
-          && repairUnitRow.getSet("nodes", String.class).equals(params.nodes)
-          && repairUnitRow.getSet("datacenters", String.class).equals(params.datacenters)
-          && repairUnitRow
-              .getSet("blacklisted_tables", String.class)
-              .equals(params.blacklistedTables)
-          && repairUnitRow.getInt("repair_thread_count") == params.repairThreadCount) {
-
-        repairUnit = RepairUnit.builder()
```
michaelsembwever, line 1074: why is this condition removed?

adejanovski: It is replaced with this check: https://github.com/thelastpickle/cassandra-reaper/pull/1107/files/aca650eef09ee895debd2c87f4e8629e5c47c17f#diff-1eb3c242ea87cb7ce434d35444bd88e1531fe3ad0aae4a3a0e6d96f3411ce56dR1078

The problem here is that we had two different ways of checking if repair units were "equal": the removed code (which is specific to the Cassandra backend) and the `equals()` method in `RepairUnit`.
Now we're left with a single method to check repair unit equality.

src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
```diff
@@ -207,10 +210,11 @@ public Response addRepairRun(
...
               .datacenters(datacentersToRepair)
               .blacklistedTables(blacklistedTableNames)
-              .repairThreadCount(repairThreadCountParam.orElse(context.config.getRepairThreadCount()));
-
-      final RepairUnit theRepairUnit = repairUnitService.getOrCreateRepairUnit(cluster, builder);
+              .repairThreadCount(repairThreadCountParam.orElse(context.config.getRepairThreadCount()))
+              .timeout(timeout);
 
+      // Repair unit overlaps shouldn't affect one shot repairs. Setting the "force" setting to true.
+      final RepairUnit theRepairUnit = repairUnitService.getOrCreateRepairUnit(cluster, builder, true);
```
michaelsembwever: I'm not sure I agree with this. Maybe I need more context… 

It should not be possible to run a on-shot repair over the top of an existing repair unit, without first warning the user. The confirmation from the user then becomes the explicit `force=true`.

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
michaelsembwever: for permanence: https://github.com/thelastpickle/cassandra-reaper/commit/aca650eef09ee895debd2c87f4e8629e5c47c17f#diff-1eb3c242ea87cb7ce434d35444bd88e1531fe3ad0aae4a3a0e6d96f3411ce56dR1078

src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
michaelsembwever: `force` got parameterised 💪

michaelsembwever (commented):
:shipit:

michaelsembwever (dismissed):

PR #1110, opened 2021-08-03
Allow repair schedules to be adaptive

### LINKED ISSUE #1061: Update schedules based on last run feedback
if timeouts occurred during a scheduled repair, the schedule should be updated as follows:

* if more than 20% of segments got a timeout extension, the number of segments per node should be doubled
* if less than 20% of segments got a timeout extension then the timeout should be extended to twice the longest segment duration

This feature should allow opt-out through an `adaptiveRepair` setting which would default to `true` in the configuration file.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-206) by [Unito](https://www.unito.io)
┆Issue Number: K8SSAND-206

adejanovski:
Fixes #1061

Adaptive schedules will be updated at the end of each repair if the run metrics require it. The schedule can get a higher/lower number of segments or a higher segment timeout depending on the latest run.
The applied rules are the following: 

- if more than 20% segments were extended, the number of segments will be raised by 20% on the schedule
- if less than 20% segments were extended (and at least one), the timeout will be set to twice the current timeout
- if no segment was extended and the maximum duration of segments is below 5 minutes, the number of segments will be reduced by 10% with a minimum of 16 segments per node.

The changes above are not aggressive on purpose (10% to 20% changes in numbers of segments and a minimum of 16 segments per node) to apply gradual changes and generate safe automated changes.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/core/RepairSchedule.java
+++ b/src/server/src/main/java/io/cassandrareaper/core/RepairSchedule.java
@@ -37,13 +37,13 @@ public final class RepairSchedule {
   private final int daysBetween;
   private final DateTime nextActivation;
   private final ImmutableList<UUID> runHistory;
-  @Deprecated private final int segmentCount;
   private final RepairParallelism repairParallelism;
   private final double intensity;
   private final DateTime creationTime;
   private final String owner;
   private final DateTime pauseTime;
   private final int segmentCountPerNode;
+  private final boolean adaptive;
 
   private RepairSchedule(Builder builder, UUID id) {
     this.id = id;
@@ -52,13 +52,13 @@ public final class RepairSchedule {
     this.daysBetween = builder.daysBetween;
     this.nextActivation = builder.nextActivation;
     this.runHistory = builder.runHistory;
-    this.segmentCount = builder.segmentCount;
     this.repairParallelism = builder.repairParallelism;
     this.intensity = builder.intensity;
     this.creationTime = builder.creationTime;
     this.owner = builder.owner;
     this.pauseTime = builder.pauseTime;
     this.segmentCountPerNode = builder.segmentCountPerNode;
+    this.adaptive = builder.adaptive;
   }
 
   public static Builder builder(UUID repairUnitId) {
@@ -101,10 +101,6 @@ public final class RepairSchedule {
     return new LongCollectionSqlType(list);
   }
 
-  public int getSegmentCount() {
-    return segmentCount;
-  }
-
   public int getSegmentCountPerNode() {
     return segmentCountPerNode;
   }
@@ -129,6 +125,10 @@ public final class RepairSchedule {
     return pauseTime;
   }
 
+  public boolean getAdaptive() {
+    return adaptive;
+  }
+
   public Builder with() {
     return new Builder(this);
   }
@@ -151,14 +151,13 @@ public final class RepairSchedule {
     private Integer daysBetween;
     private DateTime nextActivation;
     private ImmutableList<UUID> runHistory = ImmutableList.<UUID>of();
-    @Deprecated private int segmentCount = 0;
     private RepairParallelism repairParallelism;
     private Double intensity;
     private DateTime creationTime = DateTime.now();
     private String owner = "";
     private DateTime pauseTime;
     private Integer segmentCountPerNode;
-    private boolean majorCompaction = false;
+    private boolean adaptive = false;
 
     private Builder(UUID repairUnitId) {
       this.repairUnitId = repairUnitId;
@@ -170,7 +169,6 @@ public final class RepairSchedule {
       daysBetween = original.daysBetween;
       nextActivation = original.nextActivation;
       runHistory = original.runHistory;
-      segmentCount = original.segmentCount;
       repairParallelism = original.repairParallelism;
       intensity = original.intensity;
       creationTime = original.creationTime;
@@ -178,6 +176,7 @@ public final class RepairSchedule {
       pauseTime = original.pauseTime;
       intensity = original.intensity;
       segmentCountPerNode = original.segmentCountPerNode;
+      adaptive = original.adaptive;
     }
 
     public Builder state(State state) {
@@ -197,11 +196,6 @@ public final class RepairSchedule {
 
     public Builder runHistory(ImmutableList<UUID> runHistory) {
       this.runHistory = runHistory;
-      return this;
-    }
-
-    public Builder segmentCount(int segmentCount) {
-      this.segmentCount = segmentCount;
       return this;
     }
 
@@ -232,6 +226,11 @@ public final class RepairSchedule {
 
     public Builder segmentCountPerNode(int segmentCountPerNode) {
       this.segmentCountPerNode = segmentCountPerNode;
+      return this;
+    }
+
+    public Builder adaptive(boolean adaptive) {
+      this.adaptive = adaptive;
       return this;
     }
 
--- a/src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
+++ b/src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
@@ -24,6 +24,7 @@ import io.cassandrareaper.core.Cluster;
 import io.cassandrareaper.core.CompactionStats;
 import io.cassandrareaper.core.Node;
 import io.cassandrareaper.core.RepairRun;
+import io.cassandrareaper.core.RepairSchedule;
 import io.cassandrareaper.core.RepairSegment;
 import io.cassandrareaper.core.RepairUnit;
 import io.cassandrareaper.core.Segment;
@@ -37,6 +38,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.NoSuchElementException;
 import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
@@ -51,6 +53,8 @@ import java.util.stream.Collectors;
 
 import com.codahale.metrics.Gauge;
 import com.codahale.metrics.MetricRegistry;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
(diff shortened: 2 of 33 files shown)


src/server/src/main/java/io/cassandrareaper/core/RepairSchedule.java
```diff
@@ -37,13 +37,13 @@
   private final int daysBetween;
   private final DateTime nextActivation;
   private final ImmutableList<UUID> runHistory;
-  @Deprecated private final int segmentCount;
```
michaelsembwever, line 40: Why is `segmentCount` getting removed?

src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
```diff
@@ -272,10 +282,108 @@ private void endRepairRun() {
...
       }
     }
   }
 
+  /**
+   * Tune segment timeout and number of segments for adaptive schedules.
+   * Checks that the run was triggered by an adaptive schedule and gathers info on the run to apply tunings.
+   */
+  @VisibleForTesting
+  public void maybeAdaptRepairSchedule() {
```
michaelsembwever, line 296: I don't see the need for  `public` 🤷🏻‍♀️

src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
```diff
@@ -272,10 +282,108 @@ private void endRepairRun() {
...
+          = ((float) extendedSegments / (float) repairRun.get().getSegmentCount()) * 100.0;
+      LOG.info("extendedSegments = {}, total segments = {}, percent extended = {}",
+          extendedSegments,
+          repairRun.get().getSegmentCount(),
+          percentExtendedSegments);
+      tuneAdaptiveRepair(percentExtendedSegments, maxSegmentDurationInMins);
+    }
+  }
+
+  @VisibleForTesting
```
michaelsembwever, line 317: I don't see the need for `@VisibleForTesting` 🤷🏻‍♀️

src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
```diff
@@ -272,10 +282,108 @@ private void endRepairRun() {
...
+    } else if ((int) percentExtendedSegments <= PERCENT_EXTENDED_THRESHOLD && percentExtendedSegments >= 1) {
+      // The number of extended segments is moderate and timeout should be extended
+      raiseTimeoutOfUnit();
+    } else if (percentExtendedSegments == 0 && maxSegmentDuration < SEGMENT_DURATION_FOR_REDUCTION_THRESHOLD) {
+      // Segments are being executed very fast and could be lowered
+      reduceSegmentsPerNodeToScheduleForUnit();
+    }
+  }
+
+  @VisibleForTesting
```
michaelsembwever, line 332: I don't see the need for `@VisibleForTesting` 🤷🏻‍♀️

src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
```diff
@@ -272,10 +282,108 @@ private void endRepairRun() {
...
+  void tuneAdaptiveRepair(double percentExtendedSegments, int maxSegmentDuration) {
+    LOG.info("Percent extended segments: {}", percentExtendedSegments);
+    if (percentExtendedSegments > PERCENT_EXTENDED_THRESHOLD) {
+      // Too many segments were extended, meaning that the number of segments is too low
+      addSegmentsPerNodeToScheduleForUnit();
+    } else if ((int) percentExtendedSegments <= PERCENT_EXTENDED_THRESHOLD && percentExtendedSegments >= 1) {
+      // The number of extended segments is moderate and timeout should be extended
+      raiseTimeoutOfUnit();
+    } else if (percentExtendedSegments == 0 && maxSegmentDuration < SEGMENT_DURATION_FOR_REDUCTION_THRESHOLD) {
+      // Segments are being executed very fast and could be lowered
```
michaelsembwever:
```suggestion
      // Segments are being executed very fast so segment count could be lowered
```

src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
```diff
@@ -272,10 +282,108 @@ private void endRepairRun() {
...
+        = (int) Math.max(MIN_SEGMENTS_PER_NODE_REDUCTION, scheduleToTune.getSegmentCountPerNode() / 1.1);
+
+    // update schedule with new segment per node value
+    RepairSchedule newSchedule
+        = scheduleToTune.with().segmentCountPerNode(newSegmentCountPerNode).build(scheduleToTune.getId());
+    context.storage.updateRepairSchedule(newSchedule);
+
+  }
+
+  @VisibleForTesting
```
michaelsembwever, line 348: I don't see the need for `@VisibleForTesting` 🤷🏻‍♀️

src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
```diff
@@ -272,10 +282,108 @@ private void endRepairRun() {
...
+  void raiseTimeoutOfUnit() {
+    // Build updated repair unit
+    RepairUnit updatedUnit = repairUnit.with().timeout(repairUnit.getTimeout() * 2).build(repairUnit.getId());
+
+    // update unit with new timeout
+    context.storage.updateRepairUnit(updatedUnit);
+  }
+
+  @VisibleForTesting
```
michaelsembwever, line 357: I don't see the need for `@VisibleForTesting` 🤷🏻‍♀️

src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
```diff
@@ -272,10 +282,108 @@ private void endRepairRun() {
...
+    RepairSchedule newSchedule
+        = scheduleToTune.with().segmentCountPerNode(newSegmentCountPerNode).build(scheduleToTune.getId());
+    context.storage.updateRepairSchedule(newSchedule);
+
+  }
+
+  @VisibleForTesting
+  void raiseTimeoutOfUnit() {
+    // Build updated repair unit
+    RepairUnit updatedUnit = repairUnit.with().timeout(repairUnit.getTimeout() * 2).build(repairUnit.getId());
```
michaelsembwever, line 351: doubling is a big jump. any evidence/metrics on why we are choosing this? my guess would have been a 50% increase (but i really don't know)

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
```diff
@@ -1435,48 +1454,37 @@ private RepairSchedule createRepairScheduleFromRow(Row repairScheduleRow) {
...
-        session.executeAsync(
-            insertRepairScheduleByClusterAndKsPrepStmt.bind(
-                " ", repairUnit.getKeyspaceName(), newRepairSchedule.getId())));
-
-    try {
-      Futures.allAsList(futures).get();
-    } catch (InterruptedException | ExecutionException ex) {
-      LOG.error("failed to quorum update repair schedule " + newRepairSchedule.getId(), ex);
-    }
+    BatchStatement batchStatement = new BatchStatement();
```
michaelsembwever: why do we need to change this to batch in this PR?

src/server/src/main/java/io/cassandrareaper/core/RepairSchedule.java
adejanovski: It's been deprecated for a while and as I was working on segment count per node with this PR, I figured it was the right time to do some cleanup. This `segmentCount` variable is the former "global segment count" which we replaced with the segmentCountPerNode to have schedules self adjust to cluster expansions.

src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
adejanovski: Good catch, it's a leftover from trials on unit testing. I'll move it back to private.

src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
adejanovski: I'd say it's not a big risk as what can happen at worse is that the segment will take more time to be canceled if it gets stuck. I'd rather have that than multiple failed attempts as the odds of having stuck segments lowered a lot with the latest versions of Cassandra.
wdyt?

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
adejanovski: I ran into some issues when testing this part of the code. The async queries were returning errors that I couldn't figure out. I switched to a batch statement and it worked fine 🤷  I tried to debug this for a while without managing to understand the root cause, which is very unsatisfying 😕 
Nonetheless, I think we're better with a batch statement here for consistency reasons and keeping in mind that schedule creation shouldn't be happening very often.

src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
adejanovski: scratch that, I actually need this method to be `public` so that I can invoke it from the RepairRunner spy I'm creating ([here](https://github.com/thelastpickle/cassandra-reaper/pull/1110/files#diff-e06412fc9bcb60ad7dfbbce3c7cf2b725b2fe853e3fe95fe8461abd3b64412e8R1239) for example).

I'll revert the change.

michaelsembwever: [trivial] Can spy work with package-protected?  (which would be better than public)

src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
michaelsembwever: wfm

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
michaelsembwever: i'm uneasy about putting batch statements into people's clusters, like I am about LWTs. Reaper ideally should aim to be as light as possible, as folk do use their production C* as the Reaper backend… 

Can we resume debugging…? (sorry to be a pain)
A series of strongly consistent requests should not be causing a problem, and we would absolutely benefit from knowing why it is not working in practice…

src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
adejanovski: I just tried and it doesn't work, sorry.

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
adejanovski: Whatever happened when I was testing this is now gone and I can't reproduce the exception. We're back to the previous version of this method using async queries.

michaelsembwever (approved):
:shipit:

Issue #1046, opened 2021-04-09
Reaper 2.2.4 still does not show snapshots


saphireee:
List of snapshoots is empty. Older version 2.2.1 shows it but 2.2.3 not.

**Spotted on 2.2.3 (6c0281c)**

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-219) by [Unito](https://www.unito.io)
┆Issue Number: K8SSAND-219

saphireee:
version 2.2.4 also does not show snapshots

averzicco:
version `2.3.1` also does not show snapshots.

I've tried to debug the issue and apparently seems caused by this [check](https://github.com/thelastpickle/cassandra-reaper/blob/e64b689313a30e83fe68409cd6b294ef8b5b585f/src/server/src/main/java/io/cassandrareaper/jmx/ClusterFacade.java#L701-L705):
```
      if (context.config.getDatacenterAvailability().isInCollocatedMode()
          && context.jmxConnectionFactory.getHostConnectionCounters().getSuccessfulConnections(
              host.getHostname()) >= 0) {
        return SnapshotProxy.create(connect(host)).listSnapshots();
      }
```

in particular `context.config.getDatacenterAvailability().isInCollocatedMode()` return false when the `datacenterAvailability` configuration setting is `ALL` based on [this](https://github.com/thelastpickle/cassandra-reaper/blob/e64b689313a30e83fe68409cd6b294ef8b5b585f/src/server/src/main/java/io/cassandrareaper/ReaperApplicationConfiguration.java#L641-L666):
```
  public enum DatacenterAvailability {
    /* We require direct JMX access to all nodes across all datacenters */
    ALL,
    /* We require jmx access to all nodes in the local datacenter */
    LOCAL,
    /* Each datacenter requires at minimum one reaper instance that has jmx access to all nodes in that datacenter */
    EACH,
    /* Sets Reaper in sidecar mode where each Cassandra node has a collocated Reaper instance */
    SIDECAR;


    /**
     * Check if the current datacenter availability mode is to have collocation between Reaper and a DC/node.
     * @return true if we're in a collocated mode, false otherwise
     */
    public boolean isInCollocatedMode() {
      switch (this) {
        case LOCAL:
        case SIDECAR:
        case EACH:
          return true;
        default:
          return false;
      }
    }
  }
  ```

saphireee:
[...]

PR #1111, opened 2021-08-05
Fix collocation check on listSnapshots method

averzicco:
Fix the issue #1046
Basically will allow to retrieve snapshots also from reaper instances which are configured also with `DatacenterAvailability` = `ALL` (the only one excluded with the collocation check)

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/jmx/ClusterFacade.java
+++ b/src/server/src/main/java/io/cassandrareaper/jmx/ClusterFacade.java
@@ -698,8 +698,7 @@ public final class ClusterFacade {
    */
   public List<Snapshot> listSnapshots(Node host) throws ReaperException {
     try {
-      if (context.config.getDatacenterAvailability().isInCollocatedMode()
-          && context.jmxConnectionFactory.getHostConnectionCounters().getSuccessfulConnections(
+      if (context.jmxConnectionFactory.getHostConnectionCounters().getSuccessfulConnections(
               host.getHostname()) >= 0) {
         return SnapshotProxy.create(connect(host)).listSnapshots();
       }
--- a/src/server/src/main/java/io/cassandrareaper/ReaperApplication.java
+++ b/src/server/src/main/java/io/cassandrareaper/ReaperApplication.java
@@ -34,6 +34,7 @@ import io.cassandrareaper.resources.DiagEventSubscriptionResource;
 import io.cassandrareaper.resources.NodeStatsResource;
 import io.cassandrareaper.resources.PingResource;
 import io.cassandrareaper.resources.ReaperHealthCheck;
+import io.cassandrareaper.resources.ReaperResource;
 import io.cassandrareaper.resources.RepairRunResource;
 import io.cassandrareaper.resources.RepairScheduleResource;
 import io.cassandrareaper.resources.SnapshotResource;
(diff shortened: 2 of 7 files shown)


src/server/src/main/java/io/cassandrareaper/jmx/ClusterFacade.java
```diff
@@ -698,8 +698,7 @@ public boolean nodeIsAccessibleThroughJmx(String nodeDc, String node) {
    */
   public List<Snapshot> listSnapshots(Node host) throws ReaperException {
     try {
-      if (context.config.getDatacenterAvailability().isInCollocatedMode()
-          && context.jmxConnectionFactory.getHostConnectionCounters().getSuccessfulConnections(
+      if (context.jmxConnectionFactory.getHostConnectionCounters().getSuccessfulConnections(
```
adejanovski, line 701:
```suggestion
      if (!context.config.getDatacenterAvailability().isInCollocatedMode()
          && context.jmxConnectionFactory.getHostConnectionCounters().getSuccessfulConnections(
```

The check was indeed wrong as we need to make sure we're not in collocated (distributed) mode for the snapshots to be listed.
Listing snapshots involves a direct call to each node in the cluster which is most probably impossible in collocated mode.

adejanovski (changes requested):
Hi @averzicco,

thanks for pushing this PR!
As mentioned in my comment, there's a bug indeed and the condition should pass only if we're **not** in collocated mode.

src/server/src/main/java/io/cassandrareaper/jmx/ClusterFacade.java
averzicco: Hi @adejanovski 

seems fair. I've applied the change you suggested.

Thanks!

averzicco:
hi @adejanovski 

it seems that integration tests are failing because executed with `datacenterAvailability: LOCAL` but expect to be able to list snapshots.
Should I modify the integration test? 
Wouldn't this cause trouble to whom is using reaper right now with DC availability LOCAL, SIDECAR or EACH and snapshots are "working"?

adejanovski:
> it seems that integration tests are failing because executed with datacenterAvailability: LOCAL but expect to be able to list snapshots.
Should I modify the integration test?

Yes, currently they're validating the wrong behavior so they indeed need to be modified. We'll need to introduce a new test type "all" and introduce a new cucumber tag so that we can avoid running the snapshot test for other test types.
This is going to be a non trivial change, so let me know if you need me to provide more guidance.

> Wouldn't this cause trouble to whom is using reaper right now with DC availability LOCAL, SIDECAR or EACH and snapshots are "working"?

None of these settings should allow access to all nodes from all reaper instances, which gives two possibilities:

- It looks like it's working but it's only showing partial results
- Reaper should be configured with ALL instead

Either way, I'm fine with moving forward with the change.

Thanks!

averzicco:
what if we allow listing snapshots for all values of availability and warn the user on UI about partial results?

averzicco:
@adejanovski any inputs on this?

> what if we allow listing snapshots for all values of availability and warn the user on UI about partial results?

adejanovski:
> @adejanovski any inputs on this?
> 
> > what if we allow listing snapshots for all values of availability and warn the user on UI about partial results?

It's more involved as a change and requires some work on the frontend side, but if you manage to make this work and not confusing for users, then wfm 👍

averzicco:
@adejanovski added an endpoint used to retrieve the value of datacenterAvailability configuration parameter and modified the components which list snapshots to show the warning message `Reaper is configured with datacenterAvailability={datacenterAvailability}, only snapshots on reachable nodes are listed`. 

Not sure if this is confusing for users, what do you think?

PS. this is my first time dealing with react, please check if what I did makes sense

Snapshot Menu
Snapshot Menu

Node detail -> snapshot tab
Node detail

adejanovski (approved):
LGTM ✅ 

Thanks @averzicco !

PR #1116, opened 2021-08-30
Add 4.0.0 to the CI matrix

adejanovski:
Trunk can now be unstable with possible breaking changes again which means we have to make trunk tests optional and add 4.0.0 explicitly to the test matrix.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -125,11 +125,13 @@ jobs:
     name: H2 and Memory backends
     needs: build
     runs-on: ubuntu-20.04
+    continue-on-error: ${{ matrix.experimental }}
     strategy:
       matrix:
-        cassandra-version: [1.2.19, 2.0.17, 2.1.22, 2.2.19, 3.0.24, 3.11.10, 'github:apache/trunk']
+        cassandra-version: [1.2.19, 2.0.17, 2.1.22, 2.2.19, 3.0.25, 3.11.11, 4.0.0, 'github:apache/trunk']
         storage-type: [local]
         test-type: [ccm]
+        experimental: [false]
         include:
           # don't run against the following C* versions when using cassandra storage type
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -462,22 +483,32 @@ jobs:
...
         exclude:
           - storage-type: postgresql
             cassandra-version: 2.1.22
           - storage-type: postgresql
             cassandra-version: 2.2.19
           - storage-type: postgresql
-            cassandra-version: 3.0.24
+            cassandra-version: 3.0.25
+          - storage-type: postgresql
+            cassandra-version: 3.11.11
```
emerkle826, line 511: Do we want to exclude 3.11 here and only run against 4.0.0?

adejanovski: It's the case actually, on this line we're excluding the 3.11.11 run for the postgresql backend. What will remain is 4.0.0 only.

emerkle826 (approved):
This looks fine as long as we don't want to continue sidecar testing against 3.11.11 and only test against 4.0.0.

.github/workflows/ci.yaml
emerkle826: :+1:

adejanovski: Sorry, I misread.
Yes, that's intentional because we're already testing all version with the Cassandra backend, and there's no value in testing them all with the postgres backend.

adejanovski:
> This looks fine as long as we don't want to continue sidecar testing against 3.11.11 and only test against 4.0.0.

Sidecar is tested against all versions with the Cassandra backend, no worries there: 
Capture d’écran 2021-09-02 à 17 47 54

I've excluded some combinations with the postgres backend as the Cassandra version has no impact there, testing a single one is sufficient.

Thanks for the quick review!

PR #1117, opened 2021-09-01
JDK11 fixes

burmanm:
This is the changes so far. It compiles, but some tests fail still.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -46,7 +46,7 @@ jobs:
         uses: actions/setup-java@v2
         with:
           distribution: 'adopt' # See 'Supported distributions' for available options
-          java-version: '8.0.192'
+          java-version: '11'
 
       - name: Setup Node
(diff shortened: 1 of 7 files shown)


adejanovski:
Cool stuff! JDK11 support is something we really need now for Reaper.
I'll try to find some time to check the failing tests.

burmanm:
Why is it testing 1.4.1 jar?

```
java.lang.AssertionError: ClassNotFoundException for io.cassandrareaper.ReaperApplication in shaded versioned jarfile file:/home/michael/projects/git/datastax/cassandra-reaper/src/server/test-jars/cassandra-reaper-1.4.1.jar
```

I think these fail because JDK11 has stricter class loader and modules.

michaelsembwever:
> Why is it testing 1.4.1 jar?

Upgrade tests. They need fixing

adejanovski:
Merged in #1149

PR #1118, opened 2021-09-02
Fix the Base64-decoder failing if JWT_SECRET is not provided

### LINKED ISSUE #767: Reaper not starting if cluster name is shorter than 4 characters
Found in version: 1.4.6

Backend cassandra cluster with name "aa"

Cassandra reaper on start fails with exception:

```
java.lang.IllegalArgumentException: Empty key
        at javax.crypto.spec.SecretKeySpec.<init>(SecretKeySpec.java:96)
        at io.cassandrareaper.resources.auth.ShiroJwtProvider.getSigningKey(ShiroJwtProvider.java:63)
        at io.cassandrareaper.resources.auth.ShiroJwtProvider.<init>(ShiroJwtProvider.java:44)
```

To reproduce:

export JWT_SECRET=aa
start reaper -> exception

export JWT_SECRET=aabb
start reaper -> ok

The part bellow is responsible for this behaviour:

https://github.com/thelastpickle/cassandra-reaper/blob/842eb0286536407e97d03e70c795ea9caec7fc24/src/server/src/main/java/io/cassandrareaper/resources/auth/ShiroJwtProvider.java#L57-L63

And it will probably not work for postgress backend at all.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-408) by [Unito](https://www.unito.io)
┆Issue Number: K8SSAND-408

burmanm:
If the JWT_SECRET is not specified, rest of the signing options are expected to be in Base64-format (which they're not). This modification requires only the JWT_SECRET to be in Base64-format, rest are not decoded anymore.

This should also fix #767 since the shortest Base64 string is 4 characters. 

The Base64.getDecoder() is more strict and as such, this will test if those inputs are really Base64 or not (and current tests are enough to test this new behavior).

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/resources/auth/ShiroJwtProvider.java
+++ b/src/server/src/main/java/io/cassandrareaper/resources/auth/ShiroJwtProvider.java
@@ -22,7 +22,9 @@ import io.cassandrareaper.storage.CassandraStorage;
 import io.cassandrareaper.storage.PostgresStorage;
 
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 import java.security.Key;
+import java.util.Base64;
 
 import javax.crypto.spec.SecretKeySpec;
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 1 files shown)


michaelsembwever (dismissed):
LGTM. Thanks @burmanm

adejanovski (approved):

PR #1119, opened 2021-09-03
Allow editing of existing repair schedules

### LINKED ISSUE #1006: Allow to edit schedules in the UI
Schedules can't be updated, which means they have to be deleted and recreated to change a single setting.
We need to add that feature to improve the UX.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-247) by [Unito](https://www.unito.io)
┆Issue Number: K8SSAND-247

jdonenine:
Adds support within the UI for the editing of existing repair schedules.  Allows for the editing of a select set of fields as discussed in #1006 :

- days_between (which will force recomputing next_activation)
- intensity
- owner
- repair_parallelism
- segment_count_per_node

The editing is provided for via a new component that is decoupled from the schedules-list component so that it can potentially be re-used in other places:

image

The form itself includes all of the same fields that are present in the "Add Schedule" form, but most of them are presented in read-only fashion when used for editing (this is where the form component could later be enhanced to allow it to support edit and add workflows -- it doesn't do that now).

image

image

Edits are saved via a new API endpoint:  PATCH /repair_schedule/<id> that is provided here.

Fixes #1006

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/core/EditableRepairSchedule.java
+++ b/src/server/src/main/java/io/cassandrareaper/core/EditableRepairSchedule.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2021-2021 DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.cassandrareaper.core;
+
+import io.cassandrareaper.validators.NullOrNotBlank;
+import io.cassandrareaper.validators.ValidEditableRepairSchedule;
+
+import javax.validation.constraints.Max;
+import javax.validation.constraints.Min;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+import org.apache.cassandra.repair.RepairParallelism;
+
+@Data
+@AllArgsConstructor
+@NoArgsConstructor
+@ValidEditableRepairSchedule
+public class EditableRepairSchedule {
+  @NullOrNotBlank
+  protected String owner;
+
+  @JsonProperty(value = "repair_parallelism")
+  protected RepairParallelism repairParallelism;
+
+  @Min(value = 0)
+  @Max(value = 1)
+  protected Double intensity;
+
+  @JsonProperty(value = "scheduled_days_between")
+  @Min(value = 0)
+  @Max(value = 31)
+  protected Integer daysBetween;
+
+  @JsonProperty(value = "segment_count_per_node")
+  @Min(value = 1)
+  protected Integer segmentCountPerNode;
+}
--- a/src/server/src/main/java/io/cassandrareaper/core/EditableRepairSchedule.java
+++ b/src/server/src/main/java/io/cassandrareaper/core/EditableRepairSchedule.java
@@ -50,5 +50,6 @@ public class EditableRepairSchedule {
 
   @JsonProperty(value = "segment_count_per_node")
   @Min(value = 1)
+  @Max(value = 1000)
   protected Integer segmentCountPerNode;
 }
--- a/src/server/src/main/java/io/cassandrareaper/core/EditableRepairSchedule.java
+++ b/src/server/src/main/java/io/cassandrareaper/core/EditableRepairSchedule.java
@@ -23,12 +23,8 @@ import javax.validation.constraints.Max;
 import javax.validation.constraints.Min;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
-import lombok.Data;
-import lombok.NoArgsConstructor;
 import org.apache.cassandra.repair.RepairParallelism;
 
-@Data
-@NoArgsConstructor
 @ValidEditableRepairSchedule
 public class EditableRepairSchedule {
   @NullOrNotBlank
@@ -50,4 +46,52 @@ public class EditableRepairSchedule {
   @Min(value = 1)
   @Max(value = 1000)
   protected Integer segmentCountPerNode;
+
+  public EditableRepairSchedule() {
+    this.owner = null;
+    this.repairParallelism = null;
+    this.intensity = null;
+    this.daysBetween = null;
+    this.segmentCountPerNode = null;
+  }
+
+  public String getOwner() {
+    return owner;
+  }
+
+  public void setOwner(String owner) {
+    this.owner = owner;
+  }
+
+  public RepairParallelism getRepairParallelism() {
+    return repairParallelism;
+  }
+
+  public void setRepairParallelism(RepairParallelism repairParallelism) {
+    this.repairParallelism = repairParallelism;
+  }
+
+  public Double getIntensity() {
+    return intensity;
+  }
+
+  public void setIntensity(Double intensity) {
+    this.intensity = intensity;
+  }
+
+  public Integer getDaysBetween() {
+    return daysBetween;
+  }
+
+  public void setDaysBetween(Integer daysBetween) {
+    this.daysBetween = daysBetween;
+  }
+
+  public Integer getSegmentCountPerNode() {
+    return segmentCountPerNode;
+  }
+
+  public void setSegmentCountPerNode(Integer segmentCountPerNode) {
+    this.segmentCountPerNode = segmentCountPerNode;
+  }
 }
--- a/src/server/src/main/java/io/cassandrareaper/core/EditableRepairSchedule.java
+++ b/src/server/src/main/java/io/cassandrareaper/core/EditableRepairSchedule.java
@@ -43,7 +43,7 @@ public class EditableRepairSchedule {
   protected Integer daysBetween;
 
   @JsonProperty(value = "segment_count_per_node")
-  @Min(value = 1)
+  @Min(value = 0)
   @Max(value = 1000)
   protected Integer segmentCountPerNode;
 
--- a/src/server/src/main/java/io/cassandrareaper/core/EditableRepairSchedule.java
+++ b/src/server/src/main/java/io/cassandrareaper/core/EditableRepairSchedule.java
@@ -23,13 +23,11 @@ import javax.validation.constraints.Max;
 import javax.validation.constraints.Min;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
-import lombok.AllArgsConstructor;
 import lombok.Data;
 import lombok.NoArgsConstructor;
 import org.apache.cassandra.repair.RepairParallelism;
 
 @Data
-@AllArgsConstructor
 @NoArgsConstructor
 @ValidEditableRepairSchedule
 public class EditableRepairSchedule {
--- a/src/server/src/main/java/io/cassandrareaper/resources/RepairScheduleResource.java
+++ b/src/server/src/main/java/io/cassandrareaper/resources/RepairScheduleResource.java
@@ -20,6 +20,7 @@ package io.cassandrareaper.resources;
 import io.cassandrareaper.AppContext;
 import io.cassandrareaper.ReaperException;
 import io.cassandrareaper.core.Cluster;
+import io.cassandrareaper.core.EditableRepairSchedule;
 import io.cassandrareaper.core.PercentRepairedMetric;
 import io.cassandrareaper.core.RepairSchedule;
 import io.cassandrareaper.core.RepairUnit;
@@ -36,6 +37,9 @@ import java.util.List;
 import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
+import javax.validation.Valid;
+import javax.validation.constraints.NotNull;
+import javax.ws.rs.Consumes;
 import javax.ws.rs.DELETE;
 import javax.ws.rs.GET;
 import javax.ws.rs.POST;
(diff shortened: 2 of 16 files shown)


src/server/src/main/java/io/cassandrareaper/resources/RepairScheduleResource.java
```diff
@@ -76,6 +75,160 @@ public RepairScheduleResource(AppContext context) {
...
+            .segmentCountPerNode(segmentCountPerNode != null ? segmentCountPerNode : repairSchedule.getSegmentCountPerNode())
+            .build(repairSchedule.getId());
+  }
+
+  @PATCH
+  @Path("/{id}")
+  public Response patchRepairSchedule(
+          @Context UriInfo uriInfo,
+          @PathParam("id") UUID repairScheduleId,
+          @QueryParam("owner") Optional<String> owner,
```
jdonenine: @adejanovski I don't love passing these as query parameters and not as a resource in the body, but this is what was done for the PUT request that is currently used to update status, so I followed that model - was what was done for the other requests preferred or would you want to break from the convention?

src/ui/app/jsx/repair-schedule-form.jsx
```diff
@@ -0,0 +1,261 @@
...
+  _validate: function(state) {
+    if (!state) {
+      return false;
+    }
+
+    const clusterInvalid = !state.cluster || !state.cluster.length;
+    const keyspaceInvalid = !state.keyspace || !state.keyspace.length;
+    const ownerInvalid = !state.owner || !state.owner.length;
+    const scheduleTimeInvalid = (state.formType === "schedule" && (!state.startTime || !state.intervalDays)) || (state.formType === "repair" && (state.startTime || state.intervalDays));
+    // TODO What should the rules here be for the the combination of Datacenters and Nodes
```
jdonenine: I applied this logic from similar that I found in the add form, but wasn't sure how to treat datacenters and nodes.  The logic elsewhere looked like there were scenarios where you could have one but not the other?  I'll need some background on how to apply the validation here -- I could also simply exclude this logic for now since at the moment the form doesn't allow the user to edit these fields.

src/server/src/main/java/io/cassandrareaper/resources/RepairScheduleResource.java
```diff
@@ -76,6 +75,160 @@ public RepairScheduleResource(AppContext context) {
...
+    if (repairParallelismValid) {
+      try {
+        repairParallelismValid = RepairParallelism.valueOf(repairParallelism.toUpperCase()) != null ? true : false;
+      } catch (Exception e) {
+        repairParallelismValid = false;
+      }
+    }
+    boolean intensityValid = intensity != null && intensity >= 0.0D && intensity <= 1.0D;
+    // TODO What are good valid ranges for these values?
+    boolean scheduleDaysBetweenValid = scheduleDaysBetween != null && scheduleDaysBetween > 0;
```
jdonenine: Another validation question here for you @adejanovski what are reasonable boundaries for these?

src/server/src/main/java/io/cassandrareaper/resources/RepairScheduleResource.java
adejanovski: If the REST convention is to pass them in the body then I'm happy for you to do it, even if it doesn't follow what we've done elsewhere in the code (as any other OSS project, it's made of many peeps knowledge or lack of, so it's not necessarily done the right way).

src/ui/app/jsx/repair-schedule-form.jsx
adejanovski: Since those fields aren't editable (and won't be), my opinion is that you shouldn't bother with those checks.

src/server/src/main/java/io/cassandrareaper/resources/RepairScheduleResource.java
adejanovski: The interval can be 0, which means repair will run continuously, restarting right after it's finished. So I'd say `0 <= scheduleDaysBetweenValid <= 31`.
I can't think of a use case where you'd want to schedule a repair over a monthly basis (which is already super high if you ask me, the default being at least every 10 days).

src/server/src/main/java/io/cassandrareaper/resources/RepairScheduleResource.java
```diff
@@ -52,6 +51,7 @@
 
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Lists;
+import io.dropwizard.jersey.PATCH;
```
jdonenine, line 59: @adejanovski to make Checkstyle happy I had to go about changing a bunch of imports that IntelliJ had automatically organized, which I found a bit silly :-)

I've fixed things here in this PR, so we don't have to worry about it at the moment, but what would you think about removing this Checkstyle rule?

`<property name="customImportOrderRules" value="SAME_PACKAGE(2)###STANDARD_JAVA_PACKAGE###THIRD_PARTY_PACKAGE###STATIC"/>`

That enforces import ordering -- my feeling is that we should let the IDEs do their job and organize for us.  Or perhaps think about moving to something like https://github.com/google/google-java-format that can actively enforce formatting, would probably be nice to do similar in the UI code with prettier as well.  What do you think?

src/server/src/main/java/io/cassandrareaper/resources/RepairScheduleResource.java
jdonenine: Cool, I went ahead and made the change.  It has the other added benefit of simplifying the validation pieces and lets Dropwizard handle it by itself with the addition of a few custom validators.

it simplifies the resource code and you also get some nicer error responses from Dropwizard that can be useful, like:

```
PATCH http://127.0.0.1:8080/repair_schedule/da706f30-0c3e-11ec-be46-915dd6e16900
{
    "owner": "",
    "intensity": 0.5,
    "repair_parallelism": "DATACENTER_AWARE",
    "scheduled_days_between": 60,
    "segment_count_per_node": -10
}
```

```
422
{
    "errors": [
        "segmentCountPerNode must be greater than or equal to 1",
        "daysBetween must be less than or equal to 31",
        "owner must not be empty if it is provided"
    ]
}
```

Dropwizard provides that response automatically based on the validation rules configured for the class.

src/ui/app/jsx/repair-schedule-form.jsx
jdonenine: 👍 I went ahead and pulled this out.

src/server/src/main/java/io/cassandrareaper/resources/RepairScheduleResource.java
jdonenine: Perfect, thanks.  I've updated the new validations with the rules you mention.

src/server/src/main/java/io/cassandrareaper/resources/RepairScheduleResource.java
adejanovski: I understand your frustration here but I'm reluctant to change the formatting rules to rely on IDEs instead. They will most probably apply different orderings which will create some noise in PRs as there's gonna be some unnecessary line changes.

Instead, we could maybe make it so that the rule matches IntelliJ's ordering of imports: 

```
To configure the check so that it matches default IntelliJ IDEA formatter configuration (tested on v14):

group of static imports is on the bottom
groups of non-static imports: all imports except of "javax" and "java", then "javax" and "java"
imports will be sorted in the groups
groups are separated by one blank line
Note: "separated" option is disabled because IDEA default has blank line between "java" and static imports, and no blank line between "javax" and "java"

<module name="CustomImportOrder">
  <property name="customImportOrderRules"
    value="THIRD_PARTY_PACKAGE###SPECIAL_IMPORTS###STANDARD_JAVA_PACKAGE###STATIC"/>
  <property name="specialImportsRegExp" value="^javax\."/>
  <property name="standardPackageRegExp" value="^java\."/>
  <property name="sortImportsInGroupAlphabetically" value="true"/>
  <property name="separateLineBetweenGroups" value="false"/>
</module>
```

this was pasted from the checkstyle docs: https://checkstyle.org/config_imports.html#ImportOrder_Examples

Apparently there's some workaround for the java/javax imports which can't be dealt with appropriately by IntelliJ: https://stackoverflow.com/a/64243861/5106246

src/server/src/main/java/io/cassandrareaper/core/EditableRepairSchedule.java
```diff
@@ -0,0 +1,53 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+import org.apache.cassandra.repair.RepairParallelism;
+
+@Data
+@NoArgsConstructor
+@ValidEditableRepairSchedule
+public class EditableRepairSchedule {
```
adejanovski, line 29: We've discussed using Lombok in the early stages of our Reaper fork, but as interesting as it looked we ended up using the Builder pattern as convention for our POJOs and avoiding to use Lombok as it relies on IDE plugins.
While it's fine to reopen that conversation, I think it would be good to be consistent with the existing (and I reckon, undocumented) conventions currently in use.

src/server/src/main/java/io/cassandrareaper/core/EditableRepairSchedule.java
```diff
@@ -0,0 +1,53 @@
...
+  @Max(value = 1)
+  protected Double intensity;
+
+  @JsonProperty(value = "scheduled_days_between")
+  @Min(value = 0)
+  @Max(value = 31)
+  protected Integer daysBetween;
+
+  @JsonProperty(value = "segment_count_per_node")
+  @Min(value = 1)
```
adejanovski: I was wrong in suggesting 1 as minimum value: if set to 0 it currently translates to the default number of segments which is applied when the repair run is created.
Currently it prevents from editing a schedule that had 0 segment per node if the value isn't set explicitly (which is then not consistent with the initial schedule creation).
You can remove the minimum bound to allow 0 as valid.
My bad for this.

src/server/src/main/java/io/cassandrareaper/resources/RepairScheduleResource.java
```diff
@@ -76,6 +82,62 @@ public RepairScheduleResource(AppContext context) {
...
   }
 
+  @PATCH
+  @Produces(MediaType.APPLICATION_JSON)
+  @Consumes(MediaType.APPLICATION_JSON)
+  @Path("/{id}")
+  public Response patchRepairSchedule(
+      @Context UriInfo uriInfo,
+      @PathParam("id") UUID repairScheduleId,
+      @NotNull @Valid EditableRepairSchedule editableRepairSchedule
```
adejanovski, line 92: The validators are a very nice addition!

src/ui/app/jsx/repair-schedule-form.jsx
```diff
@@ -0,0 +1,274 @@
...
+      ) ||
+      (state.formType === "repair" &&
+        (state.startTime ||
+          state.intervalDays
+        )
+      );
+    const segmentsPerNodeInvalid = !state.segments ||
+      state.segments == undefined ||
+      state.segments == null ||
+      state.segments < 1 ||
```
adejanovski: We can relax this constraint as mentioned in my previous comment and allow 0 as segment count per node.

adejanovski (changes requested):
That's an impressively crafted PR @jdonenine 👏 
Lots of testing and added safety through the use of validators are very much appreciated.

I'm concerned that adding Lombok to the project requires a wider conversation with other contributors and I think we should keep it out of the scope of this PR.
I mislead you with setting the lower bound to 1 for the segment count per node. 0 is very much valid and will translate into the default number of segments (which should be 64 in the default configuration).

src/server/src/main/java/io/cassandrareaper/core/EditableRepairSchedule.java
jdonenine: No problem, at this point I don't think I'd worry much about the IDE aspect, it's so popular I believe that all the major IDEs support it well, but I also wasn't sure if there was a process to go through about the addition of a dependency overall and licensing and what not, certainly easier, since the code it replaces is trivial to just take it out for now.

Recent comment removes it, so we can handle it separately if we want it in.

src/server/src/main/java/io/cassandrareaper/core/EditableRepairSchedule.java
jdonenine: No worries at all, I updated the min to be 0 in the API and the UI.

adejanovski: cool, thanks!

src/server/src/main/java/io/cassandrareaper/core/EditableRepairSchedule.java
adejanovski: sounds good, thanks

src/ui/app/jsx/repair-schedule-form.jsx
```diff
@@ -110,7 +110,7 @@ const RepairScheduleForm = CreateReactClass({
     const segmentsPerNodeInvalid = !state.segments ||
       state.segments == undefined ||
       state.segments == null ||
-      state.segments < 1 ||
+      state.segments < 0 ||
```
adejanovski, line 112: The segment count per node is considered invalid by default because it's initialized as `''` instead of `0` here: https://github.com/thelastpickle/cassandra-reaper/pull/1119/commits/ec5580723fa0dbf813547c1a04ca391856222830#diff-166c0877429b0e6dad5328286ab5af08f46ff06c0ee1a048483f52ffa5dd9d3eR43

This results in the `Save` button being disabled unless the value is changed.
Could you make it so that 0 is the default value and the `Save` button is enabled when editing without making a change?

adejanovski (changes requested):
Aside from one additional change to make it so that the `Save` button is enabled by default in the edit schedule modal, it looks good to me 👍

src/ui/app/jsx/repair-schedule-form.jsx
jdonenine: It's definitely true that the Save button is disabled until a value is changed, but not because of the validity of the value.  The validation check is only triggered on changes to the form, so that you can't save an non-dirty form so to speak.  It will only let you save a form that has been changed -- although it's not a deep inspection that compares each value over and over again, just for the record.

But, that aside, I do notice that I have the validation wrong, I shouldn't be checking !state.segments, because it's a number and that could screw up, I will fix that now.

jdonenine: I also failed to check the intensity value there, I've added that as well in the new commit.

adejanovski (approved):
It looks great now @jdonenine and works as expected!
Could you please merge the PR using the "Squash and merge" button and rework the commit message in the GitHub UI so that it nicely describes the main changes?
Thanks!

jdonenine:
Merged, thanks for all the help and tips @adejanovski

PR #1121, opened 2021-09-03
Switch to ubuntu 20.04 and use nodejs 10.x for docker builder

Skunnyk:
- Based on what is already done on CI: use of ubuntu 20.04 LTS and nodejs
  10.x (looks like the ui part does not build with more recent version of nodejs)
- Simplify a bit the Dockerfile by removing useless lines


adejanovski (approved):
Works for me 👍 

Thanks @Skunnyk !

Issue #1071, opened 2021-05-20
When running repair for the same keyspace - prometheusMetrics are not updated


DzeryCZ:
I have scheduled repairs on multile keyspaces to run every week. 
The first run of every keyspace can be montored via `/prometheusMetrics` (grafana), but when the run finishes and a new schedule takes place, the metrics shows the previous run => `io_cassandrareaper_service_RepairRunner_repairProgress=1`.
When I restart `cassandra-reaper` container the `/prometheusMetrics` picks up the new runs.

I've tried to run version `2.2.4` and also `master` of `cassandra-reaper`.

Any tips, how to resolve this issue?

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-200) by [Unito](https://www.unito.io)
┆Issue Number: K8SSAND-200

shaotingcheng:
I have the same problem with V2.2.3
It seems the same as #959 with V2.0.5

lapo-luchini:
Will this be in 2.3.2, thus?
Is current trunk "stable"?
(I could consider building it myself)

adejanovski:
It will be in 3.0 actually, which will be released within a few days. Current `master` is considered as working... until bugs are found :) Right now it should be in a solid state as we approach release time.
You do not need to build it yourself as it is built automatically for each push to the `master` branch. You can find beta packages [here](https://cloudsmith.io/~thelastpickle/repos/reaper-beta/packages/), and can use the `master` tag of [the docker image](https://hub.docker.com/layers/thelastpickle/cassandra-reaper/master/images/sha256-90ca0361b42e4c636705983d43a906e1e6602bfa5601a955576f95eab7dd3c64?context=repo).

lapo-luchini:
Ohhh, cool, I didn't notice that. I extracted `cassandra-reaper-3.0.0-BETA-20211015.jar` from the DEB file and am testing it on FreeBSD. Thanks.

lapo-luchini:
@DzeryCZ this should be closed, I think?
(it is on my own tests)

DzeryCZ:
Yes, the functionality works as expected. 

Many thanks.

PR #1122, opened 2021-09-03
repairProgress metrics are updated on subsequent runs.

### LINKED ISSUE #959: repairProgress metric not updating for subsequent runs.
This may be something I am missing, however, I cannot find a reason for it. The first repair of `FILTEREDKEYSPACENAME` properly shows the repairProgress gauge of 0.0-1.0. Subsequent repairs is never updated (while using manually triggered repairs in cassandra reaper). I searched inside `/prometheusMetrics` however the only repairProgress I see is for the old run id with a 1 (100%) value. Which clearly is not true of the active running repair.  Looking in `/metrics?pretty=true` is the same result.

[Docs](http://cassandra-reaper.io/docs/metrics/dashboards/) pointing to using this metric 

Using cassandra-reaper 2.0.5

Paste of the raw metrics which is the only existent one for `io_cassandrareaper_service_RepairRunner_repairProgress_c5134_FILTEREDKEYSPACENAME`

```
# HELP io_cassandrareaper_service_RepairRunner_repairProgress_c5134_FILTEREDKEYSPACENAME_8c21ab30dccb11ea828babd573712197 Generated from Dropwizard metric import (metric=io.cassandrareaper.service.RepairRunner.repairProgress.c5134.FILTEREDKEYSPACENAME.8c21ab30dccb11ea828babd573712197, type=io.cassandrareaper.service.RepairRunner$$Lambda$249/274061983)
# TYPE io_cassandrareaper_service_RepairRunner_repairProgress_c5134_FILTEREDKEYSPACENAME_8c21ab30dccb11ea828babd573712197 gauge
io_cassandrareaper_service_RepairRunner_repairProgress_c5134_FILTEREDKEYSPACENAME_8c21ab30dccb11ea828babd573712197 1.0
```

(cluster/keyspace/runid obfuscated)

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-456) by [Unito](https://www.unito.io)
┆Issue Number: K8SSAND-456

Skunnyk:
- When subsequent runs are executed on the same cluster/keyspace,
 exposed metrics are not updated.
- Unregistering/registering the metric at the start of a run repair
  triggers the metrics update.
- This fixes #959 and #1071

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
+++ b/src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
@@ -157,9 +157,10 @@ final class RepairRunner implements Runnable {
   }
 
   private void registerMetric(String metricName, Gauge<?> gauge) {
-    if (!context.metricRegistry.getMetrics().containsKey(metricName)) {
-      context.metricRegistry.register(metricName, gauge);
+    if (context.metricRegistry.getMetrics().containsKey(metricName)) {
+      context.metricRegistry.remove(metricName);
     }
+    context.metricRegistry.register(metricName, gauge);
   }
 
   UUID getRepairRunId() {


adejanovski (approved):
I didn't know we'd need to recreate the metric so it gets updated again. Thanks for fixing this 🙏

PR #1123, opened 2021-09-08
Remove H2 and Postgres storage backends

### LINKED ISSUE #1120: Remove support for storage backends other than Cassandra compatible ones
Reaper was initially implemented with memory storage for dev and postgres for production use.
Later on when TLP forked the project, the Cassandra backend was added as it looked like a better fit for how Reaper was evolving.
Also, if you're using Reaper, you'll have at least one Cassandra cluster up an running already.
As features were added and as Reaper grew in complexity with the notable addition of the distributed modes (including sidecar), it got more challenging to maintain all the supported backends.
Some features require different strategies to be implemented for each backend, diluting the efforts and making it longer to ship features.
It often resulted in features that worked fine with a specific backend and were buggy with others.
Another impact is that we're not optimizing the Cassandra backend as much as we could if it was the only one we could focus on.
Working on OSS projects, while rewarding and exciting, is often done on free time on top of actual work assignements and should be made as easy as possible.
Starting with Reaper 3.0, the memory, H2 and Postgres backends will be removed to simplify the codebase and allow faster development of new features while ensuring higher quality.
The 2.3 branch will continue receiving bug/performance fixes if appropriate, but new feature development will be done on the master branch only.
The use of an embedded Cassandra server (storing data on disk) is under evaluation as an option for single reaper instance deployments in production or for testing purposes.

adejanovski:
Fixes #1120


emerkle826:
I like this PR already, and I haven't even started reviewing it yet :smile:

emerkle826 (dismissed):
This looks good to me. It's still marked Draft, so I'm not sure if you are still working on things, but this is good stuff to cut down on the maintenance.

PR #1127, opened 2021-10-07
Add schedule triggers for incremental repairs based on % unrepaired metrics

### LINKED ISSUE #1079: Make incremental schedules trigger on %unrepaired threshold
Incremental repair schedules should trigger either based on a percentage of unrepaired data or using fixed intervals.
A new field would be added to store the percent repaired threshold.
The UI would need to accomodate for both possibilities, allowing only one trigger out of the two and restricting the use of the percent based threshold to incremental repairs.

The scheduler would leverage %repaired metrics collected to trigger incremental repairs (#1057).

Recreation of original issue #1058

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-475) by [Unito](https://www.unito.io)
┆Epic: Incremental Repair
┆Fix Versions: reaper-2.3.0,k8ssandra-1.4.0
┆Issue Number: K8SSAND-475
┆Priority: Medium

adejanovski:
Fixes #1079 

The first commit in this PR implements a new trigger for schedules of incremental repairs, which is by percentage of unrepaired data. During incremental repairs, sstables are marked as repaired which allows Cassandra to keep track of how much data is repaired per table. The current implementation will collect % repaired metrics during heart beats, and retain the lowest percent repaired from all the tables that the repair definition targets. The collection of these metrics in the `Heart` class was slightly reworked to list incremental schedules only if a %repaired metric collection is allowed (enough time has passed since the previous collection). This should save some unnecessary database calls.
A new field was introduced in the `EditableRepairSchedule` class to set the threshold of unrepaired data over which repair should start. This field can only be set for incremental repairs and is complimentary to the interval in days (we still need to run repair on a regular basis even if we don't reach the threshold).
Another field was introduced to store the last repair run id for the schedule. This is used to get the endTime of the previous run, and give enough time for the % repaired metric to be refreshed before triggering a new run (two metrics extraction cycles will be required before another run can start).
The repair run history of schedules is not updated anymore as it was creating very large collections in some extreme cases and wasn't used anyway (the column isn't displayed in the UI). 
Concurrency when multiple reaper instances are competing to start schedules wasn't properly dealt with so far which could result in duplicates (especially in sidecar mode). The Reaper instance id reported in the `running_reapers` table is now used to naturally designate a scheduling leader which will be the only instance that can start a schedule. Instance IDs are sorted and the first in the list is the leader. In case of a restart, rows in the `running_reapers` table have a 3 minutes TTL which will block schedules from triggering for this duration at most. This allowed to remove [this block](https://github.com/thelastpickle/cassandra-reaper/pull/1127/files#diff-b4dc2f125d9733f96e49b6318494c669a702bc1e2e3ba806bb7b5fba5248e768L151-L207) of code that dealt with scheduling conflicts afterwards.
The `RepairRunService` class was modified to allow proper mocking and the `SchedulingManager` class was reworked so that a `RepairRunService` mock could be injected during tests.

The UI and documentations were updated accordingly to the changes of this PR.

In a second commit, the test matrix is simplified as it's not necessary to test flapping reapers with all versions of Cassandra. We're not testing Cassandra specifics in these tests but rather Reaper's resiliency. This should reduce the flakiness of CI.


src/server/src/test/java/io/cassandrareaper/service/SchedulingManagerTest.java
```diff
@@ -0,0 +1,340 @@
...
+    reaperInstances.add(context.reaperInstanceId);
+    context.storage = mock(CassandraStorage.class);
+    when(((CassandraStorage)context.storage).getRunningReapers()).thenReturn(reaperInstances);
+    SchedulingManager schedulingManager = SchedulingManager.create(context);
+    assertFalse("The eldest Reaper instance should be the scheduling leader",
+        schedulingManager.currentReaperIsSchedulingLeader());
+  }
+
+  @Test
+  public void lastRunRunIsOldEnoughTest() {
```
emerkle826: nit: should be named `lastRepairRunIsOldEnoughTest`

src/server/src/test/java/io/cassandrareaper/service/SchedulingManagerTest.java
```diff
@@ -0,0 +1,340 @@
...
+        .segmentCountPerNode(10)
+        .lastRun(repairRun.getId())
+        .percentUnrepairedThreshold(5)
+        .build(UUIDs.timeBased());
+    assertTrue("Enough time must pass since last run to trigger a new one",
+        schedulingManager.lastRunRunIsOldEnough(repairSchedule));
+  }
+
+  @Test
+  public void lastRunRunIsNotOldEnoughTest() {
```
emerkle826: nit: same here, `RunRun` should be `RepairRun`

emerkle826 (dismissed):
This looks good to me. Just a couple of nitpick renames and it's good to go!

adejanovski:
Thanks @emerkle826! I fixed the method names and merged the PR.

PR #1131, opened 2021-10-19
Allow to force create repair runs with conflicting units

adejanovski:
Similar to what was done with schedules, this PR builds on top of that previous work to allow force creation of repair runs.


michaelsembwever (approved):
LGTM. but only skimmed through it.

adejanovski:
I'll take it 😉 

Thanks!

PR #1113, opened 2021-08-24
Mount path change for config

jeffbanks:
Support reaper image permissions being applied in k8ssandra security context configurations.

* The Reaper config file is stored at /etc/cassandra-reaper.yml
* Moving to /etc/reaper/cassandra-reaper.yml
* [k8ssand-701](https://k8ssandra.atlassian.net/browse/K8SSAND-701)

PR #1133, opened 2021-10-26
Additional changes for /etc/reaper paths

jeffbanks:
Additional changes related to PR #1113, which change reference from `/etc/cassandra-reaper` to `/etc/reaper` to support applied security.

### DIFF
--- a/src/packaging/bin/cassandra-reaper
+++ b/src/packaging/bin/cassandra-reaper
@@ -45,15 +45,15 @@ if [ -z "$CLASS_PATH" ]; then
 fi
 
 if [ $# -eq 0 ]; then
-    CONFIG_PATH="/usr/local/etc/cassandra-reaper/cassandra-reaper.yaml"
+    CONFIG_PATH="/usr/local/etc/reaper/cassandra-reaper.yaml"
     if [ ! -e "$CONFIG_PATH" ]; then
-        CONFIG_PATH="/etc/cassandra-reaper/cassandra-reaper.yaml"
+        CONFIG_PATH="/etc/reaper/cassandra-reaper.yaml"
     fi
 else
     CONFIG_PATH="$@"
 fi
 
-SSL_CONFIG_PATH="/etc/cassandra-reaper/cassandra-reaper-ssl.properties"
+SSL_CONFIG_PATH="/etc/reaper/cassandra-reaper-ssl.properties"
 if [ -r "$SSL_CONFIG_PATH" ]; then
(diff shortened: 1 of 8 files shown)


jeffbanks:
@jsanda @alex Additional changes related to `reaper-operator` modifications [reaper-operator-PR#68](https://github.com/k8ssandra/reaper-operator/pull/68).  

We will need to reference updates for a new reaper image to have `reaper-operator` pass its e2e tests.

src/packaging/bin/cassandra-reaper
```diff
@@ -45,15 +45,15 @@ if [ -z "$CLASS_PATH" ]; then
 fi
 
 if [ $# -eq 0 ]; then
-    CONFIG_PATH="/usr/local/etc/cassandra-reaper/cassandra-reaper.yaml"
+    CONFIG_PATH="/usr/local/etc/reaper/cassandra-reaper.yaml"
```
adejanovski, line 48: This is going to affect users that have non docker based installations of Reaper. I'm a little nervous with changing paths like this as folks might not notice the change and lose their custom settings files.

What do you think the upgrade path would be here for debian/rpm users?

jeffbanks: @adejanovski Not good.   

Yes, if that is the case, we may want leave it using the standard `cassandra-reaper` for the `/usr/local/etc` root.  This has been changed only for naming consistency with the `/etc/reaper` change.  I am good reversing this back as we aren't trying to change anything for `/usr/local/etc`.  Can you think of any other areas that may be an issue?

adejanovski: Any path change that's not happening within the Docker container means that upgrades will be problematic. Sadly, that's pretty much everything that's in this PR 😕 
I don't think we can go through with it without breaking upgrades, I'm sorry.

jeffbanks: @adejanovski After discussing this change, I agree let's not introduce instability of the existing path name with a trivial name change.  

I will make adjustments in the respective areas of this repository (different PR) as well as the reaper-operator where the name was changed from `cassandra-reaper` to `reaper`.

Closing this one as we don't want this merged.

jeffbanks:
Closing this one as the name change may introduce too much instability and confusion.  For security permissions being applied which introduced this PR, the `/etc/cassandra-reaper` will be the root path for volumePaths identified in the `reaper-operator`.  This ensures that users of Reaper outside of a container will not be impacted with a name change.

PR #1138, opened 2021-11-09
Fix adaptive value when editing schedules

### LINKED ISSUE #1137: Repair schedules don't keep adaptive setting
After creating a repair schedule in the UI in which I enable adaptive repair, if I go to edit the repair it shows that Adaptive is `false`!

**Spotted on 3.0.0 (ae0c6af)**

https://user-images.githubusercontent.com/24530198/140812330-0f1380fb-cc51-48a4-8be5-1274b14f4538.mov

adejanovski:
Fixes #1137 

The current value of the `adaptive` field wasn't properly displayed when editing schedules.


jdonenine (dismissed):
Looks good to me.  I was able to test out making changes to the flag on multiple schedules and got the expected behavior.

jdonenine (approved):

PR #1146, opened 2021-11-15
Add missing auto-scheduling environment variables

adutra:


src/packaging/docker-services/reaper/reaper.env
```diff
@@ -46,6 +46,9 @@ REAPER_AUTO_SCHEDULING_INITIAL_DELAY_PERIOD="PT30S"
 REAPER_AUTO_SCHEDULING_PERIOD_BETWEEN_POLLS="PT20M"
 REAPER_AUTO_SCHEDULING_TIME_BEFORE_FIRST_SCHEDULE="PT10M"
 REAPER_AUTO_SCHEDULING_SCHEDULE_SPREAD_PERIOD="PT3H"
+REAPER_AUTO_SCHEDULING_ADAPTIVE=true
+REAPER_AUTO_SCHEDULING_INCREMENTAL=true
+REAPER_AUTO_SCHEDULING_PERCENT_UNREPAIRED_THRESHOLD=30
```
adejanovski:
```suggestion
REAPER_AUTO_SCHEDULING_INCREMENTAL=false
REAPER_AUTO_SCHEDULING_PERCENT_UNREPAIRED_THRESHOLD=10
```

src/server/src/main/docker/Dockerfile
```diff
@@ -38,6 +38,9 @@ ENV REAPER_SEGMENT_COUNT_PER_NODE=64 \
     REAPER_AUTO_SCHEDULING_PERIOD_BETWEEN_POLLS=PT10M \
     REAPER_AUTO_SCHEDULING_TIME_BEFORE_FIRST_SCHEDULE=PT5M \
     REAPER_AUTO_SCHEDULING_SCHEDULE_SPREAD_PERIOD=PT6H \
+    REAPER_AUTO_SCHEDULING_ADAPTIVE=false \
+    REAPER_AUTO_SCHEDULING_INCREMENTAL=true \
```
adejanovski:
```suggestion
    REAPER_AUTO_SCHEDULING_ADAPTIVE=true \
    REAPER_AUTO_SCHEDULING_INCREMENTAL=false \
```

If you don't mind, I think we should set defaults to adaptive=true and incremental=false. Incremental repairs are only safe with Cassandra 4.0+, so I wouldn't want to set defaults that can create chaos.

adejanovski (changes requested):
Thanks for the PR @adutra 🙏 
Sorry I forgot to wire these settings into the docker container...

I would just like to use safer defaults, and then we're good to go. Also, adaptive schedules can't be incremental as well (it's a full repair feature).

adejanovski (approved):

PR #1148, opened 2021-11-24
Upgrade to Dropwizard 2.0

adejanovski:
Upgrade dropwizard and other dependencies that were affected by CVEs (including shiro, netty, snakeyaml, ...).
These changes led to removing the reflection code used in testing and rely instead on dropwizard's provided test classes.

I've also made the storage object managed by Dropwizard so that the reaper instance entry could be removed from the `running_reapers` table on shutdown. This improves scheduling management as since 3.0 only the first running reaper in the list can start scheduled runs.

These changes should also help with JDK11 support.


src/server/pom.xml
```diff
@@ -312,6 +332,11 @@
             <version>2.1.7.597</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <version>3.21.0</version>
```
adejanovski, line 343: Scope test?

src/server/pom.xml
```diff
@@ -643,7 +668,7 @@
                     <io.cassandrareaper.service.SegmentRunner.sleep_time_after_postpone_in_ms>500</io.cassandrareaper.service.SegmentRunner.sleep_time_after_postpone_in_ms>
                   </systemPropertyVariables>
                   <!-- noverify is needed for ReaperTestJettyRunner.ParentLastURLClassLoader.ChildURLClassLoader.findClass(..) -->
-                  <argLine>${argLine} -noverify</argLine>
+                  <!-- <argLine>${argLine} -noverify</argLine> -->
```
adejanovski: cleanup

src/server/src/test/java/io/cassandrareaper/acceptance/BasicSteps.java
```diff
@@ -1874,6 +1874,7 @@ public void aResponseIsReturned(String statusDescription) throws Throwable {
   @Then("^the response was redirected to the login page$")
   public void theResponseWasRedirectedToTheLoginPage() throws Throwable {
     assertTrue(lastResponse.hasEntity());
+    LOG.error("HTTP response entity : {}", lastResponse.readEntity(String.class));
```
adejanovski: for debugging, remove

src/server/src/test/java/io/cassandrareaper/acceptance/ReaperCassandraIT.java
```diff
@@ -106,17 +107,24 @@ public void upgradeReaperRunner(Optional<String> version) throws InterruptedExce
...
+    ReaperTestJettyRunner runner;
+    try {
+
+      runner = new ReaperTestJettyRunner(CASS_CONFIG_FILE);
+      RUNNER_INSTANCES.add(runner);
+      Thread.sleep(100);
+      BasicSteps.addReaperRunner(runner);
+    } catch (RuntimeException e) {
+      LOG.error("Failed creating Test Jetty Runner", e);
+      e.printStackTrace();
```
adejanovski: remove please

src/server/src/test/resources/cassandra-reaper-access-control-enabled-at.yaml
```diff
@@ -39,11 +39,11 @@ server:
   type: default
   applicationConnectors:
     - type: http
-      port: 8083
+      port: 8090
```
adejanovski: remove

src/server/pom.xml
```diff
@@ -29,7 +29,8 @@
     <packaging>jar</packaging>
 
     <properties>
-        <dropwizard.version>1.3.14</dropwizard.version>
+        <dropwizard.version>2.0.25</dropwizard.version>
+        <jersey.version>2.33</jersey.version>
```
adutra, line 33: Could be upgraded to 2.35.

src/server/pom.xml
```diff
@@ -49,14 +50,15 @@
...
+        <shiro-version>1.8.0</shiro-version>
     </properties>
 
     <dependencies>
         <!--compile scope -->
         <dependency>
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
-            <version>23.0</version>
+            <version>24.1.1-jre</version>
```
adutra, line 63: This is unfortunately the last version that still has the old `Futures.addCallback` method without a 3rd argument (the executor), which seems used in Reaper codebase. We might want to upgrade to more recent versions of Guava later on (Guava is a frequent source of CVEs), but that would require code changes.

src/server/pom.xml
```diff
@@ -49,14 +50,15 @@
...
         </dependency>
         <dependency>
             <groupId>ch.qos.logback</groupId>
```
adutra, line 66: Logback can be upgraded to 1.2.7.

src/server/pom.xml
```diff
@@ -88,11 +90,6 @@
             <artifactId>dropwizard-client</artifactId>
```
adutra, line 92: Dropwizard makes a heavy usage of Jackson library – another frequent source of CVEs. Dropwizard 2.0.25 depends on Jackson 2.10.3 which is already old. I strongly recommend forcing Jackson version to the latest: 2.13.0.

This can be done by adding the following snippet to this pom:

```
    <dependencyManagement>
        <dependencies>
            <dependency>
                <groupId>com.fasterxml.jackson</groupId>
                <artifactId>jackson-bom</artifactId>
                <version>2.13.0</version>
                <type>pom</type>
                <scope>import</scope>
            </dependency>
        </dependencies>
    </dependencyManagement>
```

src/server/pom.xml
```diff
@@ -146,17 +143,39 @@
...
+                <exclusion>
+                    <groupId>javax.validation</groupId>
+                    <artifactId>validation-api</artifactId>
+                </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
+            <groupId>org.yaml</groupId>
+            <artifactId>snakeyaml</artifactId>
+            <version>1.26</version>
```
adutra: Can be upgraded to 1.29.

src/server/pom.xml
```diff
@@ -146,17 +143,39 @@
...
 		            <groupId>com.fasterxml.jackson.core</groupId>
 	                <artifactId>jackson-databind</artifactId>
 	            </exclusion>
+                <exclusion>
+                    <groupId>io.netty</groupId>
+                    <artifactId>netty-handler</artifactId>
+                </exclusion>
 	        </exclusions>
         </dependency>
         <dependency>
```
adutra, line 187: * `cassandra-migration` can be upgraded to 2.4.0.
* All prometheus artifacts can be upgraded to 0.12.0.

src/server/pom.xml
```diff
@@ -180,52 +199,59 @@
...
-            <groupId>org.glassfish.jersey.core</groupId>
-            <artifactId>jersey-server</artifactId>
-            <version>2.25.1</version>
-        </dependency>
-        <dependency>
-            <!-- required to solve dropwizard 1.1.7 & jersery-client 2.25.1 dependency issues -->
-          <groupId>org.glassfish.jersey.bundles.repackaged</groupId>
-          <artifactId>jersey-guava</artifactId>
-          <version>2.25.1</version>
+            <version>2.33</version>
```
adutra:
```suggestion
            <version>${jersey.version}</version>
```

src/server/pom.xml
```diff
@@ -180,52 +199,59 @@
...
         </dependency>
         <dependency>
             <groupId>org.glassfish.jersey.media</groupId>
             <artifactId>jersey-media-sse</artifactId>
-            <version>2.25.1</version>
+            <version>2.33</version>
```
adutra:
```suggestion
            <version>${jersey.version}</version>
```

src/server/pom.xml
```diff
@@ -180,52 +199,59 @@
...
         </dependency>
         <dependency>
             <!-- required to solve dropwizard 1.1.7 & jersery-client 2.25.1 dependency issues -->
```
adutra: The comment seems outdated, maybe remove it or rephrase it? That said, I did try removing this dependency, and the build failed. So it is definitely still required.

src/server/pom.xml
```diff
@@ -769,4 +794,29 @@
...
+            <enabled>false</enabled>
+            </releases>
+            <snapshots>
+            <enabled>true</enabled>
+            </snapshots>
+        </repository>
+    </repositories>
+    <pluginRepositories>
+        <pluginRepository>
+            <id>plugin-snapshots-repo</id>
```
adutra: Do we really need a plugin snapshot repository? I don't see any plugin sporting a snapshot version.

src/server/src/test/java/io/cassandrareaper/acceptance/ReaperCassandraEachIT.java
```diff
@@ -78,7 +77,8 @@ public void upgradeReaperRunner(Optional<String> version) throws InterruptedExce
   }
 
   private static void createReaperTestJettyRunner(Optional<String> version) throws InterruptedException {
```
adutra: The parameter `version` is now unused, please remove.

src/server/src/test/java/io/cassandrareaper/acceptance/ReaperCassandraIT.java
```diff
@@ -106,17 +107,24 @@ public void upgradeReaperRunner(Optional<String> version) throws InterruptedExce
   }
 
   private static void createReaperTestJettyRunner(Optional<String> version) throws InterruptedException {
```
adutra: Same here: unused parameter.

src/server/src/test/java/io/cassandrareaper/acceptance/ReaperCassandraSidecarIT.java
```diff
@@ -83,11 +82,17 @@ public void upgradeReaperRunner(Optional<String> version) throws InterruptedExce
     }
   }
 
+  @SuppressWarnings("IllegalCatchCheck")
   private static void createReaperTestJettyRunner(Optional<String> version) throws InterruptedException {
```
adutra: And same here: unused parameter.

src/server/src/test/java/io/cassandrareaper/acceptance/ReaperCassandraSidecarIT.java
```diff
@@ -83,11 +82,17 @@ public void upgradeReaperRunner(Optional<String> version) throws InterruptedExce
...
+      runner = new ReaperTestJettyRunner(CASS_CONFIG_FILE[RUNNER_INSTANCES.size()]);
+      RUNNER_INSTANCES.add(runner);
+      Thread.sleep(100);
+      BasicSteps.addReaperRunner(runner);
+    } catch (RuntimeException e) {
+      e.printStackTrace();
+    }
   }
 
   private static void removeReaperTestJettyRunner(ReaperTestJettyRunner runner) throws InterruptedException {
```
adutra: This method seems unused.

src/server/src/test/java/io/cassandrareaper/acceptance/ReaperCassandraSidecarIT.java
```diff
@@ -83,11 +82,17 @@ public void upgradeReaperRunner(Optional<String> version) throws InterruptedExce
     }
```
adutra: A few unused fields in this class FYI.

src/server/src/test/java/io/cassandrareaper/acceptance/ReaperCassandraSidecarIT.java
```diff
@@ -83,11 +82,17 @@ public void upgradeReaperRunner(Optional<String> version) throws InterruptedExce
...
-    Thread.sleep(100);
-    BasicSteps.addReaperRunner(runner);
+    ReaperTestJettyRunner runner;
+    try {
+      runner = new ReaperTestJettyRunner(CASS_CONFIG_FILE[RUNNER_INSTANCES.size()]);
+      RUNNER_INSTANCES.add(runner);
+      Thread.sleep(100);
+      BasicSteps.addReaperRunner(runner);
+    } catch (RuntimeException e) {
+      e.printStackTrace();
```
adutra: Remove stack trace to stderr.

adutra (commented):
Reviewed almost everything so far, will resume tomorrow and try to run the integration tests locally. Looking good so far 👍

src/server/pom.xml
adutra: Update: if upgrading to 2.35, it seems the following extra dependency is required:

```
        <dependency>
            <groupId>org.glassfish.jersey.core</groupId>
            <artifactId>jersey-common</artifactId>
            <version>${jersey.version}</version>
        </dependency>
```

Probably safer not to upgrade.

src/server/pom.xml
adutra: Unfortunately upgrading to 2.13.0 won't work out of the box. The following error pops up in acceptance tests:

```
 java.lang.RuntimeException: com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Joda date/time type `org.joda.time.DateTime` not supported by default: add Module "com.fasterxml.jackson.datatype:jackson-datatype-joda" to enable handling
```

I guess it's safer to stay on 2.10.5. Luckily enough, it seems there is currently no [CVE targeting that version](https://cve.mitre.org/cgi-bin/cvekey.cgi?keyword=jackson).

src/server/src/test/java/io/cassandrareaper/acceptance/ReaperTestJettyRunner.java
```diff
@@ -49,14 +47,22 @@
  */
 public final class ReaperTestJettyRunner {
 
+  static final Set<Integer> USED_PORTS = Sets.newConcurrentHashSet();
   static final String REAPER_APPLICATION_FQN = "io.cassandrareaper.ReaperApplication";
```
adutra: Unused field.

src/server/src/test/java/io/cassandrareaper/acceptance/ReaperTestJettyRunner.java
```diff
@@ -49,14 +47,22 @@
...
+      ConfigOverride.config("server.applicationConnectors[0].port", String.valueOf(getAnyAvailablePort())),
+      ConfigOverride.config("server.adminConnectors[0].port", String.valueOf(getAnyAvailablePort())));
+    try {
+      runnerInstance.before();
+    } catch (Exception e) { // CHECKSTYLE IGNORE THIS LINE
+      throw new RuntimeException(e);
+    }
 
     Runtime.getRuntime().addShutdownHook(new Thread() {
       @Override
```
adutra: Condition `runnerInstance != null` is always true on line 70. Same for assertion in line 78.

src/server/src/main/java/io/cassandrareaper/ReaperApplication.java
```diff
@@ -71,14 +71,14 @@
 import io.dropwizard.client.HttpClientBuilder;
```
adutra, line 71: The constructor `public ReaperApplication(AppContext context)` doesn't  seem used anymore, maybe it should be deleted?

src/server/src/test/java/io/cassandrareaper/acceptance/ReaperTestJettyRunner.java
adutra: The whole shutdown hook can be rewritten as:

```
    Runtime.getRuntime().addShutdownHook(new Thread(runnerInstance::after));
```

src/server/src/test/java/io/cassandrareaper/acceptance/ReaperTestJettyRunner.java
```diff
@@ -80,267 +86,32 @@ public SimpleReaperClient getClient() {
...
-      }
-    }
-
-    private static int getAnyAvailablePort() {
-      try (ServerSocket s = new ServerSocket(0)) {
-        return USED_PORTS.add(s.getLocalPort()) ? s.getLocalPort() : getAnyAvailablePort();
-      } catch (IOException ex) {
-        throw new IllegalStateException("no available ports", ex);
-      }
+  public Object getContext() {
```
adutra: I dont think you need reflection here. I suggest instead:

1. Add a getter to `ReaperApplication`:

```
  public AppContext getContext() {
    return context;
  }
```

2. Change this method and the following one to:

```
  public AppContext getContext() {
    ReaperApplication application = runnerInstance.getApplication();
    return application.getContext();
  }

  String getContextStorageClassname() {
    AppContext context = getContext();
    return context.storage.getClass().getName();
  }
```

adutra (changes requested):
Review completed, thanks @adejanovski for your help :-)

Please also address my previous comments that are not marked resolved.

src/server/src/main/java/io/cassandrareaper/ReaperApplication.java
adejanovski: Good catch! It was probably used in the integration tests through reflection. I'll remove it.

src/server/src/test/java/io/cassandrareaper/acceptance/ReaperCassandraSidecarIT.java
adejanovski: I'll clean up 👍

src/server/src/test/java/io/cassandrareaper/acceptance/ReaperCassandraSidecarIT.java
adejanovski: gotcha

src/server/src/test/java/io/cassandrareaper/acceptance/ReaperCassandraSidecarIT.java
adejanovski: indeed 👍

src/server/src/test/java/io/cassandrareaper/acceptance/ReaperTestJettyRunner.java
adejanovski: Sounds good 👍

src/server/src/test/java/io/cassandrareaper/acceptance/ReaperTestJettyRunner.java
```diff
@@ -80,267 +76,22 @@ public SimpleReaperClient getClient() {
...
-    }
-
-    private static int getAnyAvailablePort() {
-      try (ServerSocket s = new ServerSocket(0)) {
-        return USED_PORTS.add(s.getLocalPort()) ? s.getLocalPort() : getAnyAvailablePort();
-      } catch (IOException ex) {
-        throw new IllegalStateException("no available ports", ex);
-      }
-    }
+  public AppContext getContext() {
```
adutra, line 79: `getContext()` now returns `AppContext` directly instead of `Object`. This has a side-effect on all call sites: the cast to `AppContext` is now useless and should be removed.

src/server/src/test/java/io/cassandrareaper/acceptance/BasicSteps.java
```diff
@@ -271,7 +246,7 @@ public void reaper_has_no_cluster_in_storage() throws Throwable {
           String responseData = response.readEntity(String.class);
           Assertions.assertThat(responseData).isNotBlank();
           List<String> clusterNames = SimpleReaperClient.parseClusterNameListJSON(responseData);
-          if (!((AppContext) runner.runnerInstance.getContext()).config.isInSidecarMode()) {
+          if (!((AppContext) runner.getContext()).config.isInSidecarMode()) {
```
adutra: This is the kind of places where the cast to `AppContext` becomes obsolete.

adutra (changes requested):
Almost ready to go 👍

src/server/pom.xml
adejanovski: right 👍  removed

src/server/pom.xml
adejanovski: We don't indeed. Removed ✅

src/server/src/test/java/io/cassandrareaper/acceptance/ReaperTestJettyRunner.java
adejanovski: I removed all casts 👍

adutra (dismissed):

PR #1159, opened 2022-01-10
fix: :sparkles: ready for Openshift

sentoz:
This fix fixes the issue with the permissions to modify /etc/cassandra-reaper.yaml by the configure-persistence.sh and configure-webui-authentication.sh scripts, when the container starts in OpenShift.

Note that the user is added to the 0 group. This group does not carry additional security privileges, but allows containers to be correctly deployed in the OpenShift cluster. The OpenShift cluster, unlike Kubernetes, automatically replaces the user in the container with a randomly generated one. And the only sign of this user is belonging to the 0 group.


adejanovski:
@burmanm, what's your opinion on this change?

PR #1169, opened 2022-02-17
Fixing broken Dropwizard Link

Rooks103:
Looks like Reaper moved to Dropwizard 2.0 as part of #1148 so fixing this link to point to the newer Dropwizard.io page.


adejanovski (approved):
Thans @Rooks103 !

PR #1171, opened 2022-03-11
Recompute segment replicas

### LINKED ISSUE #1105: cassandra cluster on K8s has a pod restarted and reaper is unable to get the new pod IP instead keep trying the old ip of the pod.
Cassandra cluster on K8s has a pod restarted and reaper is unable to get the new pod IP instead keep trying the old ip of the pod. This is memory storage. 

INFO   [2021-07-19 18:17:11,326] [pool-12-thread-5] i.c.j.JmxConnectionFactory - Adding new JMX Proxy for host 10.1.11:7199
ERROR  [2021-07-19 18:17:31,326] [pool-12-thread-5] i.c.j.JmxConnectionFactory - Failed creating a new JMX connection to 10.1.102.11:7199
java.lang.RuntimeException: io.cassandrareaper.ReaperException: Failure when establishing JMX connection to 10.1.102.11:7199

adejanovski:
Fixes the K8ssandra issue reported [here](https://github.com/k8ssandra/k8ssandra/issues/1314).
Fixes #1105

When nodes get replaced or if pods restart in Cassandra, they may get a new IP which is different from the one recorded at the segment creation time.
While capturing the list of IPs on the segments is important for efficient scheduling of segments, we need to recompute the actual list when segments are scheduled so that we can handle IP changes.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
+++ b/src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
@@ -36,6 +36,7 @@ import io.cassandrareaper.storage.IDistributedStorage;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.NoSuchElementException;
@@ -81,6 +82,7 @@ final class RepairRunner implements Runnable {
 
   private final AppContext context;
   private final ClusterFacade clusterFacade;
+  private final RepairRunService repairRunService;
   private final UUID repairRunId;
   private final String clusterName;
   private final String metricNameForMillisSinceLastRepairPerKeyspace;
@@ -100,6 +102,7 @@ final class RepairRunner implements Runnable {
     LOG.debug("Creating RepairRunner for run with ID {}", repairRunId);
     this.context = context;
     this.clusterFacade = clusterFacade;
+    this.repairRunService = RepairRunService.create(context);
     this.repairRunId = repairRunId;
     Optional<RepairRun> repairRun = context.storage.getRepairRun(repairRunId);
     assert repairRun.isPresent() : "No RepairRun with ID " + repairRunId + " found from storage";
@@ -407,12 +410,15 @@ final class RepairRunner implements Runnable {
                 repairRunId);
 
     Optional<RepairSegment> nextRepairSegment = Optional.empty();
+    Collection<String> potentialReplicas = new HashSet<>();
     for (RepairSegment segment : nextRepairSegments) {
-      Collection<String> potentialReplicas = repairUnit.getIncrementalRepair()
+      Map<String, String> potentialReplicaMap = this.repairRunService.getDCsByNodeForRepairSegment(
+          cluster, segment.getTokenRange(), repairUnit.getKeyspaceName(), repairUnit);
+      potentialReplicas = repairUnit.getIncrementalRepair()
           ? Collections.singletonList(segment.getCoordinatorHost())
-          : segment.getReplicas().keySet();
+          : potentialReplicaMap.keySet();
       JmxProxy coordinator = clusterFacade.connect(cluster, potentialReplicas);
-      if (nodesReadyForNewRepair(coordinator, segment, segment.getReplicas(), repairRunId)) {
+      if (nodesReadyForNewRepair(coordinator, segment, potentialReplicaMap, repairRunId)) {
         nextRepairSegment = Optional.of(segment);
         break;
       }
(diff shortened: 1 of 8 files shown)


src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
```diff
@@ -407,12 +410,15 @@ private void startNextSegment() throws ReaperException, InterruptedException {
                 repairRunId);
 
     Optional<RepairSegment> nextRepairSegment = Optional.empty();
+    Collection<String> potentialReplicas = new HashSet<>();
     for (RepairSegment segment : nextRepairSegments) {
-      Collection<String> potentialReplicas = repairUnit.getIncrementalRepair()
+      Map<String, String> potentialReplicaMap = this.repairRunService.getDCsByNodeForRepairSegment(
+          cluster, segment.getTokenRange(), repairUnit.getKeyspaceName(), repairUnit);
+      potentialReplicas = repairUnit.getIncrementalRepair()
           ? Collections.singletonList(segment.getCoordinatorHost())
```
adutra, line 418: Sorry if the question is stupid, but I'm wondering what happens when it's the segment's coordinator host that had its IP replaced? Would `segment.getCoordinatorHost()` return the new IP?

src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
```diff
@@ -610,33 +616,6 @@ private boolean repairSegment(final UUID segmentId, Segment segment)
           killAndCleanupRunner();
         }
 
-        return false;
-      } else if (
-          !Sets.difference(
```
adutra, line 615: Why don't we need this brach anymore?

src/server/src/test/java/io/cassandrareaper/service/RepairRunnerTest.java
```diff
@@ -958,16 +935,51 @@ protected JmxProxy connectImpl(Node host) throws ReaperException {
...
+    when(clusterFacade.getRangeToEndpointMap(any(), anyString()))
+        .thenReturn((Map)ImmutableMap.of(
+            Lists.newArrayList("0", "100"), Lists.newArrayList(nodeSetAfterTopologyChange),
+            Lists.newArrayList("100", "200"), Lists.newArrayList(nodeSetAfterTopologyChange)));
+    when(clusterFacade.getEndpointToHostId(any())).thenReturn(nodeMapAfterTopologyChange);
     when(clusterFacade.tokenRangeToEndpoint(any(), anyString(), any()))
       .thenReturn(Lists.newArrayList(nodeSetAfterTopologyChange));
     context.repairManager.resumeRunningRepairRuns();
     // The repair should now fail as the list of replicas for the segments are different from storage
     await().with().atMost(20, TimeUnit.SECONDS).until(() -> {
```
adutra, line 948: I'm having a hard time trying to understand the intent here. This Awaitility call always fails, throws a `ConditionTimeoutException`, which the test expects to receive, and the rest of the test is basically dead code. Could you refactor this to better express the intent? Afaict the intent seems to be: "wait 20 seconds and then assert that the run state is not ERROR". I wonder if we couldn't simplify this to "await at most 20 seconds until the run state is DONE", wdyt?

Also, is this comment still relevant?

> The repair should now fail as the list of replicas for the segments are different from storage

Afaict the repair is not failing.

adutra (changes requested):

src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
adejanovski: That's actually a very good question :)
To give a little bit of context, incremental repairs create "fake" segments in Reaper. One per host, because it has to run on all ranges of each node, not on subranges.
Hence, when the repair run is created, the coordinator host for each segment is set and it cannot be recomputed afterwards like we do for token ranges.
We yet have to get issues raised on that particular issue, but I'd consider it outside of the scope for this PR, since the implementation of incremental repair is fairly different.
I guess a workaround could be to use host ids instead of ip addresses for the coordinator hosts and resolve the IP address right before triggering the segments.

src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
adejanovski: This is precisely what this PR fixed: if the set of replicas changed for a segment, it meant the IP addresses we pre-recorded aren't valid anymore and we need to stop the repair.
We can't have that behavior anymore since we're recomputing the list of live replicas before starting the segment.
The only thing that can still happen is a cluster expansion/shrink which invalidates our token ranges, and that's still covered with a failing repair run (Reaper will fail to get a list of replicas for the token range as it now spans over two or more separate token ranges).

src/server/src/test/java/io/cassandrareaper/service/RepairRunnerTest.java
adejanovski: I agree that it's hard to understand. I'll try your way 👍

adutra (approved):

PR #1173, opened 2022-03-18
Make heap size configurable through env variable

adejanovski:
This PR makes it possible to set the heap size for Reaper through an environment variable.
Previously, we were using a hardcoded 2GB heap for non containerized deployments and no heap setting for containers (which defaults to 1G heaps).
Defaults were preserved for both types of deployments.


emerkle826 (approved):
:+1:

PR #1174, opened 2022-03-22
UI tooling update

jdonenine:
After forgetting on two different machines I was trying out to install bower...

* Remove the manually installed dependency on bower by moving it within the npm dependencies and scripts
* Consolidates the UI dev/build process to be exclusively controlled through npm
* The `bower install` will happen automatically following an `npm install`

Also updated the README to note that it requires Node <= 14 (had all sorts of strange problem on 16 at first)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -54,9 +54,6 @@ jobs:
         with:
           node-version: '10.x'
 
-      - name: Setup bower
-        run: npm install -g bower
-
       - name: Setup Python
         uses: actions/setup-python@v1
         with:
@@ -102,9 +99,6 @@ jobs:
         uses: actions/setup-node@v1
         with:
           node-version: '10.x'
-
-      - name: Setup bower
-        run: npm install -g bower
 
       - name: Setup Python
         uses: actions/setup-python@v1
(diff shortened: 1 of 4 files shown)


jdonenine:
@adejanovski what's the easiest way to build a container image that I could test this with?  I'm able to test locally, but I'd like to test the whole build workflow if possible since I'm touching the maven setup here.

adejanovski:
> @adejanovski what's the easiest way to build a container image that I could test this with? I'm able to test locally, but I'd like to test the whole build workflow if possible since I'm touching the maven setup here.

You can build the Docker image with: `mvn -B -pl src/server/ docker:build -Ddocker.directory=src/server/src/main/docker` 

It'll create the image as `cassandra-reaper:latest` locally.
That part is tested in CI as `Docker Image Tests` though, but I reckon we don't test the UI in CI.

jdonenine:
> You can build the Docker image with: `mvn -B -pl src/server/ docker:build -Ddocker.directory=src/server/src/main/docker`

Thanks, that worked (once I moved to something other than my m1 Mac anyhow), I'll test it out tonight a bit more.

jdonenine:
I was able to get the image built and tested in a cluster with this image `jeffdinoto/cassandra-reaper:9ea7d6847eca657b156811171baaba3d0b527646` all was well, so I think if we find the change useful for the developer workflow it should be good to go forward.

adejanovski (approved):
very nice 👍 
I'll merge the PR. Thanks @jdonenine

Issue #1176, opened 2022-03-22
Add JDK 11 to build container


ossarga:
The build container only installs JDK 8 and is unable to build versions 3.1 and above.

ossarga:
Opened pull request https://github.com/thelastpickle/cassandra-reaper/pull/1177 to fix this issue.

ossarga:
Changes merged. Closing issue.

PR #1177, opened 2022-03-23
ISSUE #1176: Add JDK 11 to build container

ossarga:
* Added logic to switch to JDK 8 when building versions 3.0 and below
* Added logic to allow commands to be passed when running docker compose to force build the JAR


ossarga:
Fixes issue #1177

adejanovski (approved):
Thanks @ossarga !
Totally forgot to update this when we shipped 3.1.0 😅

ossarga:
Thanks for the quick review @adejanovski !

PR #1190, opened 2022-04-20
Don't exit reaper if endpointToRange is empty and segment count 0

xgerman:
We have seen reaper exiting dueing this condition and this turns
it into a reaper exception instead of hitting the assertion in
computeGlobalSegmentCount.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/service/RepairRunService.java
+++ b/src/server/src/main/java/io/cassandrareaper/service/RepairRunService.java
@@ -182,6 +182,9 @@ public final class RepairRunService {
 
     } catch (ReaperException e) {
       LOG.warn("couldn't connect to any host: {}, life sucks...", targetCluster.getSeedHosts(), e);
+    } catch (IllegalArgumentException e) {
+      LOG.error("Couldn't get endpoints for tokens");
+      throw new ReaperException("Couldn't get endpoints for tokens", e);
     }
 
     if (segments.isEmpty() && !repairUnit.getIncrementalRepair()) {
--- a/src/server/src/test/java/io/cassandrareaper/service/RepairRunServiceTest.java
+++ b/src/server/src/test/java/io/cassandrareaper/service/RepairRunServiceTest.java
@@ -779,4 +779,90 @@ public final class RepairRunServiceTest {
     Set<String> datacenters = RepairRunService.getDatacentersToRepairBasedOnParam(Optional.of("dc1,dc2"));
     assertEquals("Datacenters were not parsed correctly", 2, datacenters.size());
   }
+
+
+  @Test(expected = ReaperException.class)
+  public void generateSegmentsTestEmpty() throws ReaperException, UnknownHostException {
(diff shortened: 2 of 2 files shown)


src/server/src/main/java/io/cassandrareaper/service/RepairRunService.java
```diff
@@ -168,6 +168,10 @@ List<Segment> generateSegments(
       Map<List<String>, List<RingRange>> replicasToRange = buildReplicasToRangeMap(rangeToEndpoint);
       String cassandraVersion = clusterFacade.getCassandraVersion(targetCluster);
 
+      if (segmentCountPerNode == 0 && endpointToRange.isEmpty()) {
```
adejanovski: Is there a relationship between the fact that `segmentCountPerNode == 0` and the fact that the endpoint to range mapping is empty?
My assumption would be that the segment count value should be unrelated to the failure.

Instead of adding a conditional, would it be possible to catch the `IllegalArgumentException` in the surrounding try, and rethrow it as a `ReaperException`? This would allow to leverage the existing Precondition check from `computeGlobalSegmentCount`.

adejanovski (changes requested):
Thanks for creating this PR @xgerman 🙏 
I've suggested a small change before we can move forward and merge it.

src/server/src/main/java/io/cassandrareaper/service/RepairRunService.java
xgerman: Thanks for reviewing - I made the changes.

adejanovski:
Hi @xgerman,

[looks like](https://github.com/thelastpickle/cassandra-reaper/runs/6284275270?check_suite_focus=true#step:7:91) you have a trailing whitespace on line 170 of `RepairRunService.java`, which is failing the CI build.
Would it be possible to move your new catch expression to the existing surrounding `try/catch` block?

```
      int globalSegmentCount = 0;
      globalSegmentCount = computeGlobalSegmentCount(segmentCountPerNode, endpointToRange);

      segments = filterSegmentsByNodes(
              sg.generateSegments(
                  globalSegmentCount,
                  tokens,
                  repairUnit.getIncrementalRepair(),
                  replicasToRange,
                  cassandraVersion),
              repairUnit,
              endpointToRange);
    } catch (ReaperException e) {
      LOG.warn("couldn't connect to any host: {}, life sucks...", targetCluster.getSeedHosts(), e);
    } catch (IllegalArgumentException e) {
      LOG.error("Couldn't get endpoints for tokens");
      throw new ReaperException("Couldn't get endpoints for tokens", e);
    }

```

Otherwise it'll get caught by the existing catch and won't get properly re-thrown.
Moving the logging to error level seems appropriate as well here since we're failing the repair.

Thanks!

xgerman:
Hi @adejanovski.

Thank you for the review. I implemented the changes.

adejanovski (approved):
LGTM, thanks @xgerman and sorry for the time it took to get this merged.

PR #1197, opened 2022-05-31
Better handle runner concurrency

adejanovski:
The current way of limiting the number of concurrent runners is done by checking the number of runners in the local Reaper instance. Paused runs still have a runner thread, which gets counted as well, resulting in paused repairs potentially blocking other repairs.
Another problem that can arise is that different reaper instance could start runs in a different order, resulting in different runs being authorized depending on the instance.
The current set of changes orders the runners by repair run creation date (the run id is a timeuuid), to prioritize older runs, and also takes into account the state of the run, so that paused runs let running one be processed.


adutra (approved):

max-melentyev:
Is it possible that this prevents reaper from running parallel repairs on different clusters? It looks like old code checked existing runs filtering them by cluster name but now there is no filtering.

adejanovski:
> Is it possible that this prevents reaper from running parallel repairs on different clusters? It looks like old code checked existing runs filtering them by cluster name but now there is no filtering.

Yes, it's the case. And we need to address this so that the limit applies per cluster.

adejanovski:
I created [an issue](https://github.com/thelastpickle/cassandra-reaper/issues/1269) to track this.
@max-melentyev, we happily accept PRs if you're willing to contribute the fix ;)

max-melentyev:
I'll take a look if I can fix it.
Meanwhile, could you please review #1255 and #1267 ?

PR #1209, opened 2022-07-01
Avoid exiting the JVM when no scheduling leader is present

### LINKED ISSUE #1206: All repair schedules turned off, but seeing an id: null error
Customer has disabled all repairs based on an issue they're seeing with snapshots. In the meantime, while all repairs are disabled, they're getting the following error:

```
INFO   [2022-06-16 00:09:45,093] [SchedulingManagerTimer] i.c.s.SchedulingManager - Repair schedule '24aa0620-8826-11ec-ba6f-0338dd77ee5b' is paused 
ERROR  [2022-06-16 00:10:45,091] [SchedulingManagerTimer] i.c.s.SchedulingManager - failed managing schedule for run with id: null 
ERROR  [2022-06-16 00:10:45,091] [SchedulingManagerTimer] i.c.s.SchedulingManager - catch exception 
java.lang.IndexOutOfBoundsException: Index: 0, Size: 0
	at java.util.ArrayList.rangeCheck(ArrayList.java:659)
	at java.util.ArrayList.get(ArrayList.java:435)
	at io.cassandrareaper.service.SchedulingManager.currentReaperIsSchedulingLeader(SchedulingManager.java:263)
	at io.cassandrareaper.service.SchedulingManager.run(SchedulingManager.java:108)
	at java.util.TimerThread.mainLoop(Timer.java:555)
	at java.util.TimerThread.run(Timer.java:505)
ERROR  [2022-06-16 00:10:45,091] [SchedulingManagerTimer] i.c.s.SchedulingManager - SchedulingManager failed. Exiting JVM. 
INFO   [2022-06-16 00:10:50,715] [JettyShutdownThread] i.c.s.CassandraStorage - Reaper is stopping, removing this instance from running reapers... 
```

I don't see anything similar in the reported issues, any ideas on what could be causing this? As you can see, the JVM is shutting down due to the error.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1590) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1590
┆priority: Medium

adejanovski:
Fixes #1206

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/service/RepairScheduleService.java
+++ b/src/server/src/main/java/io/cassandrareaper/service/RepairScheduleService.java
@@ -154,15 +154,17 @@ public final class RepairScheduleService {
   }
 
   private void unregisterScheduleMetrics(UUID repairScheduleId) {
-    RepairSchedule schedule = context.storage.getRepairSchedule(repairScheduleId).get();
-    RepairUnit repairUnit = context.storage.getRepairUnit(schedule.getRepairUnitId());
-    String metricName = metricName("millisSinceLastRepairForSchedule",
-        repairUnit.getClusterName(),
-        repairUnit.getKeyspaceName(),
-        schedule.getId());
+    Optional<RepairSchedule> schedule = context.storage.getRepairSchedule(repairScheduleId);
+    if (schedule.isPresent()) {
+      RepairUnit repairUnit = context.storage.getRepairUnit(schedule.get().getRepairUnitId());
+      String metricName = metricName("millisSinceLastRepairForSchedule",
+          repairUnit.getClusterName(),
+          repairUnit.getKeyspaceName(),
+          schedule.get().getId());
 
-    if (context.metricRegistry.getMetrics().containsKey(metricName)) {
-      context.metricRegistry.remove(metricName);
+      if (context.metricRegistry.getMetrics().containsKey(metricName)) {
+        context.metricRegistry.remove(metricName);
+      }
     }
   }
 
--- a/src/server/src/main/java/io/cassandrareaper/service/RepairScheduleService.java
+++ b/src/server/src/main/java/io/cassandrareaper/service/RepairScheduleService.java
@@ -155,17 +155,17 @@ public final class RepairScheduleService {
 
   private void unregisterScheduleMetrics(UUID repairScheduleId) {
     Optional<RepairSchedule> schedule = context.storage.getRepairSchedule(repairScheduleId);
-    if (schedule.isPresent()) {
-      RepairUnit repairUnit = context.storage.getRepairUnit(schedule.get().getRepairUnitId());
+    schedule.ifPresent(sched -> {
+      RepairUnit repairUnit = context.storage.getRepairUnit(sched.getRepairUnitId());
       String metricName = metricName("millisSinceLastRepairForSchedule",
           repairUnit.getClusterName(),
           repairUnit.getKeyspaceName(),
-          schedule.get().getId());
+          sched.getId());
 
       if (context.metricRegistry.getMetrics().containsKey(metricName)) {
         context.metricRegistry.remove(metricName);
       }
-    }
+    });
   }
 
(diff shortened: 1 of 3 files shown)


src/server/src/main/java/io/cassandrareaper/service/RepairScheduleService.java
```diff
@@ -154,15 +154,17 @@ private void registerScheduleMetrics(UUID repairScheduleId) {
...
+    if (schedule.isPresent()) {
+      RepairUnit repairUnit = context.storage.getRepairUnit(schedule.get().getRepairUnitId());
+      String metricName = metricName("millisSinceLastRepairForSchedule",
+          repairUnit.getClusterName(),
+          repairUnit.getKeyspaceName(),
+          schedule.get().getId());
+
+      if (context.metricRegistry.getMetrics().containsKey(metricName)) {
+        context.metricRegistry.remove(metricName);
+      }
```
adejanovski: In the integration tests, we have several reaper that compete to delete the schedule, which made it so that on reaper could try to remove it after another one removed it. Hence, the optional would be empty and the test would fail.
This change allows avoiding such failures that create flakiness in the tests.

src/server/src/main/java/io/cassandrareaper/service/RepairScheduleService.java
```diff
@@ -154,15 +154,17 @@ private void registerScheduleMetrics(UUID repairScheduleId) {
   }
 
   private void unregisterScheduleMetrics(UUID repairScheduleId) {
-    RepairSchedule schedule = context.storage.getRepairSchedule(repairScheduleId).get();
-    RepairUnit repairUnit = context.storage.getRepairUnit(schedule.getRepairUnitId());
```
burmanm, line 158: Wouldn't simple ``if(schedule != null)`` have sufficed? The optional's use for "isPresent + get" is considered a bad pattern. Not to mention there's multiple ``.get()`` calls. Optional works better (and more pretty) if used with a lambda:

```java
Optional<RepairSchedule> schedule = context.storage.getRepairSchedule(repairScheduleId);
schedule.ifPresent(s -> {
  RepairUnit repairUnit = context.storage.getRepairUnit(s).getRepairUnitId();
  // etc
});
```

src/server/src/main/java/io/cassandrareaper/service/RepairScheduleService.java
burmanm: This comment brings me to another question, is that race condition really intended? If so, this isn't really safe alternative either? If we get the repairUnit and then another thread removes it, we're now re-removing data from metricRegistry since these two threads are racing to delete the metric.

adejanovski: It's not several threads in the same Reaper instance, but several Reaper instances running in different JVMs. The problem here is that the metric won't get cleaned up on other Reaper instances, making that cleanup mostly inefficient in distributed mode 😕 
I think we could improve this by listing the schedules and checking if there are corresponding schedules for the repair unit. This way we don't need to have the schedule in storage for cleanup to happen.
Let me try that.

adejanovski: Now that I think of it, unlike during tests a schedule deletion will only be sent to a single instance. This means metrics never get fully cleaned up in a distributed setup.
We'd need to regularly clean them up. I'll open another ticket to handle this.

src/server/src/main/java/io/cassandrareaper/service/RepairScheduleService.java
adejanovski: Sounds good, I'll use the lambda form 👍

src/server/src/main/java/io/cassandrareaper/service/RepairScheduleService.java
adejanovski: I've created the follow up issue: #1211

burmanm (approved):

PR #1224, opened 2022-09-07
Updated the registry clusters entrypoint to support the use of custom username and password combinations

paulslaven:
The current implementation of the register clusters entrypoint sets the password to "admin" in the credentials file and fails on authentication if any other username/password combination is used. This value is picked up by the get_password method in the spreaper, as this finds the value in the file, and thus attempts authentication with this hard coded value.

This PR maintains the default support for the default username and password while allowing those who change the username/password for the Cassandra Reaper to also use these via environment variable (REAPER_AUTH_USER and REAPER_AUTH_PASSWORD).

This would enable more flexible use of this register-cluster method while allowing better security though stronger credentials in its usage.

### DIFF
--- a/src/server/src/main/docker/entrypoint.sh
+++ b/src/server/src/main/docker/entrypoint.sh
@@ -87,8 +87,22 @@ if [ "$1" = 'register-clusters' ]; then
     REAPER_PORT=$4
   fi
 
+  if [ -z "$REAPER_AUTH_USER" ]; then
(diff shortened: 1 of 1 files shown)


adejanovski (approved):
Thanks @paulslaven, LGTM 👍

paulslaven:
Great project @adejanovski, thanks for the review 👍

PR #1244, opened 2022-11-18
Allow for OPTIONS requests to be passed through auth filters

### LINKED ISSUE #1245: Allow for OPTIONS requests to be passed through auth filters
Browsers do not pass authorization content in OPTIONS requests.
This causes all preflight checks in the browser to fail when making a request to an authenticated endpoint when doing so with CORS enabled.

jdonenine:
Fixes #1245 

* Browsers do not pass authorization content in OPTIONS requests
* This causes all preflight checks in the browser to fail when making a request to an authenticated endpoint when doing so with CORS enabled
* The change makes it possible to utilize a JWT from within the browser via fetch with CORS -- allowing for the separation of the frontend from the backend
* This also modifies the Shiro configuration such that it properly secures the cookies that it generates for CORS usage.

I think this behavior should probably be considered a bug, so I tied the behavior to the already in-use environment variable `REAPER_ENABLE_CROSS_ORIGIN` which when set to `true` will cause all OPTIONS requests to be allowed without an auth check.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/resources/RequestUtils.java
+++ b/src/server/src/main/java/io/cassandrareaper/resources/RequestUtils.java
@@ -24,10 +24,18 @@ import javax.ws.rs.HttpMethod;
 import com.google.common.annotations.VisibleForTesting;
 
 public final class RequestUtils {
-  public static final String ENABLE_CORS_ENV_VAR_NAME = "REAPER_ENABLE_CROSS_ORIGIN";
+  private static boolean isCorsEnabled = false;
 
   private RequestUtils() {}
 
+  public static void setCorsEnabled(boolean enabled) {
+    isCorsEnabled = enabled;
+  }
+  
+  public static boolean isCorsEnabled() {
+    return isCorsEnabled;
+  }
+  
   public static boolean isOptionsRequest(ServletRequest request) {
     if (request != null && request instanceof HttpServletRequest) {
       if (((HttpServletRequest) request).getMethod().equalsIgnoreCase(HttpMethod.OPTIONS)) {
@@ -35,18 +43,5 @@ public final class RequestUtils {
       }
     }
     return false;
-  }
-
-  @VisibleForTesting
-  static boolean isCorsEnabled(String corsEnabledEnvVarValue) {
-    if (corsEnabledEnvVarValue != null) {
-      return Boolean.parseBoolean(corsEnabledEnvVarValue.trim().toLowerCase());
-    }
-    return false;
-  }
-
-  public static boolean isCorsEnabled() {
-    String corsEnabledEnvVarValue = System.getenv(ENABLE_CORS_ENV_VAR_NAME);
-    return isCorsEnabled(corsEnabledEnvVarValue);
   }
 }
--- a/src/server/src/main/java/io/cassandrareaper/resources/RequestUtils.java
+++ b/src/server/src/main/java/io/cassandrareaper/resources/RequestUtils.java
@@ -29,11 +29,11 @@ public final class RequestUtils {
   public static void setCorsEnabled(boolean enabled) {
     isCorsEnabled = enabled;
   }
-  
+
   public static boolean isCorsEnabled() {
     return isCorsEnabled;
   }
-  
+
   public static boolean isOptionsRequest(ServletRequest request) {
     if (request != null && request instanceof HttpServletRequest) {
       if (((HttpServletRequest) request).getMethod().equalsIgnoreCase(HttpMethod.OPTIONS)) {
--- a/src/server/src/main/java/io/cassandrareaper/resources/RequestUtils.java
+++ b/src/server/src/main/java/io/cassandrareaper/resources/RequestUtils.java
@@ -21,8 +21,6 @@ import javax.servlet.ServletRequest;
 import javax.servlet.http.HttpServletRequest;
 import javax.ws.rs.HttpMethod;
 
-import com.google.common.annotations.VisibleForTesting;
-
 public final class RequestUtils {
   private static boolean isCorsEnabled = false;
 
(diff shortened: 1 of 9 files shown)


src/server/src/main/java/io/cassandrareaper/resources/RequestUtils.java
```diff
@@ -0,0 +1,52 @@
...
+package io.cassandrareaper.resources;
+
+import javax.servlet.ServletRequest;
+import javax.servlet.http.HttpServletRequest;
+import javax.ws.rs.HttpMethod;
+
+import com.google.common.annotations.VisibleForTesting;
+
+public final class RequestUtils {
+  public static final String ALLOW_ALL_OPTIONS_REQUESTS_ENV_VAR_NAME = "ALLOW_ALL_OPTIONS_REQUESTS";
```
jdonenine: The more I think about this, the more I think this should just be switched on with the already used `REAPER_ENABLE_CROSS_ORIGIN` environment variable instead of having the additional flag.  Simplifies things as I think this is really needed to really use CORS with auth. 🤷

jdonenine: I did indeed go ahead and make the switch here, figured that's what I'd prefer, so I'd just put it forth here to figure out.

jdonenine:
So, it turns out that this still isn't sufficient to get through the CSRF protections in browsers because of the way the cookies that Shiro creates are created.  I'm still working on that aspect of things, I think it's just a shiro configuration to be made, but it's elusive to say the least.  Going to make this a draft for the moment while I sort that out.

src/server/src/main/java/io/cassandrareaper/resources/RequestUtils.java
```diff
@@ -0,0 +1,52 @@
...
+      return Boolean.parseBoolean(corsEnabledEnvVarValue.trim().toLowerCase());
+    }
+    return false;
+  }
+
+  public static boolean isCorsEnabled() {
+    String corsEnabledEnvVarValue = System.getenv(ENABLE_CORS_ENV_VAR_NAME);
+    return isCorsEnabled(corsEnabledEnvVarValue);
+  }
+}
```
adejanovski: After reviewing this and thinking about another requirement we have (which is to shut down the sessions despite being active after the configured timeout), I'm of the opinion that we shouldn't rely on env variable.
We can statically set a `isCorsEnabled` variable in RequestUtils which will be initialized in the ReaperApplication.java initialization phase, and feed on the corresponding config setting: `config.isEnableCrossOrigin()`

```suggestion
public final class RequestUtils {
  private static boolean isCorsEnabled = false;

  private RequestUtils() {}

  public static void setCorsEnabled(boolean enabled) {
    isCorsEnabled = enabled;
  }
  
  public static boolean isCorsEnabled() {
    return isCorsEnabled;
  }
  
  public static boolean isOptionsRequest(ServletRequest request) {
    if (request != null && request instanceof HttpServletRequest) {
      if (((HttpServletRequest) request).getMethod().equalsIgnoreCase(HttpMethod.OPTIONS)) {
        return true;
      }
    }
    return false;
  }
}
```

Initialization can be done as follows in ReaperApplication.java: 

```
 if (config.isAccessControlEnabled()) {
      RequestUtils.setCorsEnabled(config.isEnableCrossOrigin());
      SessionHandler sessionHandler = new SessionHandler();
      sessionHandler.setMaxInactiveInterval((int) config.getAccessControl().getSessionTimeout().getSeconds());
      environment.getApplicationContext().setSessionHandler(sessionHandler);
      environment.servlets().setSessionHandler(sessionHandler);
      environment.jersey().register(new ShiroExceptionMapper());
      environment.jersey().register(new LoginResource());
      environment.jersey().register(new ShiroJwtProvider(context));
    }
```

wdyt?

src/server/src/main/java/io/cassandrareaper/resources/auth/ShiroJwtVerifyingFilter.java
```diff
@@ -17,32 +17,51 @@
...
 
 public final class ShiroJwtVerifyingFilter extends AccessControlFilter {
 
   private static final Logger LOG = LoggerFactory.getLogger(ShiroJwtVerifyingFilter.class);
 
-  public ShiroJwtVerifyingFilter() {}
+  private final boolean isCorsEnabled;
+
+  public ShiroJwtVerifyingFilter() {
+    isCorsEnabled = RequestUtils.isCorsEnabled();
```
adejanovski: Why do we need to duplicate this variable in multiple classes instead of just using `RequestUtils.isCorsEnabled()` when we need it?

adejanovski (changes requested):

src/server/src/main/java/io/cassandrareaper/resources/auth/ShiroJwtVerifyingFilter.java
jdonenine: this was just so that it was only initialized once on startup and not called through each time.

adejanovski: With the change I'm proposing, we should be able to get rid of the duplicates, right?

jdonenine: Yup, wouldn't matter anymore.  I'm going to try to take a look at it tonight.

src/server/src/main/java/io/cassandrareaper/resources/RequestUtils.java
jdonenine: OK, so the latest commit switches to initializing with config property instead of directly using the env variable... duh, I should have seen that to start 😔 so thanks for that suggestion!

But, I wasn't sure why you suggested tangling up the enabling of CORS within the isAccessControlEnabled() block as you mention here?

I initialized things in place with the other CORS related configuration that already happened within ReaperApplication.java's startup code.

adejanovski: > But, I wasn't sure why you suggested tangling up the enabling of CORS within the isAccessControlEnabled() block as you mention here?

I was under the impression that we needed it to allow OPTIONS calls when auth is enabled. Isn't it what you were trying to do here?
Either way, I'm fine with initializing it outside of that block 👍

jdonenine: Nah, the two can be unrelated.

adejanovski (approved):
Awesome sauce DiNoto 🥳 
Thanks!

PR #1254, opened 2022-12-20
Properly expire shiro sessions after timeout

### LINKED ISSUE #1225: accessControl sessionTimeout cassandra-reaper
Hi,

OS :
CentOS Linux release 7.9.2009 (Core)

I certify that the config sessiontimout do not work in file cassandra-reaper.yaml

"accessControl:
  sessionTimeout: PT10M"

The session stay open after 10 minutes. I testing change to PT60S with the same results.

Otherwise all work correctly but no the sessiontimout.

Somebody do have a idea please ? A bug may be ?

Best regards.

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1790) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1790
┆priority: Medium

adejanovski:
Fixes #1225

Logs out after the configured session timeout, despite the session refresh that the UI does in the background.
I've also added an expiry date to the JWTs which uses the session timeout. So far, the JWTs we were generating never expired.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/resources/auth/RestPermissionsFilter.java
+++ b/src/server/src/main/java/io/cassandrareaper/resources/auth/RestPermissionsFilter.java
@@ -20,6 +20,7 @@ package io.cassandrareaper.resources.auth;
 import io.cassandrareaper.resources.RequestUtils;
 
 import java.io.IOException;
+import java.util.Date;
 
 import javax.servlet.ServletRequest;
 import javax.servlet.ServletResponse;
@@ -42,6 +43,17 @@ public final class RestPermissionsFilter extends HttpMethodPermissionFilter {
       throws IOException {
     if (isCorsEnabled() && RequestUtils.isOptionsRequest(request)) {
       return true;
+    }
+
+    Subject subject = getSubject(request, response);
+
+    if (!subject.getPrincipals().getRealmNames().contains("jwtRealm")
+        && !RequestUtils.getSessionTimeout().isNegative()
+        && subject.getSession().getStartTimestamp().before(
+        new Date(System.currentTimeMillis() - RequestUtils.getSessionTimeout().toMillis()))) {
+      // Session has lived longer than its timeout already. Force logout.
+      subject.logout();
+      return false;
     }
     return super.isAccessAllowed(request, response, mappedValue);
   }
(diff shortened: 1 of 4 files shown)


src/server/src/main/java/io/cassandrareaper/resources/auth/RestPermissionsFilter.java
```diff
@@ -43,6 +44,17 @@ public boolean isAccessAllowed(ServletRequest request, ServletResponse response,
     if (isCorsEnabled() && RequestUtils.isOptionsRequest(request)) {
       return true;
     }
+
+    Subject subject = getSubject(request, response);
+
+    if (!subject.getPrincipals().getRealmNames().contains("jwtRealm")
+        && !RequestUtils.getSessionTimeout().isNegative()
+        && subject.getSession().getStartTimestamp().before(
```
jdonenine, line 52: This is going to "timeout" now, even if there were legitimate requests that would otherwise keep things alive, is that what you're going for?

adejanovski: Yeah, this was raised as a security concern. If the UI is running in a browser, the session will never timeout due to the background calls.
Not sure what's the usual practice, but requiring regular connections seems legit.
If the default timeout is too low, it can be raised to a higher value in config.
Do you have concerns with that approach?

jdonenine: Nope, no particular concern, just making sure I understood.  The session time is now a hard set number, no amount of application activity would keep it open and at some point the user would get logged out unexpectedly.  Not the experience you'd want, but I don't think that's a big deal for the way Reaper's UI is used though.

adejanovski: Maybe we can make this an opt-in feature, so that the default behavior remains the same but users with stronger requirements can enable it?

jdonenine: My guess is that changing the default to be more secure is fine.  The trade here in my mind is a slight usability downgrade (which I think is actually unlikely to really impact folks) for a security requirement that would serve to keep running the tool viable in more strict environments.

jdonenine (approved):

PR #1255, opened 2022-12-20
Fix mapping of repairProgress metrics

max-melentyev:
It has [3 variable segments](https://github.com/thelastpickle/cassandra-reaper/blob/7cc571bbc05f65260aea67012a89c56c803fee66/src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java#L130-L135) but MapperConfig was trying to match only 2 of them.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/metrics/PrometheusMetricsConfiguration.java
+++ b/src/server/src/main/java/io/cassandrareaper/metrics/PrometheusMetricsConfiguration.java
@@ -44,6 +44,8 @@ public final class PrometheusMetricsConfiguration {
     final Map<String,String> millisSinceLastRepairMetricLabels = new HashMap<>();
     millisSinceLastRepairMetricLabels.put("cluster", "${0}");
     millisSinceLastRepairMetricLabels.put("keyspace", "${1}");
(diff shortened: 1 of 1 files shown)


src/server/src/main/java/io/cassandrareaper/metrics/PrometheusMetricsConfiguration.java
```diff
@@ -44,6 +44,8 @@ public static CustomMappingSampleBuilder getCustomSampleMethodBuilder() {
     final Map<String,String> millisSinceLastRepairMetricLabels = new HashMap<>();
     millisSinceLastRepairMetricLabels.put("cluster", "${0}");
     millisSinceLastRepairMetricLabels.put("keyspace", "${1}");
+    millisSinceLastRepairMetricLabels.put("repairid", "${2}");
+    // Legacy label replaced with repairid
```
max-melentyev, line 48: In other metrics same value is mapped to repairid:
https://github.com/thelastpickle/cassandra-reaper/blob/7cc571bbc05f65260aea67012a89c56c803fee66/src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java#L144-L159

adejanovski (approved):

PR #1262, opened 2023-01-10
Improve status reporting in UI

### LINKED ISSUE #1217: Force running repairs to always show up in Running on Repairs page
Sometimes I have repair segements that will hang and need to be aborted and retried (almost always completing successfully on the second try) but with the default display set to 10, it can hide running repairs that are stuck until changing the `Display:`.. is there a way to force repairs in State: RUNNING to always show up in the Repairs tab?

## Depends upon

- [x] #1253 

┆Issue is synchronized with this [Jira Task](https://k8ssandra.atlassian.net/browse/K8SSAND-1719) by [Unito](https://www.unito.io)
┆friendlyId: K8SSAND-1719
┆priority: Medium

Miles-Garnsey:
The status page doesn't always show repairs which are actually running. We should sort the returned results according to which statuses are most "interesting" and show the most interesting rows first.

**Fixes**

#1217


adejanovski (commented):
Here's the Cucumber scenario to verify the feature is correctly implemented:

```
@sidecar
  Scenario Outline: Verify that ongoing repairs are prioritized over finished ones when listing the runs
    Given that reaper <version> is running
    And reaper has no cluster in storage
    When an add-cluster request is made to reaper with authentication
    Then reaper has the last added cluster in storage
    And a new repair is added for "test" and keyspace "test_keyspace"
    And I add and abort 10 repairs for "test" and keyspace "test_keyspace2"
    Then when I list the last 10 repairs, I can see 1 repairs at "NOT_STARTED" state
    And when I list the last 10 repairs, I can see 9 repairs at "ABORTED" state
    When the last added cluster is deleted
  ${cucumber.upgrade-versions}
```

The 2 new steps to implement are:
```
And I add and abort ?? repairs for "???" and keyspace "???"
Then when I list the last ?? repairs, I can see ? repairs at "????" state
```

src/server/src/test/java/io/cassandrareaper/acceptance/BasicSteps.java
```diff
@@ -2895,4 +2895,88 @@ public void startFakeRepairSchedule() {
...
+              .assertThat(response.getStatus())
+              .isEqualTo(Response.Status.CREATED.getStatusCode())
+              .withFailMessage(responseData);
+          String id = "";
+          try {
+            id = new ObjectMapper().readValue(responseData, String.class);
+            testContext.addCurrentRepairId(UUID.fromString(id));
+          } catch (Throwable e) {
+            Assertions.fail("response deserialisation failed");
+          }
```
adejanovski: You should rather catch a RuntimeException and display its content in the output, because right now we cannot know what's the underlying error

```suggestion
          } catch (RuntimeException e) {
            LOG.error("response deserialisation failed", e);
            LOG.error("Response data was: {}", responseData);
          }
```

src/server/src/test/java/io/cassandrareaper/acceptance/BasicSteps.java
```diff
@@ -2895,4 +2895,88 @@ public void startFakeRepairSchedule() {
           Response.Status.NOT_FOUND);
     });
   }
+
+  @And("^I add 11 and abort the most recent 10 repairs for cluster \"([^\"]*)\" and keyspace \"([^\"]*)\"$")
+  public void addAndAbortRepairs(String clusterName, String keyspace) throws Throwable {
```
adejanovski: `11` and `10` should be variables, so that we can reuse this step if needed with different numbers.

```suggestion
  @And("^I add (\\d+) and abort the most recent (\\d+) repairs for cluster \"([^\"]*)\" and keyspace \"([^\"]*)\"$")
  public void addAndAbortRepairs(int nbRepairRuns, int abortedRepairRuns, String clusterName, String keyspace) throws Throwable {
```

src/server/src/test/java/io/cassandrareaper/acceptance/BasicSteps.java
```diff
@@ -2895,4 +2895,88 @@ public void startFakeRepairSchedule() {
...
+      Set<String> tables = Sets.newHashSet();
+      testContext.addClusterInfo(clusterName, keyspace, tables);
+      HashMap<String, String> params = Maps.newHashMap();
+      params.put("clusterName", clusterName);
+      params.put("keyspace", keyspace);
+      params.put("owner", "test_user");
+
+      RUNNERS.parallelStream().forEach(runner -> {
+        Integer iter = 1;
+        while (iter <= 11) {
```
adejanovski: You seem to be creating just 10 repairs here, not 11.
Also I don't think you should use a parallel stream across runners, because each runner will create 10 runs here.
Just pick the first runner by using `RUNNERS.get(0).callReaper(...)` at line 2913, and remove the surrounding stream.

src/server/src/test/java/io/cassandrareaper/acceptance/BasicSteps.java
Miles-Garnsey: Doing:

```
            LOG.error("response deserialisation failed", e);
            LOG.error("Response data was: {}", responseData);
            Assertions.fail("response deserialisation failed");
```

Because we still want it to fail right?

src/server/src/test/java/io/cassandrareaper/acceptance/BasicSteps.java
Miles-Garnsey: I was wondering about that. The runners all run against the same cluster? Changed as suggested.

Miles-Garnsey: RE creating too few repairs, with this logic I create 1,2,3,4,5,6,7,8,9,10,11 - which is 11 I'm pretty sure? 

I used my fingers to count, so someone needs to rescind my math postgrad if I'm wrong about this.

src/server/src/test/java/io/cassandrareaper/acceptance/BasicSteps.java
```diff
@@ -2915,18 +2915,19 @@ public void addAndAbortRepairs(String clusterName, String keyspace) throws Throw
...
             .isEqualTo(Response.Status.CREATED.getStatusCode())
             .withFailMessage(responseData);
-        String id = "";
+        UUID id = UUID.randomUUID();
         try {
-          id = new ObjectMapper().readValue(responseData, String.class);
-          testContext.addCurrentRepairId(UUID.fromString(id));
+          RepairRun repairRun = new ObjectMapper().readValue(responseData, RepairRun.class);
+          id = repairRun.getId();
+          testContext.addCurrentRepairId(id);
```
adejanovski: I checked the RepairRun resource and how we deserialize responses from the API in the tests and understood what's wrong here.

A POST operation to the "/repair_run" endpoint will return a RepairRunStatus object.
To deserialize it, you'll have to do the same as is done on line 1174:

```
RepairRunStatus run = SimpleReaperClient.parseRepairRunStatusJSON(responseData);
```

then you can access `run.getId()`.

```
          RepairRunStatus run = SimpleReaperClient.parseRepairRunStatusJSON(responseData);
          id = run.getId();
          ...
```

adejanovski (changes requested):

src/server/src/test/java/io/cassandrareaper/acceptance/BasicSteps.java
Miles-Garnsey: Thanks for clarifying, I thought it was a RepairRun object... I've made the required changes and serialisation now appears to work OK!

src/server/src/test/java/io/cassandrareaper/acceptance/BasicSteps.java
Miles-Garnsey: I've made the requested changes. Thinking on this, should this not be several steps? So perhaps the spec should be:
* When I add 11 repairs
* When I set the state on the most recent 10 repairs to ABORTED

That would better promote reusability I think.

src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
```diff
@@ -608,7 +608,9 @@ public Response getRepairRunsForCluster(
       @QueryParam("limit") Optional<Integer> limit) {
 
     LOG.debug("get repair run for cluster called with: cluster_name = {}", clusterName);
-    final Collection<RepairRun> repairRuns = context.storage.getRepairRunsForCluster(clusterName, limit);
+    final Collection<RepairRun> repairRuns = context
+        .storage
+        .getRepairRunsForClusterPrioritiseRunning(clusterName, limit);
```
adejanovski, line 615: You need to do the same on line 665. It's the method that lists repair runs throughput all registered clusters.
Currently it's still using the old `getRepairRunsForCluster()`  method which isn't optimized to prioritize running repairs.

src/server/src/main/java/io/cassandrareaper/storage/MemoryStorage.java
```diff
@@ -193,6 +197,24 @@ public List<RepairRun> getRepairRunsForCluster(String clusterName, Optional<Inte
...
+        .filter(
+            row -> row.getClusterName().equals(clusterName.toLowerCase(Locale.ROOT))).collect(Collectors.toList()
+        );
+    Comparator<RepairRun> comparator = new Comparator<RepairRun>() {
+      @Override
+      public int compare(RepairRun o1, RepairRun o2) {
+        return o1.getRunState().compareTo(o2.getRunState());
+      }
+    };
+    Collections.sort(foundRepairRuns, comparator);
```
adejanovski: This has one annoying side effect, which is that DONE runs will be prioritized over ABORTED ones, even if the ABORTED run is more recent than the DONE one.
That's going to be misleading for users which aren't going to find what they're looking for.
Would it be possible to create a `RunStateSet`  class of some sorts, which would group the states in two distinct sets? Something like `NOT_FINISHED` (`RUNNING`, `PAUSED`, `NOT_STARTED`) and `FINISHED` (the other states). And then we could use it as comparator to avoid reordering the runs within each set.
We need to retain the natural order within each list.

src/server/src/main/resources/db/cassandra/032_add_2i_status.cql
```diff
@@ -0,0 +1,18 @@
...
+--
+--  Unless required by applicable law or agreed to in writing, software
+--  distributed under the License is distributed on an "AS IS" BASIS,
+--  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+--  See the License for the specific language governing permissions and
+--  limitations under the License.
+--
+-- Add a secondary index on `state` to the `repair_run` table.
+
+CREATE INDEX state2i ON repair_run_by_cluster_v2 (repair_run_state);
```
adejanovski:
```suggestion
CREATE INDEX IF NOT EXISTS state2i ON repair_run_by_cluster_v2 (repair_run_state);
```

src/server/src/test/java/io/cassandrareaper/acceptance/BasicSteps.java
```diff
@@ -2895,4 +2895,109 @@ public void startFakeRepairSchedule() {
...
+
+  @Then("^when I list the last (\\d+) repairs, I can see (\\d+) repairs at (\\d+) state$")
+  public void listRepairs(Integer limit, Integer expectedRepairsCount, String expectedState) {
+    synchronized (BasicSteps.class) {
+      RUNNERS.parallelStream().forEach(runner -> {
+        HashMap<String, String> params = Maps.newHashMap();
+        params.put("limit", limit.toString());
+        Response resp = runner.callReaper(
+            "GET",
+            "/repair_run/cluster/" + TestContext.TEST_CLUSTER,
```
adejanovski, line 2970: You may want to test `"/repair_run"` to cover both endpoints that allow listing repairs.

adejanovski (changes requested):
Almost there, I still have a few request so that we don't create confusion amongst users.

src/server/src/main/java/io/cassandrareaper/storage/MemoryStorage.java
Miles-Garnsey: > We need to retain the natural order within each list.

By "natural ordering", you mean time based ordering? I feel like we keep miscommunication on this bit 😬

src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
Miles-Garnsey: The problem is that on line 665, it is adding the runs from ALL clusters into a list. I'll need to re-sort the runs so that the statuses are in the right order irrespective of cluster. Once you let me know the precise way you want the sorting done, I might just encapsulate it in a function so that we aren't duplicating that logic in multiple places.

src/server/src/main/java/io/cassandrareaper/storage/MemoryStorage.java
adejanovski: yes, time based ordering.

src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
adejanovski: Right, you'll need to merge lists from different clusters and re-sort to apply the limits.
FYI, repair ids are timeuuids, which makes it possible to apply the sorting on them directly using their time component.

src/server/src/main/java/io/cassandrareaper/storage/MemoryStorage.java
adejanovski: Repair run ids are timeuuid, that should help if you need to sort them in descending order.

Miles-Garnsey:
I've created a new method on `RepairRun` - `SortByRunState` which implements the logic you want using the existing function `isTerminated()`. I've then subbed that method in for `MemoryStorage` and inside the `RepairRunResource` calls (CassandraStorage uses different logic which already partitions the terminated/unterminated statuses, so I think that is OK but please let me know if I'm wrong). 

My only concern with this is that I don't like having methods on RepairRun since it doesn't have a lot right now and is almost a model object (having few methods beyond a `builder`); is there a better place for this do you think? Maybe even in RepairRunResource, but then the storage layer would need to call the presentation layer, which seems questionable too?

Maybe we shouldn't be doing the sorting in the storage layer at all since this type of sorting is purely a presentation layer concern (so perhaps implement in 'RepairRunResource' and leave responses from storage layers unsorted)? I'm open to feedback on the design here, as you know the patterns in this codebase better than me.

src/server/src/main/java/io/cassandrareaper/storage/MemoryStorage.java
Miles-Garnsey: All done, tests passing.

src/server/src/test/java/io/cassandrareaper/acceptance/BasicSteps.java
Miles-Garnsey: Also done, and appears to be passing.

src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
Miles-Garnsey: This appears to be done and working now, although the cucumber test doesn't test against multiple clusters. I'm not sure if I should add a multi-cluster test to confirm that this endpoint does indeed work?

src/server/src/main/java/io/cassandrareaper/storage/MemoryStorage.java
Miles-Garnsey: One thing about sorting on timeuuids: it looks like right now repairs in the UI are sorted by start time, not creation time (which I think the timeuuid corresponds to). Do you want to update this behaviour to sort of creation time?

Miles-Garnsey:
Manual testing suggests that this isn't working again for some reason. I've tried flipping the ordering in these conditions, but neither ordering appears to give us the result we want in the UI:

```
  if (!o1.getRunState().isTerminated() && o2.getRunState().isTerminated()) {
          return 1; // o2 appears first.
        }  else if (o1.getRunState().isTerminated() && !o2.getRunState().isTerminated()) {
          return -1; // o1 appears first.
```

More concerning, I haven't seen the cucumber tests failing either way, so they don't appear to be detecting errors. 

It may be the case that there is some ordering being applied in the UI, since I note that the repair runs are always ordered by start time (not creation time, which UUID should give us, I think?) I'll investigate further.

src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
```diff
@@ -663,17 +663,15 @@ public Response listRepairRuns(
...
-        runStatuses.addAll(
-            (List<RepairRunStatus>) getRunStatuses(runs, desiredStates)
-                .stream()
-                .filter((run) -> !keyspace.isPresent()
-                    || ((RepairRunStatus)run).getKeyspaceName().equals(keyspace.get()))
-                .collect(Collectors.toList()));
-      }
-
+      List<RepairRun> repairRuns = Lists.newArrayList();
+      clusters.forEach(clstr -> repairRuns.addAll(context.storage.getRepairRunsForCluster(clstr.getName(), limit)));
```
adejanovski: I think we should call `context.storage.getRepairRunsForClusterPrioritiseRunning()` running here, otherwise the limits will be applied as before and any long running repair could get hidden by more recent finished runs.

src/server/src/main/java/io/cassandrareaper/core/RepairRun.java
```diff
@@ -344,4 +350,20 @@ public RepairRun build(UUID id) {
       return new RepairRun(this, id);
     }
   }
+
+    public static void SortByRunState (List<RepairRun> repairRunCollection) {
```
adejanovski: The RepairRun class is a Java Bean, so it shouldn't define such static methods indeed.
Given how the code is organized in this project, I think this method rather belongs in the RepairRunService class.

adejanovski (changes requested):

src/server/src/main/java/io/cassandrareaper/core/RepairRun.java
adejanovski: Note that we could also define it as a [default method](https://www.baeldung.com/java-static-default-methods) in the IStorage interface.

Miles-Garnsey: I'll put it in the RepairRunService, that sounds like a good idea. I can see an argument for putting it in IStorage, depends whether you see sorting as a data access consideration, or as a presentation consideration.

src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
Miles-Garnsey: Ah, I think that's why my manual tests were failing! Agreed, I'll fix this. Thanks for the suggestion.

Miles-Garnsey:
Manual testing confirms this PR works as intended. The outstanding questions are:

1. Should we restructure the cucumber test functions so that starting and aborting repairs are separate functions?
2. Should we make this a multicluster test to better cover the `repair_runs/` endpoint (which queries for all clusters, and isn't currently fully covered by the e2e tests).

src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
adejanovski: I think that would be fairly hard to achieve with the limited resources we have at our disposal.
A mocked test wouldn't help much I guess, so we're left with manual testing for multi cluster 🤷

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
```diff
@@ -957,6 +963,76 @@ public Collection<RepairRun> getRepairRunsForCluster(String clusterName, Optiona
...
+    // We've set up the RunState enum so that values are declared in order of "interestingness",
+    // we iterate over the table via the secondary index according to that ordering.
+    for (RunState state :
+        Arrays
+            .stream(RunState.values())
+            .filter(v ->
+                Arrays.asList("RUNNING", "PAUSED", "NOT_STARTED")
+                    .contains(v.toString()))
+            .collect(Collectors.toList())
+    ) {
```
adejanovski: **suggestion (non-blocking):** Do you think this could be simplified to the following?

```
for (String state:Arrays.asList("RUNNING", "PAUSED", "NOT_STARTED")){
```

I'm not sure why we need the stream().filter().collect() here, but I could be missing something.

adejanovski (dismissed):
I have a suggestion to simplify the code a little bit.
No need to try testing multi cluster stuff in cucumber, we won't have enough resources in GHA to do so.
Also, it's fine to keep create/abort in the same cucumber step for now. If we need it later, we can refactor it in two separate steps.

src/server/src/main/java/io/cassandrareaper/storage/CassandraStorage.java
Miles-Garnsey: We could, it would be simpler. I'll make this change.

adejanovski (approved):

PR #1270, opened 2023-02-01
Fix running concurrent repairs on different clusters

### LINKED ISSUE #1269: Apply concurrent repair run limits per cluster
As reported [here](https://github.com/thelastpickle/cassandra-reaper/pull/1197#issuecomment-1411973745), the repair run concurrency limit is applied globally instead of per cluster.

We need to make it so that this limit is applied per cluster so that we have the ability to control the repair pressure per cluster while avoiding having some clusters repairs block others.

max-melentyev:
After #1197 active repairs in one cluster prevent repairs in others. This diff makes repairs in different clusters independent again.

Fixes #1269

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
+++ b/src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
@@ -204,10 +204,11 @@ final class RepairRunner implements Runnable {
   public void run() {
     Thread.currentThread().setName(clusterName + ":" + repairRunId);
     Map<UUID, RepairRunner> currentRunners = context.repairManager.repairRunners;
-    // We only want the repair runners that are in RUNNING state.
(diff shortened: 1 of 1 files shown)


adejanovski (changes requested):
Hi @max-melentyev,

thanks for pushing this PR! It looks like an easy enough fix 😅 

Do you think it would be possible to add unit tests for this change?
I'm thinking of a modification of `isAllowedToRunTest()` and `isNotAllowedToRunTest()` in `RepairRunnerTest.java` which would check the new behavior by adding runs for a second cluster.

max-melentyev:
I thought about this, but existing tests invoke only isAllowedToRun() while changes are in the run() method. I didn't find tests for run() and I'm not that familiar with code and java in general to write such test from scratch :)

However I've tested this PR along with #1255 and #1267 on multiple clusters manually.

adejanovski (approved):

Issue #1038, opened 2021-03-17
java.lang.AssertionError: Unknown keyspace test_keyspace


therb1:
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=13951848)

BUG?

to reproduce the bug you need:
1) Start keyspace test_keyspace repair in webui
2) Stop cassandra-reaper service
2) DROP keyspace that is already in the process of being repaired
3) Start cassandra-reaper

```

Mar 17 01:13:17 server.example.com cassandra-reaper[30949]: com.google.common.util.concurrent.ExecutionError: java.lang.AssertionError: Unknown keyspace test_keyspace
Mar 17 01:13:17 server.example.com cassandra-reaper[30949]:         at com.google.common.cache.LocalCache$Segment.get(LocalCache.java:2216)
Mar 17 01:13:17 server.example.com cassandra-reaper[30949]:         at com.google.common.cache.LocalCache.get(LocalCache.java:4147)
Mar 17 01:13:17 server.example.com cassandra-reaper[30949]:         at com.google.common.cache.LocalCache$LocalManualCache.get(LocalCache.java:5053)
Mar 17 01:13:17 server.example.com cassandra-reaper[30949]:         at io.cassandrareaper.jmx.ClusterFacade.getRangeToEndpointMap(ClusterFacade.java:276)
Mar 17 01:13:17 server.example.com cassandra-reaper[30949]:         at io.cassandrareaper.service.RepairRunner.<init>(RepairRunner.java:92)
Mar 17 01:13:17 server.example.com cassandra-reaper[30949]:         at io.cassandrareaper.service.RepairRunner.create(RepairRunner.java:170)
Mar 17 01:13:17 server.example.com cassandra-reaper[30949]:         at io.cassandrareaper.service.RepairManager.startRunner(RepairManager.java:362)
Mar 17 01:13:17 server.example.com cassandra-reaper[30949]:         at io.cassandrareaper.service.RepairManager.startRepairRun(RepairManager.java:328)
Mar 17 01:13:17 server.example.com cassandra-reaper[30949]:         at io.cassandrareaper.service.RepairManager.resumeUnkownRunningRepairRuns(RepairManager.java:167)
Mar 17 01:13:17 server.example.com cassandra-reaper[30949]:         at io.cassandrareaper.service.RepairManager.resumeRunningRepairRuns(RepairManager.java:138)
[...]

PR #1273, opened 2023-02-07
Changes Exception to a  Reaper Exception

xgerman:
This should mitigate #1038 

Instead of crashing the JVM we are wrapping the error into a ReaperException to mitigate the issue.


xgerman:
Will fix checkstyle violations when this looks like a patch we want to pursue...

adejanovski:
Hi @xgerman, we can move forward with this PR and merge it once you get CI to pass 👍

adejanovski (approved):
Thanks @xgerman !

PR #1274, opened 2023-02-07
Fix docker build

max-melentyev:
- Support for different host arch
- Run bower explicitly instead of from npm postinstall hook. Hook requires --unsafe-perm option now.
- Build packages with `-a all` as there is no arch dependent code.


adejanovski (approved):
Thanks @max-melentyev !

PR #1306, opened 2023-06-08
Further storage refactoring

Miles-Garnsey:
This PR continues refactoring the storage classes. It breaks the MemoryStorage class out into delegated classes, turning MemoryStorage into MemoryStorageFacade. CassandraStorage is likewise renamed CassandraStorageFacade since most of its methods are now also delegates (there are some residual ones that we will deal with in the next round of refactoring).

The DAOs are now lined up on the domain model from `core`, with each sitting in its own package containing:

1. An interface.
2. A Cassandra DAO class implementing the interface.
3. A memory based DAO implementing the interface

IStorage has been substantially broken down and now extends a variety of smaller interfaces again adhering to the domain model from the `core` package.

To preserve the existing semantics for callers, I have retained MemoryStorageFacade and CassandraStorageFacade, which we will attempt to now completely eliminate in the next round of refactoring. Several methods have been eliminated from these classes where they were only required internally within the DAO.

We also fix a bug where the implementation of `deleteRepairRun` differed between the memory and Cassandra implementations.

Still to do:

1. This PR does not touch anything from IDistributedStorage, that needs to be straightened out as some methods listed there have made their way into CassRepairSegmentDao.
2. Eliminate the use of CassandraStorageFacade and MemoryStorageFacade, and replace IStorage uses with the smaller interfaces which map to the domain model. This will mostly affect the classes in `resources`.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/storage/JsonParseUtils.java
+++ b/src/server/src/main/java/io/cassandrareaper/storage/JsonParseUtils.java
@@ -38,10 +38,11 @@ public final class JsonParseUtils {
 
   private static final Logger LOG = LoggerFactory.getLogger(JsonParseUtils.class);
 
-  private static final ObjectReader LIST_READER
-      = new ObjectMapper().readerFor(new TypeReference<List<RingRange>>() {});
-  private static final ObjectReader MAP_READER
-      = new ObjectMapper().readerFor(new TypeReference<Map<String, String>>() {});
+  private static final ObjectReader LIST_READER  = new ObjectMapper()
+        .readerFor(new TypeReference<List<RingRange>>() {});
+  private static final ObjectReader MAP_READER = new ObjectMapper()
+        .readerFor(new TypeReference<Map<String, String>>() {
+        });
   private static final ObjectWriter WRITER = new ObjectMapper().writer();
 
   private JsonParseUtils() {
--- a/src/server/src/main/java/io/cassandrareaper/storage/JsonParseUtils.java
+++ b/src/server/src/main/java/io/cassandrareaper/storage/JsonParseUtils.java
@@ -38,11 +38,11 @@ public final class JsonParseUtils {
 
   private static final Logger LOG = LoggerFactory.getLogger(JsonParseUtils.class);
 
-  private static final ObjectReader LIST_READER  = new ObjectMapper()
-        .readerFor(new TypeReference<List<RingRange>>() {});
-  private static final ObjectReader MAP_READER = new ObjectMapper()
-        .readerFor(new TypeReference<Map<String, String>>() {
-        });
+  private static final ObjectReader LIST_READER
+      = new ObjectMapper().readerFor(new TypeReference<List<RingRange>>() {});
+  private static final ObjectReader MAP_READER
+      = new ObjectMapper().readerFor(new TypeReference<Map<String, String>>() {});
+
   private static final ObjectWriter WRITER = new ObjectMapper().writer();
 
   private JsonParseUtils() {
--- a/src/server/src/main/java/io/cassandrareaper/storage/MemoryStorageFacade.java
+++ b/src/server/src/main/java/io/cassandrareaper/storage/MemoryStorageFacade.java
@@ -67,7 +67,7 @@ public final class MemoryStorageFacade implements IStorage {
   private final MemRepairSegment memRepairSegment = new MemRepairSegment(this);
   private final MemRepairUnitDao memRepairUnitDao = new MemRepairUnitDao();
   private final MemRepairRunDao memRepairRunDao = new MemRepairRunDao(memRepairSegment, memRepairUnitDao);
-  private final MemRepairScheduleDao memRepairScheduleDao = new MemRepairScheduleDao(this);
+  private final MemRepairScheduleDao memRepairScheduleDao = new MemRepairScheduleDao(memRepairUnitDao);
 
   @Override
   public boolean isStorageConnected() {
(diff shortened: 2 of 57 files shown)


src/server/src/main/java/io/cassandrareaper/storage/JsonParseUtils.java
```diff
@@ -38,10 +38,11 @@ public final class JsonParseUtils {
...
 
-  private static final ObjectReader LIST_READER
-      = new ObjectMapper().readerFor(new TypeReference<List<RingRange>>() {});
-  private static final ObjectReader MAP_READER
-      = new ObjectMapper().readerFor(new TypeReference<Map<String, String>>() {});
+  private static final ObjectReader LIST_READER  = new ObjectMapper()
+        .readerFor(new TypeReference<List<RingRange>>() {});
+  private static final ObjectReader MAP_READER = new ObjectMapper()
+        .readerFor(new TypeReference<Map<String, String>>() {
+        });
```
adejanovski: **suggestion:** style looked better before and is now inconsistent between the two lines.

src/server/src/main/java/io/cassandrareaper/storage/MemoryStorageFacade.java
```diff
@@ -0,0 +1,331 @@
...
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Implements the StorageAPI using transient Java classes.
+ */
+public final class MemoryStorageFacade implements IStorage {
+
+  private static final Logger LOG = LoggerFactory.getLogger(MemoryStorageFacade.class);
+  private final MemRepairSegment memRepairSegment = new MemRepairSegment(this);
```
adejanovski, line 55: **nitpick:** naming of memory storage related classes could maybe stick with the `Memory` prefix instead of `Mem` for consistency. Java conventions allow long and expressive names, so it would probably be fine.

adejanovski: This comment applies to all `mem...`  objects and `Mem...` classes referenced here.

src/server/src/main/java/io/cassandrareaper/storage/cassandra/CassandraStorageFacade.java
```diff
@@ -81,30 +87,30 @@
...
+  private final CassRepairRunDao cassRepairRunDao;
   private PreparedStatement saveHeartbeatPrepStmt;
   private PreparedStatement deleteHeartbeatPrepStmt;
 
-  private final RepairUnitDao repairUnitDao;
-  private final RepairScheduleDao repairScheduleDao;
-  private final ClusterDao clusterDao;
-  private final EventsDao eventsDao;
-  private final MetricsDao metricsDao;
+  private final CassRepairUnitDao cassRepairUnitDao;
```
adejanovski, line 104: **nitpick:** Same comment than for Memory classes, using `Cassandra` instead of `Cass` could improve consistency in object namings.

src/server/src/main/java/io/cassandrareaper/storage/cassandra/migrations/Migration025.java
```diff
@@ -45,10 +45,10 @@ public static void migrate(Session session, String keyspace) {
             "INSERT INTO " + V2_TABLE + "(cluster_name, id, repair_run_state) values (?, ?, ?)");
         LOG.info("Converting {} table...", V1_TABLE);
         ResultSet results = session.execute("SELECT * FROM " + V1_TABLE);
-        for (Row row:results) {
+        for (Row row : results) {
```
adejanovski, line 48: **suggestion:** the style of for loops is inconsistent throughout the codebase. It would be nice to harmonize it and add a checkstyle rule for it. Not in the scope of this PR though.

src/server/src/main/java/io/cassandrareaper/storage/cluster/MemClusterDao.java
```diff
@@ -0,0 +1,129 @@
...
+    return true;
+  }
+
+  public boolean addClusterAssertions(Cluster cluster) {
+    Preconditions.checkState(
+          Cluster.State.UNKNOWN != cluster.getState(),
+          "Cluster should not be persisted with UNKNOWN state");
+
+    // TODO – unit tests need to also always set the paritioner
+    //Preconditions.checkState(cluster.getPartitioner().isPresent(), "Cannot store cluster with no partitioner.");
```
adejanovski, line 78: **Suggestion:** Maybe we could delete these lines if they were carried over from the existing code?

```suggestion
```

adejanovski (dismissed):
Looks good 👍 
I've left a few non blocking comments. Check them and let me know what you think.

src/server/src/main/java/io/cassandrareaper/storage/cassandra/migrations/Migration025.java
Miles-Garnsey: I'm not going to get into this one this time. Seems like a big job.

src/server/src/main/java/io/cassandrareaper/storage/cluster/MemClusterDao.java
Miles-Garnsey: Done!

src/server/src/main/java/io/cassandrareaper/storage/JsonParseUtils.java
Miles-Garnsey: Fixed.

src/server/src/main/java/io/cassandrareaper/storage/MemoryStorageFacade.java
Miles-Garnsey: I'm happy to make this change in the next PR if that's OK with you. It will take a bit of work and seems like it should be undertaken in a refactoring week so as not to distract from other work.

src/server/src/main/java/io/cassandrareaper/storage/cassandra/CassandraStorageFacade.java
Miles-Garnsey: I'm happy to make this change in the next PR if that's OK with you. It will take a bit of work and seems like it should be undertaken in a refactoring week so as not to distract from other work.

adejanovski (approved):

PR #1307, opened 2023-06-08
Fix issues when listing repair runs using state priorities

### LINKED ISSUE #1272: Reaper web not displaying all running repairs
For some reason webpage of reaper not displaying all of running jobs. In this example. Default 10 row displaying 4 jobs only, extending pagination to 25 and  2 more jobs on list. Missing 2 jobs which stuck(issue reported in different task #1271).
Cassandra Reaper 3.2.1 version
reaper less
reaper_more

adejanovski:
Fixes #1272

Several changes have been made to the `RepairRunResource.java` file, including updating the
`getRepairRunsForCluster` method to use `clusterName` instead of `cluster_name` as a path
parameter, updating the `listRepairRuns` method to use `clusterName` instead of `cluster_name`
as a query parameter. These changes fix bugs that have been existing for a while it seems, but were hidden by how repairs were listed by pure chronological order and client side filters applied in the UI.

In the RepairSegmentDao.java file, the `getSegmentAmountForRepairRun` method has
been updated to use `getRepairSegmentCountByRunIdPrepStmt` instead of
`getRepairSegmentCountByRunIdAndStatePrepStmt`, which was a bug.

A bug was fixed in the `RepairRunDao.getRepairRunsForClusterPrioritiseRunning()` method, which was not really applying a limit on the `flattenedUuids` list as  `subList()` returns a new list, it doesn't modify in place the list it's applied to. 

Additionally, this pull request includes updates to the acceptance tests to add fake
clusters and verify that ongoing repairs are prioritized over finished ones when listing them.
This checks that repair runs across multiple clusters are prioritized based on their state and that depending on the filters applied, we get the expected runs displayed.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
+++ b/src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
@@ -655,7 +655,7 @@ public final class RepairRunResource {
       @QueryParam("limit") Optional<Integer> limit) {
 
     try {
-      final Set desiredStates = splitStateParam(state);
+      final Set<String> desiredStates = splitStateParam(state);
       if (desiredStates == null) {
         return Response.status(Response.Status.BAD_REQUEST).build();
       }
@@ -664,14 +664,14 @@ public final class RepairRunResource {
             ? Collections.singleton(context.storage.getCluster(cluster.get()))
             : context.storage.getClusters();
 
-      List<RepairRunStatus> runStatuses = Lists.newArrayList();
       List<RepairRun> repairRuns = Lists.newArrayList();
       clusters.forEach(clstr -> repairRuns.addAll(
           context.storage.getRepairRunsForClusterPrioritiseRunning(clstr.getName(), limit))
       );
+      List<RepairRunStatus> runStatuses = Lists.newArrayList();
       RepairRunService.sortByRunState(repairRuns);
       runStatuses.addAll(
-          (List<RepairRunStatus>) getRunStatuses(
+          getRunStatuses(
               repairRuns.subList(0, min(repairRuns.size(), limit.orElse(1000))), desiredStates)
               .stream()
               .filter((run) -> !keyspace.isPresent()
@@ -700,7 +700,7 @@ public final class RepairRunResource {
     return runStatuses;
   }
 
-  static Set splitStateParam(Optional<String> state) {
+  static Set<String> splitStateParam(Optional<String> state) {
     if (state.isPresent()) {
       final Iterable<String> chunks = RepairRunService.COMMA_SEPARATED_LIST_SPLITTER.split(state.get());
       for (final String chunk : chunks) {
--- a/src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
+++ b/src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
@@ -650,7 +650,7 @@ public final class RepairRunResource {
   @GET
   public Response listRepairRuns(
       @QueryParam("state") Optional<String> state,
-      @QueryParam("clusterName") Optional<String> cluster,
+      @QueryParam("cluster_name") Optional<String> cluster,
       @QueryParam("keyspace_name") Optional<String> keyspace,
       @QueryParam("limit") Optional<Integer> limit) {
 
--- a/src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
+++ b/src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
@@ -604,9 +604,9 @@ public final class RepairRunResource {
    * @return all know repair runs for a cluster.
    */
   @GET
-  @Path("/cluster/{cluster_name}")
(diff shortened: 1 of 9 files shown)


src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
```diff
@@ -604,9 +604,9 @@ public Response abortRepairRunSegment(@PathParam("id") UUID repairRunId, @PathPa
    * @return all know repair runs for a cluster.
    */
   @GET
-  @Path("/cluster/{cluster_name}")
+  @Path("/cluster/{clusterName}")
```
Miles-Garnsey, line 607: Suggestion: I think this might be the wrong way around. Generally shouldn't URI query strings use snake case? Would it be better to change this in the UI?

Issue: Also, as this is a breaking change to the API I'm not sure what our approach should be here. Does the API have a versioning policy?

src/server/src/main/java/io/cassandrareaper/storage/cassandra/RepairSegmentDao.java
```diff
@@ -175,7 +175,7 @@ private void prepareStatements() {
 
   public int getSegmentAmountForRepairRun(UUID runId) {
     return (int) session
-        .execute(getRepairSegmentCountByRunIdAndStatePrepStmt.bind(runId))
+        .execute(getRepairSegmentCountByRunIdPrepStmt.bind(runId))
```
Miles-Garnsey, line 178: Question: It looks like this didn't actually have enough parameters bound, how was it ever running?

Miles-Garnsey (commented):
I have a few questions, issues/suggestions here. I haven't manually tested yet pending what you'd like to do about those. 

I'm also confused how the manual testing we did on the last PR failed to catch this issue.

The manual testing I did basically mirrored what was in the previous acceptance test - do you know why that failed to catch the problems? I don't want to miss another issue this time around...

src/server/src/main/java/io/cassandrareaper/storage/cassandra/RepairSegmentDao.java
adejanovski: it was failing :)

src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
adejanovski: This one is a pathParam, so it doesn't matter that much. The query param from listRepairRuns on the other hand could indeed be a breaking change.
Let me see if I can turn this around.

src/server/src/main/java/io/cassandrareaper/storage/cassandra/RepairSegmentDao.java
Miles-Garnsey: I don't recall ever seeing it in the logs, and I know manual testing worked. Maybe that was on the other endpoint - weird.

src/server/src/test/resources/io.cassandrareaper.acceptance/integration_reaper_functionality.feature
```diff
@@ -415,16 +415,26 @@ Feature: Using Reaper
     Then reaper has no longer the last added cluster in storage
   ${cucumber.upgrade-versions}
 
-@sidecar
+  @sidecar
+  @current_test
```
Miles-Garnsey: I think you probably want to remove this before merging.

adejanovski: I totally do 👍

Miles-Garnsey (dismissed):
I've had a tough time testing this, partially because the current behaviour is so dependant on what order particular repair runs were added in, as well as whether all clusters vs a specific cluster are filtered on. 

When testing I:

1. Built reaper from a clean state on this branch.
2. Ran the new test.
3. Switched to main, rebuilt and ran Reaper. 

Looking at the assertions [here](https://github.com/thelastpickle/cassandra-reaper/pull/1307/files#diff-73bc53f3d7c182329cc414bbd8f4b38ea5615e3b87856103418e08d0eb4c9799R427) to compare my results I found that;

1. Without any filtering on cluster, I got the expected number of paused repairs showing up (I didn't check that the number of ABORTED repairs was corrrect).
2. When filtering on `fake1` cluster, I got the expected display of repairs, both paused and aborted.
3. When filtering on cluster `test`, I got the expected paused repairs, but no aborted repairs, which is indeed wrong.
4. When filtering on `fake2` I get one paused and only 3 aborted repairs, where it should have been 1 and 9 respectively.

Given this behaviour on `master`, I am not convinced that we are necessarily replicating the issues described in the ticket with the new acceptance test, since the issue described no filtering on the clusters. 

However, I am satisfied that we are diagnosing a bug, since we are missing aborted repairs which could be shown under the given limit settings. I have confirmed that this behaviour is resolved by the new branch, which (on my manual testing) leads to the expected results being displayed for all three clusters. 

Once the `@current_test` tag is removed from the acceptance tests, this is ready to merge.

PR #1235, opened 2022-10-27
Topology change resilience for incremental repair

Miles-Garnsey:
Ensure that, in an incremental repair, the replica list is updated on every repair run. This addresses cases where the node IPs change.

Fixes https://github.com/thelastpickle/cassandra-reaper/issues/1213

PR #1310, opened 2023-06-21
Add missing Astra migrations

### LINKED ISSUE #1309: Latest two migrations are missing for Astra
We've recently introduced new db migrations that are missing for Astra.
We need to include them so that Astra can be used as storage backend for Reaper in the latest versions.

adejanovski:
Fixes #1309


Miles-Garnsey:
Approving on the basis that these changes mirror those in #1262 and #1235, two of my PRs. I have not manually tested, please let me know if you want me to @adejanovski.

I've also raised a ticket to add tests for Astra, since there's no way for us to know whether we are breaking Astra backend functionality at present.

PR #1317, opened 2023-07-10
Storage refactor - final removal of IStorage

Miles-Garnsey:
This PR proposes to:

1. Remove all delegated functions from IStorage and its two implementations (CassandraStorageFacade and MemStorageFacade), repurposing them as StorageContext objects which just manage sessions and initialisation of the individual DAO instances.
2. Give all Service classes (and Resource classes, where required) access to the individual DAO Ifaces instead of the monolithic IStorage via DI from the ReaperApplication main class.

By the completion of this PR, there should be no further reliance on IStorage from within Resources or Services. The only case where IStorage or implementations should be used is in initialising the DAOs and passing them into the Resources from the ReaperApplication class.


Miles-Garnsey:
I'm going to have to leave this here for now due to competing priorities. The storage class is not fully removed, and things are a bit messy still. All tests pass and no functionality is broken, but there is still a lot more work to do (estimate 1-2 weeks) to make everything consistent and well structured. 

Whether we merge this depends on our priorities. Given the amount of work that may take place on Reaper in the next month or two, I lean towards thinking it would be good to merge to avoid rebases. But the current state is very inconsistent and not exactly pretty, so there is a counterargument that we should wait until the job is done properly. 

The TODO to progress looks like:

1. IDistributedStorage is in a bit of a mess, and entangled between IConcurrencyDao (yet to be defined, but we have a CassandraConcurrencyDao to work from) and IRepairSegmentDao. Attempts to disentangle are complicated by an emerging circular dependancy between segments and concurrency which requires careful thought to resolve.
2. CassandraStorageFacade is now mostly separated from IStorage, but still facades for IDistributedStorage.
3. A final step is still required to completely eliminate IStorage, which we are still using (especially in tests) to instantiate the new DAOs in the right order (there is a relatively well structured dependancy web between them which means that - especially for mocks - we need a way to ensure that they are instantiated in the right order). We need to eventually instantiate the DAOs in a more sensible way and call their methods without recourse to their IStorage parent.
4. IStorage is mostly empty at this point, but there's some more work to be done around the metrics DAO (again entangled with concurrency/distributed storage considerations).

There's a lot of work left to do here.

src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
```diff
@@ -76,10 +76,213 @@ public final class RepairRunResource {
...
+    this.repairRunService = RepairRunService.create(context, repairRunDao);
+    this.repairRunDao = repairRunDao;
+  }
+
+  /**
+   * @return Response instance in case there is a problem, or null if everything is ok.
+   * @throws ReaperException any caught runtime exception that can be re-thrown
+   */
+  @Nullable
+  static Response checkRequestForAddRepair(
```
adejanovski, line 93: **Thought**: This PR is fairly big and it would've been better not moving code around that isn't subject to DAO/storage changes. It's making the review more challenging, as there are also reformatting changes from the IDE that are obviously hard to avoid.
Unless moving these methods was necessary and I'm missing a point?

adejanovski (approved):

src/server/src/main/java/io/cassandrareaper/resources/RepairRunResource.java
Miles-Garnsey: Ah, checkstyle enforces that declarations in a class have to be in a certain order, and when you reformat code in IDEA (to achieve that) it seems it alphabetically arranges the method calls too :/ Apologies if this is making it harder. 

I can try to find a way to turn it off in future, but trying to get intellij's formatting to adhere to the huge number of checkstyle rules in this project is challenging.

Miles-Garnsey: On this topic, perhaps we should consider providing an IDEA codestyle config in this repo to make this job easier. I've lost mine a couple of times and it takes an age to configure it back to the correct state.

Issue #1314, opened 2023-07-05
Separate JmxProxy interface from the JMX implementation


Miles-Garnsey:
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=32468690)

Some installations of Reaper cannot use JMX for security/policy reasons. We will make HTTP endpoints available through the [Cassandra Management API](https://github.com/k8ssandra/management-api-for-apache-cassandra) which can replace the existing JMX endpoints that Reaper calls. 

Within the ClusterFacade is a [preHeatJmxConnections](https://github.com/thelastpickle/cassandra-reaper/blob/e11a83125fe0cfdc5bb81b0cbcc24e0fe8b960f3/src/server/src/main/java/io/cassandrareaper/jmx/ClusterFacade.java#L130) method. 

When a configuration option in the cassandra-reaper.yaml file indicates that the HTTP connection should be used instead, and an HTTP endpoint is supplied in the config, `preHeatJmxConnections` should return an HTTP based implementation of the JmxProxy interface. It is likely that we will rename this interface once it is no longer JMX dependant.

One point of complication is that both the management API and Reaper will need to collaborate to somehow replace the server-originated push pattern referenced by `NotificationListener`. To execute this, Reaper will either need to poll an endpoint on the management API for the notifications, or we will need to come up with another methodology.

It should be noted that in any case where metrics are obtained via JMX, the new metrics endpoints within management API can be used instead.

# Definition of Done

- [ ] `JmxProxy` has been renamed to `CassandraManagementProxy`, `preHeatJmxConnections` method has been renamed to `connectToManagementMechanism`.
- [ ] A new class `CassandraHttpManager` implementing `CassandraManagementProxy` exists.
- [ ] The `cassandra-reaper.yaml` configuration file accepts a new parameter `useManagementApi`, and a new parameter `managementApiPort`.
- [ ] When `useManagementApi` is specified in `cassandra-reaper.yaml` the new implementation `CassandraHttpManager` is used throughout the codebase instead of the `JmxProxyImpl` that is always currently used.
- [ ] No MBeans are available through `CassandraManagementProxy`. All of the current methods called via the MBeans and Cassandra interfaces should reside on `ICassandraManagementProxy`.

Miles-Garnsey:
Closed via #1321

PR #1321, opened 2023-07-19
HTTP endpoints groundwork

Miles-Garnsey:
Rename a number of methods and classes in preparation for providing a new HTTP management infrastructure (instead of JMX.)

I think I will keep this PR focused on refactoring rather than delivering features, and deliver the features themselves in a further PR.

At present, it is built on a base of the branch from #1317, because I believe we intend to merge that branch ahead of this one. 

Partially solves #1314.


adejanovski (approved):
Awesome big ass refactor @Miles-Garnsey 👏 
Well done!

PR #1328, opened 2023-07-27
Acceptance tests for HTTP management mechanism.

Miles-Garnsey:
The new HTTP management mechanism that relies on management API to replace JMX needs tests to ensure that all functionality still works. 

I have used the ReaperIT class as a base and replicated all tests there, these are expected to fail right now because CCM doesn't spin up Cassandra with the management API included. We will need to add this before merging.

**Solves issue**
https://github.com/thelastpickle/cassandra-reaper/issues/1347


Miles-Garnsey:
Closing since this PR forms the basis for the integration PR we've already raised.

PR #1336, opened 2023-08-08
Remove JMX specific interfaces from ICassandraManagementProxy

burmanm:
This refactors the current state of ICassandraManagementProxy a little by removing JMX specific methods from the interface as they're not going to be implemented by the HTTP server.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/ReaperApplication.java
+++ b/src/server/src/main/java/io/cassandrareaper/ReaperApplication.java
@@ -233,7 +233,7 @@ public final class ReaperApplication extends Application<ReaperApplicationConfig
 
     HttpClient httpClient = createHttpClient(config, environment);
 
-    if (context.managementConnectionFactory instanceof JmxManagementConnectionFactory) {
+    if (config.getHttpManagement() == null || !config.getHttpManagement().getEnabled()) {
       ScheduledExecutorService ses = environment.lifecycle().scheduledExecutorService("Diagnostics").threads(6).build();
       final DiagEventSubscriptionResource eventsResource = new DiagEventSubscriptionResource(context, httpClient, ses,
           context.storage.getEventsDao());
--- a/src/server/src/main/java/io/cassandrareaper/ReaperApplication.java
+++ b/src/server/src/main/java/io/cassandrareaper/ReaperApplication.java
@@ -232,13 +232,16 @@ public final class ReaperApplication extends Application<ReaperApplicationConfig
     environment.jersey().register(addCryptoResource);
 
     HttpClient httpClient = createHttpClient(config, environment);
-    ScheduledExecutorService ses = environment.lifecycle().scheduledExecutorService("Diagnostics").threads(6).build();
-    final DiagEventSubscriptionResource eventsResource = new DiagEventSubscriptionResource(context, httpClient, ses,
-        context.storage.getEventsDao());
-    environment.jersey().register(eventsResource);
-    final DiagEventSseResource diagEvents = new DiagEventSseResource(context, httpClient, ses,
-        context.storage.getEventsDao());
-    environment.jersey().register(diagEvents);
+
+    if (context.managementConnectionFactory instanceof JmxManagementConnectionFactory) {
+      ScheduledExecutorService ses = environment.lifecycle().scheduledExecutorService("Diagnostics").threads(6).build();
+      final DiagEventSubscriptionResource eventsResource = new DiagEventSubscriptionResource(context, httpClient, ses,
+          context.storage.getEventsDao());
+      environment.jersey().register(eventsResource);
+      final DiagEventSseResource diagEvents = new DiagEventSseResource(context, httpClient, ses,
+          context.storage.getEventsDao());
+      environment.jersey().register(diagEvents);
+    }
 
     if (config.isAccessControlEnabled()) {
       SessionHandler sessionHandler = new SessionHandler();
--- a/src/server/src/main/java/io/cassandrareaper/management/ICassandraManagementProxy.java
+++ b/src/server/src/main/java/io/cassandrareaper/management/ICassandraManagementProxy.java
@@ -32,17 +32,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.SortedMap;
 import java.util.concurrent.ExecutionException;
-import javax.management.AttributeList;
-import javax.management.InstanceNotFoundException;
-import javax.management.IntrospectionException;
 import javax.management.JMException;
-import javax.management.ListenerNotFoundException;
-import javax.management.MBeanInfo;
-import javax.management.NotificationFilter;
-import javax.management.NotificationListener;
-import javax.management.ObjectName;
-import javax.management.QueryExp;
-import javax.management.ReflectionException;
 import javax.management.openmbean.CompositeData;
 import javax.management.openmbean.TabularData;
 import javax.validation.constraints.NotNull;
@@ -51,7 +41,7 @@ import com.datastax.driver.core.VersionNumber;
 import org.apache.cassandra.repair.RepairParallelism;
 
 
-public interface ICassandraManagementProxy extends NotificationListener {
+public interface ICassandraManagementProxy {
 
   Duration DEFAULT_JMX_CONNECTION_TIMEOUT = Duration.ofSeconds(5);
 
@@ -102,8 +92,6 @@ public interface ICassandraManagementProxy extends NotificationListener {
    * @return list of tokens in the cluster
    */
   List<BigInteger> getTokens();
-
-  boolean isConnectionAlive();
 
   /**
    * @return true if any repairs are running on the node.
@@ -157,19 +145,6 @@ public interface ICassandraManagementProxy extends NotificationListener {
   void forceKeyspaceCompaction(boolean var1, String var2, String... var3) throws IOException, ExecutionException,
       InterruptedException;
 
-
-  // From MBeanServerConnection
-  Set<ObjectName> queryNames(ObjectName name, QueryExp query)
-      throws IOException;
-
-  MBeanInfo getMBeanInfo(ObjectName name)
-      throws InstanceNotFoundException, IntrospectionException,
-      ReflectionException, IOException;
-
-  AttributeList getAttributes(ObjectName name, String[] attributes)
-      throws InstanceNotFoundException, ReflectionException,
-      IOException;
-
   // From CompactionManagerMBean
 
   List<Map<String, String>> getCompactions();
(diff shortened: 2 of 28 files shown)


src/server/src/main/java/io/cassandrareaper/management/ICassandraManagementProxy.java
```diff
@@ -195,16 +170,6 @@ AttributeList getAttributes(ObjectName name, String[] attributes)
   // From StreamManagerMBean
   Set<CompositeData> getCurrentStreams();
 
-
-  void addConnectionNotificationListener(NotificationListener listener);
```
emerkle826, line 199: **Thought:**
I understand that some of these methods won't be used/implemented by the Http proxy implementation. The problem is that this interface is used everywhere that the JMX proxy was used. Removing these methods from the interface means that lots of places will now have to reference the Proxy object as a JMX type specifically. I know this PR is likely still a work in progress, but changes to this interface are going to have ripple effects in a lot of the other code that is currently using it.

adejanovski: I agree, can't we just make a noop implementation of these methods?
I'd rather keep the interface as is.

burmanm: Thing is, we can't use JMX to get metrics, compactionstats and that sort of things. And we can't return JMX types from those either. All that needs an implementation that requests the stuff from our /metrics endpoint, thus it needs to be in its own interfaces.

And so, any leakage of JMX to the HTTP endpoint indicates a place where we need to refactor the code, otherwise Reaper will not function correctly. 

There's a bit more work to this featureset than I'd say was originally estimated.

adejanovski: ok, I guess we indeed need separate interfaces for JMX and HTTP specific methods then, right? and the JMX/HTTP implementation classes will implement each their new interface in addition to te ICassandraManagementProxy?

> There's a bit more work to this featureset than I'd say was originally estimated.

Sure

burmanm: Well, I was thinking we would abstract some of those methods, such as MetricsProxy's and make it return those GenericMetrics instead of JMXStat and that sort of stuff.

It'll reduce the surface of the generic interfaces a lot if we stop returning JMX stuff everywhere and instead just make the conversion on the JMX implementation already.

Miles-Garnsey:
Hey @burmanm, I think this is a bit different to how I'd envisaged this working, which is my fault, as we had only discussed this verbally and I didn't have a proper ticket raised. 

I've raised a ticket to describe the way I thought this was going to work [here](https://github.com/thelastpickle/cassandra-reaper/issues/1346). We can change this if we decide that another path is better, but the way I thought this might work was to retain all of the existing `addNotification` type logic so that the HTTP and JMX both present as "JMX compliant" interfaces to Reaper. That was why I retained all of the JMX (inc. notifications) logic in the interface, so that Reaper effectively just kept doing what it always has.

Shimming HTTP to look like JMX would probably involve spinning up a new thread to poll management API whenever `addNotification` is called, and then hooking the events received from that thread into the `handle` method on RepairStatusHandler in the same way that the JMX implementation does it.

What you're doing here is a bit more ambitious, and seems like it would make Reaper agnostic as to whether it was using JMX or HTTP. The way I was envisioning this working was that we'd retain the existing JMX logic at higher levels of the application and just shim the new HTTP implementation so that it looks like JMX.

What you're proposing here is good because it helps reduce Reaper's JMX dependancy. However, I'm just concerned about the work that this might require given we're on a tight schedule. I'll leave it up to you and @adejanovski  to decide which is the better solution given timelines etc.

burmanm:
If we wish to retain the usage of addNotification, we should remove ICassandraManagementProxy and instead HTTPManagementProxy should extend JMXManagementProxy as the JMXManagementProxy is the only place that uses those methods - nothing else in Reaper uses them. SegmentRunner does not use them, repairStatusHandlers do not use them.

Creating a JMX emulator sounds like a lot of extra work for no benefit, just to make couple of extra methods that are not used by anything.

burmanm:
As for the notifications method done using this implementation, it's already done in PR #1339 .

Miles-Garnsey:
> If we wish to retain the usage of addNotification, we should remove ICassandraManagementProxy and instead HTTPManagementProxy should extend JMXManagementProxy as the JMXManagementProxy is the only place that uses those methods - nothing else in Reaper uses them. SegmentRunner does not use them, repairStatusHandlers do not use them.
> 
> Creating a JMX emulator sounds like a lot of extra work for no benefit, just to make couple of extra methods that are not used by anything.

`HTTPManagementProxy` doesn't extend `JMXManagementProxy` though, they're both implementations of the interface `ICassandraManagementProxy`. We do not want to carry any logic from `JMXManagementProxy` into `HTTPManagementProxy`.

My issue here is that the higher levels of the application are now using `instanceof` to break the abstraction of `ICassandraManagementProxy`. Higher levels of the application  should not know which implementation they are using, they should interact only with `ICassandraManagementProxy`.

If you can remove all `instanceOf` checks and then find you can also remove some methods from the interface because they are fully internal to `JMXManagementProxy`, then that's fine with me. Just remove them from  the interface and mark them private in `JMXManagementProxy`. But I think that these methods are implicitly required by other bits of logic in the handlers and you'll find this harder than you realise.

Also, note that we probably don't need to touch `DiagEventSubscriptionResource` for this exercise as we aren't going to be touching diagnostic events.

Miles-Garnsey:
@burmanm and I had a chat. It sounds like this PR is closer than I'd expected to what I was envisioning for this work. The main difference is that Michael thinks he can keep things like `*notificationListener` out of the interface (`ICassandraManagementProxy`). He'll then have an HTTP-native version of all of those things, which calls `SegmentRunner.handle()` to integrate with the existing logic.

To do that, he does need to do some `instanceOf` checks (which I don't like) to distinguish between JMX and HTTP implementation. But those checks only happen in the `DiagEventSubscriptionService` which we've discussed isn't going to ever be supported by the HTTP implementation.

My preference would actually be to remove all of the diagnostic events related stuff from both implementations and the management interface, but that seems like a refactoring task, not immediately related to this work.

There is another case where those checks are done to do with the `metricsProxy` but that is a separate conversation to a degree (and considering you call the same method in either event maybe that check is not necessary).

I'll look at the overall PR [here](https://github.com/thelastpickle/cassandra-reaper/pull/1339/files#diff-e93dfc38667ec30813a646f7f3ec602bd219e1a658ac93e495557b17aed5a941) in more detail tomorrow, but I'm happy to let this run for a bit right now, on the basis that any changes we do want to make might well end up being refactorings rather than functional changes.

src/server/src/main/java/io/cassandrareaper/ReaperApplication.java
```diff
@@ -232,13 +232,16 @@ public void run(ReaperApplicationConfiguration config, Environment environment)
...
     HttpClient httpClient = createHttpClient(config, environment);
-    ScheduledExecutorService ses = environment.lifecycle().scheduledExecutorService("Diagnostics").threads(6).build();
-    final DiagEventSubscriptionResource eventsResource = new DiagEventSubscriptionResource(context, httpClient, ses,
-        context.storage.getEventsDao());
-    environment.jersey().register(eventsResource);
-    final DiagEventSseResource diagEvents = new DiagEventSseResource(context, httpClient, ses,
-        context.storage.getEventsDao());
-    environment.jersey().register(diagEvents);
+
+    if (context.managementConnectionFactory instanceof JmxManagementConnectionFactory) {
```
emerkle826: *Suggestion:*
Instead of using `instanceof` when we need to know if it's a JMX-capable implementation, can we add a `isJmxCompatible()` method to the `ICassandraManagementProxy` interface the returns `false` by default and only the `JmxCassandraManagementProxy`class overrides it to return true? I think it's a slightly cleaner approach.

src/server/src/main/java/io/cassandrareaper/management/MetricsProxy.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018-2019 The Last Pickle Ltd
+ * Copyright 2020-2020 The Last Pickle Ltd
```
emerkle826: *Question:*
Do we need to change this stuff in the headers?

emerkle826 (approved):
This looks good to me. I agree with Miles' comments about possibly refactoring some of this down the road. For now this is good. Just the one suggestion to try to reduce the `instanceof` checks, but it's not required to change at this point for my approval.

src/server/src/main/java/io/cassandrareaper/management/MetricsProxy.java
burmanm: I think these renames of files caused it to have a weird copyright information.

src/server/src/main/java/io/cassandrareaper/ReaperApplication.java
burmanm: I'm not entirely sure where the hate for instanceof comes from, but from the same file, we can of course use this method instead of instanceof:

```java
if (context.config.getHttpManagement() == null || !context.config.getHttpManagement().getEnabled()) {
```

Because that's what the initialization of HTTP or JMX endpoint uses. Not sure how that's different to instanceof check, both are single lines and have the exact same semantics in this case. Especially since it's done in the same file. Actually, I'd even say that instance of makes more sense since it's a single check and not dual one (and we don't repeat the same procedure everywhere then). 

So why pollute interface with the info, when the other parts don't use such either?

emerkle826: That's fine. I think what I would like to see is a more involved refactor later on. The hate for `instanceof` isn't hate on it so much (it's in the JDK), it's just that it indicates the interface design is sub-optimal here. Ideally we never have to use it. But to do so would require redesigning a lot of Reaper.

I'm fine with the code PR as it is.

emerkle826 (approved):

src/server/src/main/java/io/cassandrareaper/ReaperApplication.java
Miles-Garnsey: This conversation was EXACTLY my thought process too :) I also prefer using the context.config since config is globally available and immutable, and doesn't break the design of the types. But in the end the result is the same, so the bigger refactor is probably the most important thing in the end.

burmanm: I'm infinitely interested in how this upcoming refactor removes the initializeManagement() method and the new instance creation of JMX/HTTPManagementProxy. 

https://github.com/thelastpickle/cassandra-reaper/blob/master/src/server/src/main/java/io/cassandrareaper/ReaperApplication.java#L304

Because I'm super puzzled how we create that instance in that case. And what on earth is the benefit of avoiding all this.

Miles-Garnsey: Let's talk about the upcoming refactor when we do it, and stay focused on getting the current milestone's work out for now.

src/server/src/main/java/io/cassandrareaper/management/MetricsProxy.java
Miles-Garnsey: @burmanm can you fix this prior to us approving? Cheers!

burmanm: That's a weird req (it passed tests), but changed now.

PR #1339, opened 2023-08-10
Mimic storage notifications using the HTTP API

### LINKED ISSUE #1325: Ensure Reaper can call the new Management API async notifications endpoints
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=34006288)

We are attempting to provide an alternative to JMX using the HTTP API in the management API project. 

As a part of this work, we need to ensure that notification based logic has non-JMX equivalents. The new logic should be implemented in the [HttpCassandraManagementProxy](https://github.com/thelastpickle/cassandra-reaper/blob/84a742ae50fac1751556b0ceebb1413f65308e58/src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java#L60).

We want to ignore diagnostic events as a part of this work. As a result, the only functions we want to touch are related to the `NotificationListener` interface, which has a single method - 

```
  public void handleNotification(Notification notification, Object handback);
```

When the `connect` method is called in the JmxCassandraManagementProxy, it calls `mbeanServerConn.addNotificationListener(ObjectNames.STORAGE_SERVICE, proxy, null, null);` which sets up a listener for events from the storage service.

When those events are received, they are processed by handleNotification` [here](https://github.com/thelastpickle/cassandra-reaper/blob/56394b3eddc59cebb6b14d945032a1926d8f44f2/src/server/src/main/java/io/cassandrareaper/management/jmx/JmxCassandraManagementProxy.java#L778), which filters them down to just those of type "progress". These are progress notifications for repairs.

The only notifications we need to provide via management API are those from the storage service with `type == "repair"`. Others relate to diagnostic events. The listeners for the storage service are added [here](https://github.com/thelastpickle/cassandra-reaper/blob/56394b3eddc59cebb6b14d945032a1926d8f44f2/src/server/src/main/java/io/cassandrareaper/management/jmx/JmxCassandraManagementProxy.java#L284-L288) in the JMX implementation. They are handled [here](https://github.com/thelastpickle/cassandra-reaper/blob/56394b3eddc59cebb6b14d945032a1926d8f44f2/src/server/src/main/java/io/cassandrareaper/management/jmx/JmxCassandraManagementProxy.java#L778) in the JMX implementation, so the HTTP impl will need to basically shim things out to that point so that arguments to `SegmentRunner.handle()` ([here](https://github.com/thelastpickle/cassandra-reaper/blob/56394b3eddc59cebb6b14d945032a1926d8f44f2/src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java#L469)) appear to `SegmentRunner.handle()` as though they come from JMX, not HTTP.

**Definition of done**

- [ ] We have an HTTP equivalent for `addNotificationListener`, which needs to set up a new thread to poll the repairs endpoint on management API.
- [ ] We have a new `handleNotification` method in the HttpCassandraManagementProxy, which needs to eventually trigger a callback to the `SegmentRunner`s as implementors of the `repairStatusHandler` interface. (see e.g. `processNewApiNotification` in the JMX implementation.) 
- [ ] The relevant HTTP endpoints are called using the new Management API Java client (which should be auto-generated by the management api release process).
- [ ] When the state of a watched object changes, the state change is added to the history of the object. This ensures that all state changes (even for states which last for only ms in between polls) are propagated down the wire.

**Note**

Because we are ignoring the diagnostic events here, we do not need to look at any other instances of `addNotificationListener` for other mbean/event types.

burmanm:
This branch is based on the PR #1336


src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
```diff
@@ -33,29 +35,50 @@
...
 import javax.management.openmbean.TabularData;
 import javax.validation.constraints.NotNull;
 
+import com.codahale.metrics.InstrumentedScheduledExecutorService;
 import com.codahale.metrics.MetricRegistry;
+import com.google.common.collect.Maps;
 import org.apache.cassandra.repair.RepairParallelism;
 
 public class HttpCassandraManagementProxy implements ICassandraManagementProxy {
+  public static final int DEFAULT_POLL_INTERVAL_IN_MILLISECONDS = 5000;
```
Miles-Garnsey, line 69: Suggestion: There are a lot of notifications specific fields here, can we possibly delegate this functionality out into a new class instead of having it inside HttpCassandraManagementProxy?

src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
```diff
@@ -33,29 +35,50 @@
...
-                                      InetSocketAddress endpoint
+                                      InetSocketAddress endpoint,
+                                      ScheduledExecutorService executor
   ) {
     this.metricRegistry = metricRegistry;
     this.rootPath = rootPath;
     this.endpoint = endpoint;
+    this.statusTracker = new InstrumentedScheduledExecutorService(executor, metricRegistry);
+
+    // TODO Perhaps the poll interval should be configurable through context.config ?
```
Miles-Garnsey, line 97: Suggestion: suggest we leave this for a later PR. We're under time pressure right now, so unless/until a user needs it let's keep things as bare bones as possible.

src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
```diff
@@ -149,13 +172,22 @@ public int triggerRepair(
       List<RingRange> associatedTokens,
       int repairThreadCount)
       throws ReaperException {
-    return 1; //TODO: implement me
 
+    String jobId = "repair-1"; // TODO Make the actual implementation to call the mgmt-api to repair
```
Miles-Garnsey: Question: I don't quite understand this comment, do we need to get the jobID from calling the GET /repairs endpoint and enumerating all repairs? 

In the JMX implementation, do we need the repairID in order to subscribe to the notifications for a given repair instance?

src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
```diff
@@ -149,13 +172,22 @@ public int triggerRepair(
...
+    int repairNo = Integer.parseInt(jobId.substring(7));
+
+    repairStatusExecutors.putIfAbsent(repairNo, Executors.newSingleThreadExecutor());
```
Miles-Garnsey, line 230: Question: So if I'm reading this correctly, what is happening is that for each RepairID we are adding a new executor into the map, and each of those will continually and asynchronously poll the mgmt API HTTP endpoint for status updates?

src/server/src/main/java/io/cassandrareaper/management/http/models/Job.java
```diff
@@ -0,0 +1,128 @@
...
+
+package io.cassandrareaper.management.http.models;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import com.google.common.annotations.VisibleForTesting;
+import org.apache.cassandra.utils.progress.ProgressEventType;
+
+public class Job {
```
Miles-Garnsey: Question: I'm a little unclear on the purpose of this class. `SegmentRunner.handle` takes arguments as follows:

```
public void handle(
      int repairNo,
      Optional<ActiveRepairService.Status> status,
      Optional<ProgressEventType> progress,
      String message,
      ICassandraManagementProxy cassandraManagementProxy) 
```

If we're just passing the job details directly to `SegmentRunner.handle`, why do we need an additional class `Job` to intermediate here?

src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
```diff
@@ -33,29 +35,50 @@
...
+                                      InetSocketAddress endpoint,
+                                      ScheduledExecutorService executor
   ) {
     this.metricRegistry = metricRegistry;
     this.rootPath = rootPath;
     this.endpoint = endpoint;
+    this.statusTracker = new InstrumentedScheduledExecutorService(executor, metricRegistry);
+
+    // TODO Perhaps the poll interval should be configurable through context.config ?
+    this.scheduleJobPoller(statusTracker, DEFAULT_POLL_INTERVAL_IN_MILLISECONDS);
```
Miles-Garnsey: Nit: no need for `this` I don't think.

src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
```diff
@@ -228,5 +260,39 @@ public String getUntranslatedHost() {
     return "";
   }
 
+  private Job getJobStatus(String id) {
+    // Poll with HTTP client the job's status
+    return new Job("repair", "repair-12345678"); // Stub, but this is the format that mgmt-api returns
+  }
+
+  private void scheduleJobPoller(ScheduledExecutorService scheduler, int pollInterval) {
```
Miles-Garnsey: Question: I'm confused by how the concurrency is meant to work here. Intuitively, I'd have thought that scheduleJobPoller should be called for each repairID and should then add an executor to `repairStatusExecutors` to commence polling the endpoint . But it seems that instead `triggerRepair` is called for each repairID, and adds to both `repairStatusExecutors` and `repairStatusHandlers`, but not to `jobTracker`.

So a few questions about this:

1. Why do we need a separate `repairStatusHandler` for each repair? Is that the way the current JMX implementation works? It looks like maybe, given that new `SegmentRunners` (the implementor of `RepairStatusHandler` are created for every call of `repairSegment` and presumably this is called per segment).
2. Maybe if you could let me know what each of the below maps is meant to hold this would help me understand where you're going with this:

```
private final ConcurrentMap<Integer, RepairStatusHandler> repairStatusHandlers = Maps.newConcurrentMap();
  private final ConcurrentMap<String, JobStatusTracker> jobTracker = Maps.newConcurrentMap();
  private final ConcurrentMap<Integer, ExecutorService> repairStatusExecutors = Maps.newConcurrentMap();
```

src/server/src/main/java/io/cassandrareaper/management/http/models/Job.java
Miles-Garnsey: NB: if this is because we want to have a model for the data returned in json form from the mgmt API side, please note that all models should be imported from the mgmt API client, not declared in Reaper.

Miles-Garnsey (changes requested):
Thanks for this @burmanm, it is good to see how your proposed notifications mechanism might work! 

One Issue: can we possibly start putting some unit tests into this PR? I know higher level tests won't pass because we don't yet have the work done on management API, but it would be good to see unit tests if possible.

I've left a number of other comments (mostly questions) throughout the code. I'll also continue reading through to try to better understand how this will work. 

Perhaps it will be clearer to me once some of the mgmt API stuff is complete and we can start to see the integration between the two more clearly.

src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
burmanm: In the JMX implementation, the repairID is used to match the repair to the notifications, yes. Since I didn't want to rewrite that implementation, I simply mimicked the same behavior here. 

The jobId comes as the return payload when we submit the repair call, yes (just like all the other mgmt-api async calls do)

src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
burmanm: Similar to JMX implementation, yes.

src/server/src/main/java/io/cassandrareaper/management/http/models/Job.java
burmanm: It's redeclared here (and probably commented also) since there's no mgmt-api client in the Reaper, thus there's nothing to use. Once that's in, then this class disappears.

src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
burmanm: > Why do we need a separate repairStatusHandler for each repair? Is that the way the current JMX implementation works?

Yes, this is how the JMX implementation works.

> private final ConcurrentMap<Integer, RepairStatusHandler> repairStatusHandlers = Maps.newConcurrentMap();
 > private final ConcurrentMap<Integer, ExecutorService> repairStatusExecutors = Maps.newConcurrentMap();

These are both the same as in JMX implementation, since the implementation itself is pretty much the same as the JMX one. No logic changes. 

> private final ConcurrentMap<String, JobStatusTracker> jobTracker = Maps.newConcurrentMap();

This is used to track the last seen notification from the mgmt-api side. So that we know which notifications are to be sent to the repairStatusHandlers and which ones have been processed already.

src/server/src/main/java/io/cassandrareaper/management/http/models/Job.java
Miles-Garnsey: @burmanm the management API client is now available via this PR https://github.com/k8ssandra/management-api-for-apache-cassandra/pull/341

We should be able to eliminate the Job class and begin integration between mgmt API and  now I think. Let me know if you agree?

src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
Miles-Garnsey: @burmanm, what do you think of this suggestion?

burmanm: I think they're very tightly integrated to this functionality (there's no usage outside this class), so I wouldn't. And there's only 3 in the end, the polling interval (which we could move to config in any case), the executor and one more map. It follows the same idea as JMX used to in its functionality.

src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
burmanm: Yep.

src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020-2020 The Last Pickle Ltd
+ * Copyright 2023-2023 DataStax, Inc.
```
Miles-Garnsey, line 2: Hmmm, I think this should be 

```
 * Copyright 2020-2022 The Last Pickle Ltd
 * Copyright 2023 DataStax, Inc.
 ```

But I'm not a lawyer.

File: src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java [hunk 9]
```diff
@@ -192,13 +217,31 @@ public int triggerRepair(
       List<RingRange> associatedTokens,
       int repairThreadCount)
       throws ReaperException {
-    return 1; //TODO: implement me
 
+    String jobId;
```
Miles-Garnsey, line 221: Question: should we merge this now or wait for me to finish [this](https://github.com/k8ssandra/management-api-for-apache-cassandra/pull/358) PR so that we can trigger via the new endpoint?

src/server/src/test/java/io/cassandrareaper/management/http/HttpCassandraManagementProxyTest.java
```diff
@@ -83,4 +109,98 @@ private <T extends Object> T jsonFromResourceFile(String filename, Class<T> claz
...
+        Mockito.mock(InetSocketAddress.class), executorService, mockClient);
+    when(connectionFactory.connectAny(any())).thenReturn(httpCassandraManagementProxy);
+
+    RepairStatusHandler repairStatusHandler = Mockito.mock(RepairStatusHandler.class);
+
+    int repairNo = httpCassandraManagementProxy.triggerRepair(BigInteger.ZERO, BigInteger.ONE, "ks",
+        RepairParallelism.PARALLEL,
+        Collections.singleton("table"), true, Collections.emptyList(), repairStatusHandler, Collections.emptyList(), 1);
+
+    assertEquals(123456789, repairNo);
```
Miles-Garnsey, line 134: Issue: can we make some assertions around which functions have been called and how many times?

src/server/src/test/java/io/cassandrareaper/management/http/HttpCassandraManagementProxyTest.java
```diff
@@ -83,4 +109,98 @@ private <T extends Object> T jsonFromResourceFile(String filename, Class<T> claz
...
+    doReturn(ConcurrentUtils.constantFuture(null)).when(executorService).submit(any(Callable.class));
+
+    HttpCassandraManagementProxy httpCassandraManagementProxy = new HttpCassandraManagementProxy(
+        Mockito.mock(MetricRegistry.class), "",
+        Mockito.mock(InetSocketAddress.class), executorService, mockClient);
+    when(connectionFactory.connectAny(any())).thenReturn(httpCassandraManagementProxy);
+
+    RepairStatusHandler repairStatusHandler = Mockito.mock(RepairStatusHandler.class);
+
+    int repairNo = httpCassandraManagementProxy.triggerRepair(BigInteger.ZERO, BigInteger.ONE, "ks",
```
Miles-Garnsey, line 130: Question: can we figure out whether the beginToken/endToken is still the right way to call this function? From what I could see it seemed like the current underlying JMX call actually uses the RingRange, but I'd have to go back to confirm how this works.

src/server/src/test/java/io/cassandrareaper/management/http/HttpCassandraManagementProxyTest.java
Miles-Garnsey: Issue: can we please test the individual methods here? And also look at what happens under failure conditions?

Miles-Garnsey (changes requested):
Here's a preliminary review, I'm still working through this as I've been trying to get my PR merged today. I think maybe some more tests might be good, but not sure how hard they might be to implement given all of the concurrency stuff you have here.

src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
burmanm: Checkstyle won't accept that format. In any case, this type of copyright info is not necessary in EU at least.

src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
burmanm: I'd merge this first so you can then create proper tests with the new repair endpoints. Otherwise we'll end up without tests.

src/server/src/test/java/io/cassandrareaper/management/http/HttpCassandraManagementProxyTest.java
burmanm: Isn't that part of your PR? In this test those values mean nothing, they're simply ignored.

src/server/src/test/java/io/cassandrareaper/management/http/HttpCassandraManagementProxyTest.java
burmanm: > Issue: can we make some assertions around which functions have been called and how many times?

Done in the next test. It's not what this test would test as it doesn't care if any method is called. This is intended to test that the maps have correct amount of entries (if someone refactors the repair endpoint for example, but forgets to insert correct places, then this test would fail and tell that info).

> Issue: can we please test the individual methods here? And also look at what happens under failure conditions?

This is already calling the individual methods. triggerRepair & removeRepairStatusHandlers are the interesting methods this one calls directly.

src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
Miles-Garnsey: Same in AU, but we need to consider US.

src/server/src/test/java/io/cassandrareaper/management/http/HttpCassandraManagementProxyTest.java
Miles-Garnsey: I think it only applies to Cassandra < 2.2 from memory, that's why I'm wondering if it is a good idea to merge this now, since additional work might yield a slightly different solution for parts of this PR.

burmanm: I mean this PR doesn't care even if it's triggerRepair() and no parameters at all. That has no impact to functionality here.

src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
burmanm: Well, then that would be outside PR which would modify the checkstyle and copyright creation in the project. Right now the years are all over the place in every file.

src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
```diff
@@ -348,4 +391,54 @@ public String getUntranslatedHost() {
...
+    // Poll with HTTP client the job's status
+    try {
+      return apiClient.getJobStatus(id);
+    } catch (ApiException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  @VisibleForTesting
+  private void scheduleJobPoller(ScheduledExecutorService scheduler, int pollInterval) {
```
Miles-Garnsey: Let's remove this ScheduledExecutorService argument and call this.scheduler?

src/server/src/test/java/io/cassandrareaper/management/http/HttpCassandraManagementProxyTest.java
```diff
@@ -83,4 +109,98 @@ private <T extends Object> T jsonFromResourceFile(String filename, Class<T> claz
     return new ObjectMapper().readValue(
         this.getClass().getResource(filename).openStream(), clazz);
   }
+
+  @Test
+  public void testRepairHandlers() throws Exception {
```
Miles-Garnsey: Maybe consider more descriptive names for tests?

src/server/src/test/java/io/cassandrareaper/management/http/HttpCassandraManagementProxyTest.java
```diff
@@ -83,4 +109,98 @@ private <T extends Object> T jsonFromResourceFile(String filename, Class<T> claz
...
+    HttpCassandraManagementProxy httpCassandraManagementProxy = new HttpCassandraManagementProxy(
+        mock(MetricRegistry.class), "",
+        mock(InetSocketAddress.class), executorService, mockClient);
+    when(connectionFactory.connectAny(any())).thenReturn(httpCassandraManagementProxy);
+
+    final AtomicInteger callTimes = new AtomicInteger(0);
+    RepairStatusHandler workAroundHandler = (repairNumber, status, progress, message, cassandraManagementProxy)
+        -> callTimes.incrementAndGet();
+
+    //    RepairStatusHandler repairStatusHandler = mock(RepairStatusHandler.class);
```
Miles-Garnsey: Please remove comment.

Miles-Garnsey (approved):
Three comments to address, otherwise good to merge to the integration branch!

PR #1349, opened 2023-08-11
Fix broken CI due to AdoptOpenJDK removal from GHA runners

### LINKED ISSUE #1348: CI is broken due to adopt jdk removal from GHA
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=35643097)

AdoptOpenJDK is being removed from GHA runners which breaks our CI.
We need to update the jmx.remote file in the JDK directory to setup JMX authentication.
This needs to be done in the new JDK directory.

## Definition of Done
- [x] CI is passing again with the new JDK

adejanovski:
Fixes #1348

### DIFF
--- a/.github/scripts/configure-ccm.sh
+++ b/.github/scripts/configure-ccm.sh
@@ -70,6 +70,7 @@ case "${TEST_TYPE}" in
         mkdir -p ~/.local
         cp ./.github/files/jmxremote.password ~/.local/jmxremote.password
         chmod 400 ~/.local/jmxremote.password
+        ls -lrt /opt/hostedtoolcache/
         # Our move to support jdk11 makes it possible to get either jdk 8.0.192 or jdk 8.0.292 which both require to be configured properly.
         for jdk in /opt/hostedtoolcache/Java_Adopt_jdk/8*/; 
         do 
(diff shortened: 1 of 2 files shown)


.github/scripts/configure-ccm.sh
```diff
@@ -25,10 +25,11 @@ function set_java_home() {
     else
         export CASSANDRA_USE_JDK11=false
     fi
-    for jdk in /opt/hostedtoolcache/Java_Adopt_jdk/${major_version}*/; 
+    for jdk in /opt/hostedtoolcache/Java_Temurin-Hotspot_jdk/${major_version}*/; 
     do 
         export JAVA_HOME="${jdk/}"x64/
         echo "JAVA_HOME is set to $JAVA_HOME"
+        export JAVA_TOOL_OPTIONS="-Dcom.sun.jndi.rmiURLParsing=legacy"
```
burmanm, line 32: question: Why was this suddenly needed? This shouldn't change the version of Java otherwise.

burmanm (approved):

.github/scripts/configure-ccm.sh
adejanovski: I think we were running on older versions of the jdk which didn't exhibit this issue.

PR #1359, opened 2023-08-17
Implement Http proxy methods

emerkle826:
- getClusterName
- getLiveNodes
- clearSnapshot
- listSnapshots
- takeSnapshot

This PR gets at least the first 2 IT tests passing and using a client from Management API branch/PR here:
branch: https://github.com/k8ssandra/management-api-for-apache-cassandra/tree/client-gen-snapshots
PR: https://github.com/k8ssandra/management-api-for-apache-cassandra/pull/348

All of the methods in the Proxy implemented here were done so to get the first test to pass.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
+++ b/src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
@@ -29,6 +29,7 @@ import java.math.BigInteger;
 import java.net.InetSocketAddress;
 import java.net.UnknownHostException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -44,27 +45,32 @@ import com.codahale.metrics.MetricRegistry;
 import com.datastax.mgmtapi.client.api.DefaultApi;
 import com.datastax.mgmtapi.client.invoker.ApiClient;
 import com.datastax.mgmtapi.client.invoker.ApiException;
+import com.datastax.mgmtapi.client.model.EndpointStates;
+import com.datastax.mgmtapi.client.model.SnapshotDetails;
+import com.datastax.mgmtapi.client.model.TakeSnapshotRequest;
+import com.google.common.annotations.VisibleForTesting;
 import org.apache.cassandra.repair.RepairParallelism;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class HttpCassandraManagementProxy implements ICassandraManagementProxy {
   private static final Logger LOG = LoggerFactory.getLogger(HttpCassandraManagementProxy.class);
-  String host;
-  MetricRegistry metricRegistry;
-  String rootPath;
-  InetSocketAddress endpoint;
-  DefaultApi apiClient;
+  final String host;
+  final MetricRegistry metricRegistry;
+  final String rootPath;
+  final InetSocketAddress endpoint;
+  final DefaultApi apiClient;
 
   public HttpCassandraManagementProxy(MetricRegistry metricRegistry,
                                       String rootPath,
                                       InetSocketAddress endpoint
   ) {
+    this.host = endpoint.getHostString();
     this.metricRegistry = metricRegistry;
     this.rootPath = rootPath;
     this.endpoint = endpoint;
     this.apiClient = new DefaultApi(
-        new ApiClient().setBasePath("https://" + endpoint.getHostName() + ":" + endpoint.getPort() + rootPath));
+        new ApiClient().setBasePath("http://" + endpoint.getHostName() + ":" + endpoint.getPort() + rootPath));
   }
 
   @Override
@@ -101,7 +107,29 @@ public class HttpCassandraManagementProxy implements ICassandraManagementProxy {
 
   @Override
   public String getClusterName() {
-    return null; // TODO: implement me.
+    // Cluster name is part of Endpoint States.
+    try {
+      EndpointStates endpoints = apiClient.getEndpointStates();
+      if (endpoints == null) {
+        LOG.error("No endpoint state data retrieved.");
+        return null;
+      }
+      List<Map<String, String>> states = endpoints.getEntity();
+      if (states == null || states.isEmpty()) {
+        LOG.error("Endpoint state data was null or empty:\n" + endpoints.toJson());
+        return null;
+      }
+      // cluster name should be the same for all states
+      String clusterName = states.get(0).get("CLUSTER_NAME");
+      if (clusterName == null) {
+        LOG.error("Endpoint state data did not contain a CLuster Name:\n" + endpoints.toString());
+        // no need to return null here, we'll do it on the next line
+      }
+      return clusterName;
+    } catch (ApiException ae) {
+      LOG.error("Failed to retrieve Cluster Name from Endpoint states data", ae);
+    }
+    return null;
   }
 
   @Override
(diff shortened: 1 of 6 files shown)


src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
```diff
@@ -101,7 +106,19 @@ public String getPartitioner() {
...
+      assert endpoints != null;
+      List<Map<String, String>> states = endpoints.getEntity();
+      assert states != null;
+      assert !states.isEmpty();
+      // cluster name should be the same for all states
+      return states.get(0).get("CLUSTER_NAME");
+    } catch (ApiException ae) {
+      LOG.error("Failed to retrieve Cluster Name from Endpoint states data", ae);
+    }
+    return null;
```
olim7t, line 132: thought: I wonder if we shouldn't throw an exception here (it would have to be an unchecked one since the method doesn't declare `ReaperException`).
Looking at the method usages, there is at least one case where null is not supported, in `Cluster.Builder.build()`.

olim7t (approved):

src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
emerkle826: Good point. When I first coded this, I wasn't sure what we could/should do if we can't get the cluster name. Stuff will break for sure.

emerkle826: Question: @Miles-Garnsey do you think we could/should define the parent interface such that it declares a throw of  ReaperException here?

Miles-Garnsey: I don't think so, logging is probably sufficient. If we throw an exception all we'd be doing is logging it and retrying I think, since there's always a chance that the next response will be right. Do we know if there is a retry somewhere? Or will we ultimately end up with an NPE in the Cluster.Builder.build() call?

However, you may want to include the endpoint states data that was bad in the log.

emerkle826: Yeah, I think it's just going to be an NPE down the path. I'll add what I can to the exception log.

Miles-Garnsey: I don't know what else we can do to be honest. I always think there's not much point throwing an exception if the caller isn't able to handle it anyway, and I'm not sure (beyond a retry) what the caller can do in this case.

emerkle826: So I can't really add much to the catch of the `ApiException`. It will only be thrown if the call to getEndpoints fails. I can log things along the way if that call succeeds, but we still are missing things we expect in the response.

emerkle826: @Miles-Garnsey I think I have covered the cases here.....

Miles-Garnsey: I think that's fine! A bit tough to review this one as I think it combines a number of commits from other PRs. Can we get those merged, rebase and then do a final review?

Miles-Garnsey (changes requested):
Issue: This is looking good, the one thing it will need is a unit test since we do have some substantial logic in the function `listSnapshots()`.

emerkle826:
> Issue: This is looking good, the one thing it will need is a unit test since we do have some substantial logic in the function `listSnapshots()`.

Good point!

src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
```diff
@@ -180,19 +208,86 @@ public void close() {
...
+      return liveNodes;
+    } catch (ApiException ae) {
+      LOG.error("Failed to fetch live nodes", ae);
+      throw new ReaperException(ae);
+    }
   }
 
-  public void clearSnapshot(String var1, String... var2) throws IOException {
-    // TODO: implement me.
+  public void clearSnapshot(String snapshotName, String... keyspaces) throws IOException {
```
Miles-Garnsey, line 227: Issue: Should have `@Override` I think?

src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
```diff
@@ -180,19 +208,86 @@ public void close() {
...
+    try {
+      apiClient.clearSnapshots(Arrays.asList(snapshotName), Arrays.asList(keyspaces));
+    } catch (ApiException ae) {
+      LOG.error("Failed to clear snapshots", ae);
+      throw new IOException(ae);
+    }
   }
 
   public List<Snapshot> listSnapshots() throws UnsupportedOperationException {
```
Miles-Garnsey, line 237: Issue: Should have `@Override` I think?

src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
```diff
@@ -180,19 +208,86 @@ public void close() {
...
+    } catch (ApiException ae) {
+      LOG.error("Failed to retrieve Snapshot details", ae);
+      throw new UnsupportedOperationException(ae);
+    }
   }
 
-  public void takeSnapshot(String var1, String... var2) throws IOException {
-    // TODO: implement me.
+  @VisibleForTesting
+  List<Snapshot> convert(SnapshotDetails snapshotDetails) {
```
Miles-Garnsey: Issue: `convertSnapshot()` would be more appropriate here.

src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
```diff
@@ -180,19 +208,86 @@ public void close() {
...
+          default:
+            break;
+        }
+      }
+      snapshots.add(snapshotBuilder.withClusterName(getClusterName()).build());
+    }
+    return snapshots;
+  }
+
+  public void takeSnapshot(String snapshotName, String... keyspaceNames) throws IOException {
```
Miles-Garnsey, line 284: Issue: `@Override` needed.

Miles-Garnsey (changes requested):
The test looks good now. I have a few issues on the fact that we haven't marked override methods with `@Override`, and also just with the naming of the conversion function which is a bit too vague.

I also should manually test this prior to approval (or get the debugger out to see what data is actually being exchanged between Reaper and Management API), but I'm  having issues getting the CCM setup script to work.

src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
emerkle826: I'm going with `convertSnapshots()` as it is likely more than one in the SnapshotDetails object that it will convert into a List.

emerkle826:
> I have a few issues on the fact that we haven't marked override methods with `@Override`

I've put the `@Override` annotation on all of the methods now, even the unimplemented ones.

emerkle826:
> but I'm having issues getting the CCM setup script to work.

I'll try to put together a script that will make this easier.

Miles-Garnsey (approved):
All feedback addressed, approving for merge into the integration branch.

Issue #1361, opened 2023-08-18
Add new suite of acceptance tests to exercise HTTP management proxy


Miles-Garnsey:
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=36209534)

The HttpCassandraManagementProxy has a number of stub methods currently, these are hard to test outside an e2e context because they rely on having management API and Cassandra available. 

It would be nice to ensure that these methods are all exercised as a part of the acceptance tests in Reaper.

Miles-Garnsey:
We elected not to proceed on this one as it was more effort than we wanted to expend, and the functionality is covered by existing tests.

PR #1363, opened 2023-08-18
HTTP Methods Tests

Miles-Garnsey:
To ensure that our HTTP methods are all working, we are introducing acceptance tests for the individual methods on HTTPCassandraManagementProxy alongside the existing e2e tests.

**Fixes issue:**
#1361


Miles-Garnsey:
We determined that this piece of work was going to be too costly to implement at this time, and would be covered by existing tests already.

Issue #1343, opened 2023-08-10
Implement repair related methods in the HttpManagementProxy


adejanovski:
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=35567395)

## Definition of Done
- [x] the `cancelAllRepairs()` method is implemented in the HttpCassandraManagementProxy class
- [x] the `triggerRepair()` method is implemented in the HttpCassandraManagementProxy class

Miles-Garnsey:
Merged to main.

PR #1370, opened 2023-08-29
Repair http methods

Miles-Garnsey:
Moves to the new HTTP client for triggering/cancelling repairs.

**Fixes issue**
#1343


Miles-Garnsey:
There is a problem in this PR because the Management API client is generating the `RingRange` type with `Integer` types, not `BigInteger`. I think this will result in a problem where the type cannot represent the tokenRange appropriately. 

Once we resolve this (we will need to look at the OpenAPI generation process) we can move forward with this PR, as it is otherwise ready.

emerkle826 (approved):

PR #1369, opened 2023-08-28
Implement compaction methods in the HttpManagementProxy (fixes #1340)

olim7t:
TODO:
- [x] merge k8ssandra/management-api-for-apache-cassandra#365 and generate new API client

PR #1371, opened 2023-08-31
Miles' compaction methods

Miles-Garnsey:
Updates the pom.xml to contain the new compaction methods. Built from Olivier's branch #1369 to confirm it works before merging.


Miles-Garnsey:
Closing in favour of #1369. We will merge that PR once tests pass.

PR #1376, opened 2023-09-05
Implement HTTP based getToken, getEndpointToHostId, getLocalEndpoint

### LINKED ISSUE #1377: Implement getTokens() in the HttpManagementProxy
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=37693416)

Method signature:

`public List<BigInteger> getTokens()`

### LINKED ISSUE #1379: Implement getLocalEndpoint() in the HttpManagementProxy
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=37693658)

Method signature:
`public String getLocalEndpoint() throws ReaperException`

### LINKED ISSUE #1380: Implement getEndpointToHostId() in the HttpManagementProxy
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=37693667)

Method signature: 
` public Map<String, String> getEndpointToHostId()`

Miles-Garnsey:
Fixes #1377
Fixes #1379
Fixes #1380

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
+++ b/src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
@@ -27,6 +27,7 @@ import io.cassandrareaper.service.RingRange;
 
 import java.io.IOException;
 import java.math.BigInteger;
+import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.UnknownHostException;
 import java.util.ArrayList;
@@ -110,18 +111,18 @@ public class HttpCassandraManagementProxy implements ICassandraManagementProxy {
   public List<BigInteger> getTokens() {
     EndpointStates endpointStates;
     try {
-      endpointStates =  apiClient.getEndpointStates();
+      endpointStates = apiClient.getEndpointStates();
       return Arrays.stream(endpointStates.getEntity().stream().filter(i ->
-              i.getOrDefault("IS_LOCAL", "false")
-              .equals("true")
-          )
-          .findFirst()
-          .orElseThrow(() -> new RuntimeException("Failed to find local endpoint"))
-          .get("TOKENS")
-          .split(","))
+                  i.getOrDefault("IS_LOCAL", "false")
+                      .equals("true")
+              )
+              .findFirst()
+              .orElseThrow(() -> new RuntimeException("Failed to find local endpoint"))
+              .get("TOKENS")
+              .split(","))
           .map(strToken -> new BigInteger(strToken)).collect(Collectors.toList());
 
-    } catch (Exception e) {
+    } catch (ApiException e) {
       LOG.error("Failed to retrieve endpoint states", e);
       return Collections.emptyList();
     }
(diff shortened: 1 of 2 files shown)


src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
```diff
@@ -119,13 +138,38 @@ public Map<List<String>, List<String>> getRangeToEndpointMap(String keyspace) th
   @NotNull
   @Override
   public String getLocalEndpoint() throws ReaperException {
-    return null; // TODO: implement me.
+    // TODO: validate that this works in all situations. I suspect that if any address translation is
+    // happening we'll see failures here, but address translation is not in scope in this phase.
+    // The logic is that host is either a DNS address, or an IP address. If it's a DNS address, we do a
+    // reverse lookup to get the IP.
+    String regex = "^(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\." +
```
burmanm: issue/question: I feel like this type of regexp is terrible for maintenance, as reading and modifying a mistake in it could take hours for someone else. Is this really needed? If the entire idea is to get just the hostnames, doesn't InetAddress provide this already?

``public static InetAddress getByName(String host)`` javadoc (and implementation) says:

```
Determines the IP address of a host, given the host's name.
The host name can either be a machine name, such as "java.sun.com", or a textual representation of its IP address.
```

That is, if we pass IP or a hostname to InetAddress.getByName(host) and so on (so basically same code path always), it should just work. No need for regexp.

Miles-Garnsey: If you read the comment you'll see what I'm trying to do. The `host` string ultimately could be either a DNS name or an IP address. My understanding though is that an endpoint (in Cassandra parlance) is always an IP address. So the idea is to determine which one we have and then either resolve it to an IP (for a name) or return it directly (for an IP).

Miles-Garnsey: Now, we could use some other function (e.g. I came across an apache-commons validation package) but that would involve either using a method marked `beta` from Guava, or adding a new dependency. So I copied and pasted this regex from [here](https://www.techiedelight.com/validate-ip-address-java/) rather than doing that.

burmanm: Like I said InetAddress.getByName should accept both, hostname and IP address. So why not use that? Or if you really wanted to separate them somehow, then InetAddress can also throw an exception if the format is not an IP (4 or 6) and you can then use that as the hostname directly.

So I fail to see why we need the regexp. I'd also steer away from that page you linked, it has naughty errors, like this:

``The IP address 172.8.7.28 isn’t valid``

That's certainly a valid IP address. And if you really liked that page, then the method 8 uses InetAddress.

adejanovski: Not directly related to the topic: @burmanm, it's either an issue or a question :) pick one when posting a comment please otherwise the contributor is left in an uncertain state with a blocking/non-blocking comment.
Thanks!

emerkle826: Issue:
As Michael said, we should be able to simply use `InetAddress.getByName()` with either a hostname or an IP address.It will use DNS to resolve a hostname to an IP. If the `host` attribute is already an IP address, `getByName()` will validate that it is a valid IP address format, so we don't have to maintain a regex to handle that.

emerkle826 (changes requested):

src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
Miles-Garnsey: Ah, OK. I didn't realise that method would accept EITHER a hostname or an IP. Changed as requested.

emerkle826 (approved):
As long as you get the checkstyle stuff fixed, this is good to go.

PR #1393, opened 2023-09-05
Implement listTablesByKeyspace() in HttpManagementProxy

### LINKED ISSUE #1384: Implement listTablesByKeyspace() in the HttpManagementProxy
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=37693727)

Method signature:
`public Map<String, List<String>> listTablesByKeyspace()`

adejanovski:
Fixes #1384

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/management/jmx/JmxCassandraManagementProxy.java
+++ b/src/server/src/main/java/io/cassandrareaper/management/jmx/JmxCassandraManagementProxy.java
@@ -518,7 +518,7 @@ public final class JmxCassandraManagementProxy implements ICassandraManagementPr
   }
 
   @Override
-  public Map<String, List<String>> listTablesByKeyspace() {
+  public Map<String, List<String>> listTablesByKeyspace() throws ReaperException {
     Map<String, List<String>> tablesByKeyspace = Maps.newHashMap();
     try {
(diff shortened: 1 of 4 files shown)


src/server/src/main/java/io/cassandrareaper/management/jmx/JmxCassandraManagementProxy.java
```diff
@@ -518,7 +518,7 @@ public void cancelAllRepairs() {
   }
 
   @Override
-  public Map<String, List<String>> listTablesByKeyspace() {
+  public Map<String, List<String>> listTablesByKeyspace() throws ReaperException {
```
Miles-Garnsey, line 521: Question: I don't know about this. For example, this method is called in the [ClusterFacade](https://github.com/thelastpickle/cassandra-reaper/blob/cf727084454145c0c70e806f5aa57f117cef45c2/src/server/src/main/java/io/cassandrareaper/management/ClusterFacade.java#L399-L401), which returns the response directly. When called there, won't this exception being hit actually end up being an unhandled exception?

There is no try/catch block at that callsite.

Miles-Garnsey (commented):
I have questions about the exception handling in these changes.

Miles-Garnsey (approved):
LGTM.

PR #1395, opened 2023-09-05
Implement topology methods in the HttpCassandraManagementProxy (fixes #1341)

### LINKED ISSUE #1341: Implement topology methods in the HttpCassandraManagementProxy
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=35567388)

## Definition of Done
- [x] the `getLocalEndpoint()` method is implemented in the HttpCassandraManagementProxy class
- [x] the `getSimpleStates()` method is implemented in the HttpCassandraManagementProxy class
- [x] the `getAllEndpointStates()` method is implemented in the HttpCassandraManagementProxy class

olim7t:
Fixes #1341

Notes:
- the ticket mentioned `getLocalHostId()` but there is no such method, I assumed it was `getLocalEndpoint()`.
- instead of implementing `getSimpleStates()` and `getAllEndpointStates()`, I merged them into a single `getNodesStatus()`, which makes much more sense given that the HTTP API is able to provide data that is already more structured.


src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
```diff
@@ -246,7 +264,7 @@ public int triggerRepair(
                       (new com.datastax.mgmtapi.client.model.RingRange())
                           .start(i.getStart().longValue())
                           .end(i.getEnd().longValue())
-                ).collect(Collectors.toList())
+                  ).collect(Collectors.toList())
```
olim7t: There are a couple of unrelated formatting changes due to the new IntelliJ rules, but most of them seem justified. I can revert them if necessary.

src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
```diff
@@ -119,7 +120,24 @@ public Map<List<String>, List<String>> getRangeToEndpointMap(String keyspace) th
   @NotNull
   @Override
   public String getLocalEndpoint() throws ReaperException {
-    return null; // TODO: implement me.
+    try {
```
Miles-Garnsey, line 139: Issue: my understanding (which may be entirely wrong!) was that an endpoint was always an IP address, while a hostID was always a UUID. Is your understanding different @olim7t ?

If I'm right, no need to implement this as I have implemented [here](https://github.com/thelastpickle/cassandra-reaper/pull/1376/files).

src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
Miles-Garnsey: I'm fine with it.

src/server/src/test/java/io/cassandrareaper/resources/view/NodesStatusHttpTest.java
```diff
@@ -0,0 +1,242 @@
...
+
+  @Test
+  public void testRegularValues() {
+    testSingleNode(
+        ImmutableMap.<String, String>builder()
+            .put("DC", "dc1")
+            .put("RACK", "rack1")
+            .put("ENDPOINT_IP", "10.0.0.1")
+            .put("IS_ALIVE", "true")
+            .put("STATUS_WITH_PORT", "NORMAL,-1181245519517832414")
```
Miles-Garnsey, line 47: Suggestion: that integer doesn't look like a port.

Miles-Garnsey (changes requested):
So from my understanding, what you're just doing here is basically making the methods `getSimpleStates()` and `getAllEndpointStates()` private to the JMX implementation, and exposing a uniform `getNodesStatus()` across both implementations and the interface? 

I'm fine with that part of it I think, as the tests still pass. RE The FailureDetectorProxy I'm not really sure what it was ever there for given it was just a facade to the XCassandraManagementProxy implementation. so that's fine too. I guess someone started a refactor and didn't finish it.

I have a problem/question with the `getLocalEndpoint()` implementation, see my comments below.

src/server/src/test/java/io/cassandrareaper/resources/view/NodesStatusHttpTest.java
olim7t: I copied the output of an actual Cassandra instance. I'm not sure why it's formatted that way, or if it even refers to "port" in the network sense. At the end of the day it doesn't matter much since we discard everything after the comma.

src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
olim7t: You're right it should be an IP, I misread the JMX implementation. I'll fix it.

src/server/src/test/java/io/cassandrareaper/resources/view/NodesStatusHttpTest.java
Miles-Garnsey: Weird, OK. I'll resolve this as it seems that you've done things the right way.

src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
Miles-Garnsey: @adejanovski, a quick question on this. Both Olivier and I have implementations for this method. His relies on a response returned from the `GET endpoints` call, while mine uses the internal field `host` within Reaper. 

There are different implications. We don't have address translators right now, so mine will more likely return the external IP for the node, while Olivier's will more likely return the internal IP. Which one is best to go with until we implement address translation properly?

adejanovski: The internal IP of the Cassandra node is what we'll need in this case, which is the one broadcasted by Cassandra.

Miles-Garnsey: OK, this is good to merge then. @olim7t  when you rebase it, feel free to overwrite my `getLocalEndpoint()` implementation with yours.

Miles-Garnsey (approved):
LGTM.

src/server/src/test/java/io/cassandrareaper/management/http/HttpCassandraManagementProxyTest.java
```diff
@@ -330,6 +330,60 @@ public void testGetCompactions() throws Exception {
...
+    ));
+    when(mockClient.getEndpointStates()).thenReturn(states);
+
+    HttpCassandraManagementProxy proxy = mockProxy(mockClient);
+    String localEndpoint = proxy.getLocalEndpoint();
+
+    assertThat(localEndpoint).isEqualTo("10.0.0.2");
+    verify(mockClient).getEndpointStates();
+
+    String untranslatedHost = proxy.getUntranslatedHost();
```
olim7t: question: do we still need `getUntranslatedHost` if it's just a pass-through to `getLocalEndpoint`?

Miles-Garnsey: Yes, because once address translators are added, these two will do something different.

olim7t: Does "untranslated" refer to the internal IP or the external one?

Miles-Garnsey: I assume the external one is "translated" from the internal one. So the external one - @adejanovski can you advise?

adejanovski: Unstranslated should refer to the internal IP. Address translation is used (IIRC) to translate private to public IPs for nodes running in remote regions and not accessible through private IPs.

olim7t: OK, I think the two methods are the same, but I will leave `getUntranslatedHost()` with the TODO comment until this gets tested with address translation.

PR #1399, opened 2023-09-07
Implement isRepairRunning

### LINKED ISSUE #1383: Remove isRepairRunning() from the HttpManagementProxy, ICassandraManagementProxy
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=37693691)

Method signature:
`public boolean isRepairRunning() throws JMException`

emerkle826:
Fixes #1383 
There was a small refactor here to get `JMException` out of the parent interface.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
+++ b/src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
@@ -46,7 +46,6 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
-import javax.management.JMException;
 import javax.management.openmbean.CompositeData;
 import javax.validation.constraints.NotNull;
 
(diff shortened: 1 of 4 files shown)


src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
```diff
@@ -229,13 +228,13 @@ public Set<Table> getTablesForKeyspace(String keyspace) throws ReaperException {
   }
 
   @Override
-  public int getPendingCompactions() throws JMException {
+  public int getPendingCompactions() throws ReaperException {
     return 1; // TODO: implement me.
   }
 
   @Override
```
Miles-Garnsey, line 235: Issue: We discussed this before, but just recording here so you have something to refer back to. I think the way this works in the JMX implementation is to go to Cassandra's JMX metrics registry and draw whether a repair is in progress from there.

The problem with just using job tracker, is that (AFAIK) it only tracks repairs initiated by this instance of Reaper. A repair started elsewhere (e.g. via nodetool) might not make its way into the job tracker, which means that this logic might miss running repairs. 

I believe that @adejanovski has been doing a lot of work around getting metrics from the new metrics endpoint here, so it might be worth syncing with him to discuss how he's done this, you may need to wait on some refactoring he's doing to the `MetricsProxy`.

Miles-Garnsey (changes requested):
See my comment down the page RE needing to refer back to Cassandra's metrics instead of just using job tracker.

One thing I might be open to is accepting this implementation to progress the work short term and uncover more issues more quickly, but then raising a new ticket to use the metrics implementation before GA. What do you think?

adejanovski (commented):
This method isn't used anymore outside of some mocks that set it's return value.
The orchestration of segments used to rely on this method until it was revamped in [Reaper 2.2](https://thelastpickle.com/blog/2021/02/22/reaper-for-apache-cassandra-2-2-release.html) where we started relying on Reaper's storage for this.
I think we can close this as is for now and refactor later to cleanup the code and remove it everywhere.

Miles-Garnsey:
We do not need to implement this, we intend to remove the method entirely in future. A no op is fine for now.

PR #1400, opened 2023-09-07
Remove close from proxy interface

### LINKED ISSUE #1385: Implement close() in the HttpManagementProxy
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=37693751)

Method signature:
`public void close()`

Probably a no op in this situation but maybe we need to unregister some notification related stuff.

emerkle826:
Fixes #1385 

`close()` is really only called by the JMX implementation to get rid of stale JMX connections. We son't have that situation with the Management API client, nor is there any real way to _close_ the client. So this essentially refactors the parent interface to remove `close()` and only implement it within the JMX proxy and factory classes.


Miles-Garnsey (approved):
A small PR is a good PR ;)

PR #1403, opened 2023-09-08
Implement getPendingCompactions in HttpManagementProxy

### LINKED ISSUE #1382: Implement getPendingCompactions() in the HttpManagementProxy
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=37693681)

Method signature:
`public int getPendingCompactions() throws JMException`

### LINKED ISSUE #1396: Implement the HttpMetricsProxy in Reaper
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=37820250)

When using the HttpManagementProxy, as metrics aren't exposed in the management API we need to query the metrics endpoint directly in the mgmt-api agent.
This endpoint runs in a separate jvm on a different port which has to be made configurable in Reaper since it can be modified in the mgmt-api config.
This endpoint being a Prometheus compliant one, we'll bring in the [Prometheus Java Client](https://github.com/prometheus/client_java) in order to interact with it.

The private `collectMetrics()` method will be implemented to match the requirements for the Prometheus client and the JmxStat class will be renamed to avoid implementation specific names. We can reuse it to store both metrics coming from JMX and from the mgmt-api.

## Definition of Done
- [x] the HttpMetricsProxy class is fully implemented
- [x] Metrics are pulled from the mgmt-api using the Prometheus client
- [x] A new config setting is added to allow modifying the mgmt-api port exposing the metrics

adejanovski:
Fixes #1382 

Since metrics are now exposed on a different port than the mgmt-api itself, this required to go through the HttpMetricsProxy which got partially implemented for that need.
It is now able to pull metrics from the metrics endpoint and parse it into JmxStat objects (which will need to be later refactored to a new name).
Only tpstats are considered for now as the others are not needed for segments orchestration.


src/server/src/main/java/io/cassandrareaper/management/http/HttpMetricsProxy.java
```diff
@@ -18,48 +18,187 @@
...
-  public List<GenericMetric> collectPercentRepairedMetrics(Node node, String keyspaceName)
+  public List<GenericMetric> collectPercentRepairedMetrics(String keyspaceName)
       throws JMException, IOException {
     throw new UnsupportedOperationException("Not implemented");
   }
+
+  private Map<String, List<JmxStat>> collectMetrics() throws IOException {
+    // TODO: add a short lived cache to avoid querying the metrics endpoint too often
+    Request request = new Request.Builder()
+        .url("http://" + proxy.getHost() + ":" + proxy.getMetricsPort() + "/metrics")
```
burmanm: issue: This will cause serious load on Reaper node (and the Cassandra node) as it will transfer all the metrics through network. This is the reason the filtering method was created, where you can append the name filtering to get the exact metric (and nothing else).

So the syntax is: ``http://localhost:9000/metrics?name=org_apache_metrics_threadetc`` and that would return only the org_apache_metrics_threadetc metric (exact name is required).

src/server/src/main/java/io/cassandrareaper/management/http/HttpMetricsProxy.java
```diff
@@ -17,49 +17,211 @@
...
+      if (labelString != null) {
+        Matcher labelMatcher = labelPattern.matcher(labelString);
+        while (labelMatcher.find()) {
+          labels.put(labelMatcher.group(1), labelMatcher.group(2));
+        }
+      }
+
+      parsedMetrics.computeIfAbsent(metricName, k -> new ArrayList<JmxStat>());
+
+      Optional<JmxStat> metric = convertToJmxStat(metricName, labels, value);
```
burmanm: question: Why are these metrics converted to JmxStat instead of creating directly the GenericMetric class, which is what the methods will return? I feel like now we have to create a refactoring ticket to remove JmxStat from HttpMetrics as this class isn't doing any JMX work.

src/server/src/main/java/io/cassandrareaper/management/http/HttpMetricsProxy.java
```diff
@@ -17,49 +17,211 @@
...
+      HttpCassandraManagementProxy proxy, Node node, Supplier<OkHttpClient> httpClientSupplier) {
+    return new HttpMetricsProxy(proxy, node, httpClientSupplier);
   }
 
   @Override
-  public List<GenericMetric> collectTpStats(Node node) throws JMException, IOException {
-    throw new UnsupportedOperationException("Not implemented");
+  public List<GenericMetric> collectTpStats() throws JMException, IOException {
+    // Collect all metrics and filter out the ones that are not thread pool metrics
+    List<GenericMetric> genericMetrics = MetricsProxy.convertToGenericMetrics(collectMetrics(), node);
```
burmanm: issue: This isn't calling ``collectMetrics`` with correct parameters. Should be ``collectMetrics(Optional.of(THREAD_POOL_METRICS_PREFIX))``.

src/server/src/main/java/io/cassandrareaper/management/http/HttpMetricsProxy.java
adejanovski: Right, I was under the impression initially that the interface was exposing JmxStats as return type, but collectMetrics isn't defined in it and is a private method.
I'll get rid of JmxStat there.

src/server/src/main/java/io/cassandrareaper/management/MetricsProxy.java
```diff
@@ -17,39 +17,149 @@
...
 import javax.management.JMException;
 
+import com.google.common.collect.Lists;
+import org.joda.time.DateTime;
+
+
 public interface MetricsProxy {
 
-  static MetricsProxy create(ICassandraManagementProxy proxy) {
+  static MetricsProxy create(ICassandraManagementProxy proxy, Node node) {
```
Miles-Garnsey, line 44: Suggestion: Not sure about this. I'd almost rather this switch (between HTTP and JMX metrics proxies) happened in ReaperApplication. I don't think this interface should know whether it is working on a JMX or HTTP backend.

src/server/src/main/java/io/cassandrareaper/management/MetricsProxy.java
```diff
@@ -17,39 +17,149 @@
...
+  List<GenericMetric> collectLatencyMetrics() throws JMException, IOException;
 
-  List<GenericMetric> collectGenericMetrics(Node node) throws JMException, IOException;
+  List<GenericMetric> collectGenericMetrics() throws JMException, IOException;
 
-  List<GenericMetric> collectPercentRepairedMetrics(Node node, String keyspaceName)
+  List<GenericMetric> collectPercentRepairedMetrics(String keyspaceName)
       throws JMException, IOException;
+
+  static List<GenericMetric> convertToGenericMetrics(Map<String, List<JmxStat>> jmxStats, Node node) {
```
Miles-Garnsey, line 67: Suggestion: Would it be worth making this non-static so you don't have to pass `node`?

src/server/src/main/java/io/cassandrareaper/management/MetricsProxy.java
```diff
@@ -17,39 +17,149 @@
...
+            .withTs(now)
+            .build();
+        metrics.add(metric);
+      }
+    }
+
+    return metrics;
+  }
+
+  static ThreadPoolStat.Builder updateGenericMetricAttribute(GenericMetric stat,
```
Miles-Garnsey, line 90: Issue: would be good to see a test for this. Copilot should be able to write something reasonable for it quite quickly I think.

Same for the other methods here.

src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
```diff
@@ -255,8 +292,15 @@ public Set<Table> getTablesForKeyspace(String keyspace) throws ReaperException {
...
-    return 1; // TODO: implement me.
+  public int getPendingCompactions() throws ReaperException {
+    List<GenericMetric> tpStatsMetrics = metricsProxy.collectTpPendingTasks();
+    for (GenericMetric metric : tpStatsMetrics) {
+      if (metric.getMetricName().equals("PendingTasks")
+          && metric.getMetricScope().equals("CompactionExecutor")) {
+        return (int) metric.getValue();
+      }
+    }
+    return 0;
```
Miles-Garnsey: Issue: I assume this only gets hit if the tpStatsMetrics doesn't have a `CompactionExecutor` item with scope = `PendingTasks`. I think in that case maybe this is an error and we should log a warning and throw, what do you think?

src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
```diff
@@ -92,14 +97,42 @@ public HttpCassandraManagementProxy(MetricRegistry metricRegistry,
...
   ) {
     this.host = endpoint.getHostString();
     this.metricRegistry = metricRegistry;
     this.rootPath = rootPath;
     this.endpoint = endpoint;
     this.apiClient = apiClient;
+    this.metricsPort = metricsPort;
     this.statusTracker = executor;
+    this.node = node;
+    this.metricsProxy = HttpMetricsProxy.create(this, node);
```
Miles-Garnsey, line 113: Suggestion: OK, this feels a bit weird. Our MetricsProxy takes a `HttpCassandraManagementProxy` as a argument to the constructor, but `HttpCassandraManagementProxy` also has an instance of the `MetricsProxy`, it feels a bit tangled. Is it possible to get some clearer separation between these classes?

Possibly this is an item for later refactoring, but I had to ask...

Miles-Garnsey: It actually looks like the proxy is called only for this line of code:

```
String url = String.format("http://%s:%s/metrics", proxy.getHost(), proxy.getMetricsPort());
```

It seems to me like maybe the HttpCassandraManagementProxy shouldn't know about the metrics port at all. And the MetricsProxy should maybe take the host as an argument to it's constructor instead of taking the whole ICassandraManagementProxy.

I haven't looked at what this would imply for the JMX side of things, but think we should try to do something a bit cleaner here.

Miles-Garnsey (changes requested):
I've done a preliminary review, I'll look more tomorrow. I have a few requested changes, but most aren't really blocking. 

Given the tests are passing, I'm almost happy to approve immediately, but given the size of this PR I want to take another look over it. I can see why it took a little while!

src/server/src/test/java/io/cassandrareaper/acceptance/BasicSteps.java
```diff
@@ -2072,17 +2073,34 @@ public void we_can_collect_the_dropped_messages_stats_from_the_seed_node() throw
             if (Response.Status.OK.getStatusCode() == response.getStatus() && StringUtils.isNotBlank(responseData)) {
               List<DroppedMessages> dropped = SimpleReaperClient.parseDroppedMessagesJSON(responseData);
               if (dropped.isEmpty()) {
-                LOG.error("Got empty response from {}", seed);
+                LOG.error("Got empty DroppedMessages response from {}", seed);
```
Miles-Garnsey, line 2076: Issue: do we want the test to fail out at this point? I think maybe we do.

Miles-Garnsey: OR, maybe this isn't an error and should be logged as a warning/debug, since in a real environment I guess sometimes messages won't have been dropped...

src/server/src/main/java/io/cassandrareaper/management/MetricsProxy.java
adejanovski: A lot of the changes here support the fact that the management proxy now invokes the metrics proxy. It wasn't the case before as the metrics proxy was using the management proxy to get the metrics (through JMX). It's now the other way around with the HTTP flavor as we're not getting the metrics through the management api client, but through a separate endpoint.
It's a fundamental change between the JMX and HTTP implementations which I couldn't do differently without a deep refactor.

src/server/src/main/java/io/cassandrareaper/management/MetricsProxy.java
adejanovski: That doesn't seem to be possible since the creation of a MetricsProxy instance is done through a static method, which can't initialize a member of the implementation class.

src/server/src/main/java/io/cassandrareaper/management/MetricsProxy.java
adejanovski: sounds good 👍

src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
adejanovski: I tried to untangle them but failed so far, it's a fairly involved change.
I'd recommend to create a refactoring ticket to work this out.
And I agree that we should not keep it like this over longer term. JMX and HTTP differ pretty heavily here due to the separate endpoints for operations and metrics in the management api. It requires some time and thought to review the concept of MetricsProxy here.

src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
adejanovski: Fair enough, I thought I was reproducing the behavior of the JMX implementation, but it turns out we re-throw JMExceptions as ReaperExceptions, which should include the case where we can't find the metric we're looking for. I'll do the same here and throw a ReaperException if the metric can't be found.

src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
Miles-Garnsey: OK, can you please create a refactoring ticket before merging.

src/server/src/main/java/io/cassandrareaper/management/MetricsProxy.java
Miles-Garnsey: We can roll this into the refactor that we need to do to disentangle these classes down the track. Please include a note in your refactoring ticket that the `create()` method probably shouldn't be the thing deciding whether we want a HTTP or JMX backed implementation.

src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
Miles-Garnsey: I don't know if it is just the diff issues I'm seeing, but this doesn't appear to have been updated to throw an exception?

Miles-Garnsey (changes requested):
I mentioned on slack, but I'm having issues with reviewing this due to noise in the diff (I think the last rebase I did was responsible, need to discuss how to do this better next time).

Can we maybe rebase this against the current integration branch so I have clearer visibility?

src/server/src/test/java/io/cassandrareaper/acceptance/BasicSteps.java
adejanovski: I'm failing fast now

src/server/src/main/java/io/cassandrareaper/management/http/HttpMetricsProxy.java
```diff
@@ -17,49 +17,282 @@
...
-    throw new UnsupportedOperationException("Not implemented");
+  public List<GenericMetric> collectLatencyMetrics() throws JMException, IOException {
+    // Not mandatory for our first iteration.
+    return collectMetrics(Optional.of(CLIENT_REQUEST_LATENCY_METRICS_PREFIX), Optional.empty());
   }
 
   @Override
-  public List<GenericMetric> collectGenericMetrics(Node node) throws JMException, IOException {
+  public List<GenericMetric> collectGenericMetrics() throws JMException, IOException {
+    // Not mandatory for our first iteration.
```
Miles-Garnsey, line 100: Issue: can we choose between throwing an exception and returning an empty and do one or the other consistently in these methods. 

Suggestion: it might be good to put a TODO on these (as that gets picked up by IDEs more, and is also something people can search for), and it would also be good to raise a ticket to complete these methods too.

src/server/src/main/java/io/cassandrareaper/management/http/HttpMetricsProxy.java
```diff
@@ -17,49 +17,282 @@
...
 
   @Override
-  public List<GenericMetric> collectPercentRepairedMetrics(Node node, String keyspaceName)
+  public List<GenericMetric> collectPercentRepairedMetrics(String keyspaceName)
       throws JMException, IOException {
-    throw new UnsupportedOperationException("Not implemented");
+    return collectMetrics(Optional.of(PERCENT_REPAIRED_METRICS_PREFIX), Optional.of(keyspaceName));
+  }
+
+  private List<GenericMetric> collectMetrics(
```
Miles-Garnsey, line 110: Suggestion: It would be good to see a test here, but this is non blocking as technically I  think this method will be tested via other tests (e.g.`testCollectTpPendingTasks()`).

src/server/src/main/java/io/cassandrareaper/management/http/HttpMetricsProxy.java
```diff
@@ -17,49 +17,282 @@
...
+      Optional<String> metricNamePrefix, Optional<String> keyspaceName) throws IOException {
+    String url = String.format("http://%s:%s/metrics", proxy.getHost(), proxy.getMetricsPort());
+    if (metricNamePrefix.isPresent()) {
+      url += "?name=" + metricNamePrefix.get();
+    }
+    LOG.debug("Collecting metrics from {}", url);
+    Request request = new Request.Builder()
+        .url(url)
```
Miles-Garnsey, line 122: Suggestion: non blocking but I don't love the use of strings where a more specific type is available. In this case `HttpUrl` would give you some nicer, more expressive methods.

src/server/src/main/java/io/cassandrareaper/management/http/HttpMetricsProxy.java
```diff
@@ -17,49 +17,282 @@
...
+    // Collect metrics with the tpstats pending tasks name
+    List<GenericMetric> genericMetrics;
+    try {
+      genericMetrics
+          = collectMetrics(Optional.of(TPSTATS_PENDING_METRIC_NAME), Optional.empty());
+    } catch (IOException e) {
+      throw new ReaperException("Error collecting metrics", e);
+    }
+    return genericMetrics.stream()
+        .filter(metric -> metric.getMetricType().equals("ThreadPools"))
```
Miles-Garnsey, line 145: Question: If there is no MetricType with value `ThreadPools` is this an error that should be thrown, or is it an expected condition which should return an empty list?

Miles-Garnsey: Also, if the answer is that this should throw, can you go back through each method to ensure that they all have consistent behaviour in that regard.

src/server/src/test/resources/cassandra-reaper-http-at.yaml
```diff
@@ -93,4 +93,30 @@ cryptograph:
   systemPropertySecret: REAPER_ENCRYPTION_KEY
 
 httpManagement:
-  enabled: true
\ No newline at end of file
+  enabled: true
+
+cassandra:
```
Miles-Garnsey: Question: Is all of this new stuff required?

Miles-Garnsey (changes requested):
Some more feedback now that I can see what's happening here a bit better. I might have yet more tomorrow once I've gone through this in more depth again. 

Sorry, this is a relatively large PR, and the amount of refactoring required (because we're making compromises to get it across the line) makes the logic much harder to make out.

src/server/src/test/resources/cassandra-reaper-http-at.yaml
adejanovski: yep, we need the driver configuration to connect to Cassandra.

src/server/src/main/java/io/cassandrareaper/management/MetricsProxy.java
adejanovski: https://github.com/thelastpickle/cassandra-reaper/issues/1423

src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
adejanovski: https://github.com/thelastpickle/cassandra-reaper/issues/1423

src/server/src/main/java/io/cassandrareaper/management/http/HttpMetricsProxy.java
adejanovski: sure thing, I'll return an empty list here.
Also I think it's the only method that needs to be implemented still, the others are done (I've removed the misleading comments there).

src/server/src/main/java/io/cassandrareaper/management/http/HttpMetricsProxy.java
adejanovski: yeah, it's tested through the other methods. Testing it again would be slightly redundant.

src/server/src/main/java/io/cassandrareaper/management/http/HttpMetricsProxy.java
adejanovski: ok, doing that.

src/server/src/main/java/io/cassandrareaper/management/http/HttpMetricsProxy.java
adejanovski: I'm not sure about the runtime behavior here, so I'd go with returning an empty list if no metric matches the requested metric type.

src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
adejanovski: It should appear after the branch was squashed and rebased.

src/server/src/main/java/io/cassandrareaper/management/http/HttpMetricsProxy.java
emerkle826: That is the case. All of the public `collectXYZ` methods just call this method, so testing them tests this method.

src/server/src/main/java/io/cassandrareaper/management/http/HttpMetricsProxy.java
Miles-Garnsey: I think this has been resolved by allowing the method to take an optional of the metric name and filter on that.

src/server/src/main/java/io/cassandrareaper/management/MetricsProxy.java
Miles-Garnsey: Appears to have been added

src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
Miles-Garnsey: Alex has said he's creating a refactoring ticket.

src/server/src/test/resources/cassandra-reaper-http-at.yaml
Miles-Garnsey: I'm still waiting on a response here.

Miles-Garnsey (approved):
@adejanovski is going to fix a duplicated block in the Reaper config for the acceptance tests, and also link to the metrics proxy refactoring ticket he raised.

Otherwise, this is approved.

adejanovski:
Here's the follow up ticket: #1423

PR #1408, opened 2023-09-14
Fix http factory DC tracking and add notification to triggerRepair

emerkle826:
- Fixes the triggerRepair call to add `notifications(true)`.
- Fixes the Http Proxy Factory class so that it actually adds Datacenters to the available list


emerkle826:
@Miles, this is going to need a bugfix in Management API to be merged, and then a new client built. The PR for that is here: https://github.com/k8ssandra/management-api-for-apache-cassandra/pull/382

Miles-Garnsey (approved):
LGTM.

PR #1410, opened 2023-09-15
Add version compatibility info to the README.

### LINKED ISSUE #1411: Document versions
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=38675037)

We should document versions for things like JRE, Node, Cassandra.

Miles-Garnsey:
We don't have any version compatibility info right now. We need to add some because our version of node is ancient.

Fixes: https://github.com/thelastpickle/cassandra-reaper/issues/1411

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,13 @@ be used only for testing purposes.
 This project is built on top of Dropwizard:
 http://dropwizard.io/
 
+Version compatibility 
+------------
+Reaper can be built using Java 8 or 11.  It is tested against Cassandra 3.11 and 4.0.  It is no longer tested against Cassandra 2.x.
+
+We have confirmed the Reaper UI will build with npm 5.6.0, node 10.0.0. 
+
+
 Dependencies
 ------------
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -40,10 +40,17 @@ be used only for testing purposes.
 This project is built on top of Dropwizard:
 http://dropwizard.io/
 
+Version compatibility 
+------------
+Reaper can be built using Java 8 or 11.  It is tested against Cassandra 3.11 and 4.0.  It is no longer tested against Cassandra 2.x.
+
+We have confirmed the Reaper UI will build with npm 5.6.0, node 10.0.0.
```
adejanovski: **todo:** It also builds with npm up to 6.14 and both node 12.x and 14.x. (it fails with 16+)

adejanovski (changes requested):

README.md
Miles-Garnsey: Updated. Obviously we haven't checked every version, so I've just said that we believe these versions will work.

adejanovski (approved):

PR #1413, opened 2023-09-19
Add instance caching to HttpManagementConnectionFactory

### LINKED ISSUE #1414: HttpManagementConnectionFactory does not use the same instance for all calls
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=38977664)

When connectImpl / connectAny is called from the HttpManagementConnectionFactory, it creates a new instance of HttpManagementProxy every time. This is different from the behavior of the JMXConnectionFactory and might cause additional memory usage if multiple instances of HttpManagementProxy are created for the same node.

burmanm:
…e for each connectImpl call

Fixes #1414


Miles-Garnsey (approved):
My testing suggests that this doesn't break any additional tests (beyond the acceptance tests already broken) and it helps make debugging much easier because fewer HTTP proxies are created. 

It'll probably also reduce GC pressure due to cleaning up those proxies I guess.

Nice work!

PR #1415, opened 2023-09-19
Add mTLS settings to ApiClient

### LINKED ISSUE #1449: Make TLS connection available to management-api
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=47207521)

burmanm:
This can't hot reload yet existing connections, as that requires the connection parameters modifications to be merged first (to use caching). And as such, no such code makes sense before that's done as it would be rewritten.

Fixes #1449


burmanm:
Changed base and rebased.

src/server/src/main/java/io/cassandrareaper/ReaperApplicationConfiguration.java
```diff
@@ -509,15 +507,6 @@ public void setCryptograph(@Nullable CryptographFactory cryptograph) {
     this.cryptograph = cryptograph;
   }
 
-  public int getMgmtApiMetricsPort() {
-    return mgmtApiMetricsPort == null ? DEFAULT_MGMT_API_METRICS_PORT : mgmtApiMetricsPort;
-  }
-
-  @JsonProperty("mgmtApiMetricsPort")
```
Miles-Garnsey, line 516: I presume this is going because we can't configure the port at the moment?

src/server/src/main/java/io/cassandrareaper/management/http/HttpManagementConnectionFactory.java
```diff
@@ -113,45 +134,106 @@ private void registerConnectionsGauge() {
...
-
-    HTTP_CONNECTIONS.computeIfAbsent(host, new Function<String, HttpCassandraManagementProxy>() {
+    return HTTP_CONNECTIONS.computeIfAbsent(node.getHostname(), new Function<String, HttpCassandraManagementProxy>() {
       @Nullable
       @Override
       public HttpCassandraManagementProxy apply(@Nullable String hostName) {
-        DefaultApi apiClient = new DefaultApi(
-            new ApiClient().setBasePath("http://" + hostName + ":" + managementPort + rootPath));
+        ReaperApplicationConfiguration.HttpManagement httpConfig = config.getHttpManagement();
+        boolean useMtls = httpConfig.getKeystore() != null && !httpConfig.getKeystore().isEmpty();
```
Miles-Garnsey, line 159: Issue: do we maybe want a check here to confirm that the truststore is also available (not just the keystore).

Miles-Garnsey (changes requested):
Overall I think this looks good. Needs tests though, can we add some?

src/server/src/main/java/io/cassandrareaper/ReaperApplicationConfiguration.java
burmanm: No, I moved it under the HttpManagement class where rest of the mgmtApi related settings are

src/server/src/main/java/io/cassandrareaper/management/http/HttpManagementConnectionFactory.java
burmanm: Amm, depends. If I have a proper certificate (for example from Let's Encrypt), then would I even need custom truststore? I think the trust is already correctly setup in the OS. In that case, only the keystore would be required.

src/server/src/main/java/io/cassandrareaper/ReaperApplicationConfiguration.java
Miles-Garnsey: Sorry, forgot to delete this comment after seeing where you'd moved it, all good.

src/server/src/main/java/io/cassandrareaper/management/http/HttpManagementConnectionFactory.java
Miles-Garnsey: I guess that is indeed the case. It won't be the common case for our users, but theoretically it could work so no need to insist on a custom trust store.

Miles-Garnsey:
For some reason the tests aren't running here by the way, hopefully just a GHA glitch and once you add a test for the new functionality they'll run.

adejanovski:
> For some reason the tests aren't running here by the way, hopefully just a GHA glitch and once you add a test for the new functionality they'll run.

Tests didn't run because the target branch wasn't `master`. I think we need a new commit to be pushed so that CI triggers.

adejanovski:
@burmanm, as discussed here's the path forward with this PR:

Modify the http-api integration test suite to use TLS all the time. That'll require adding the encryption related files in tree and modifying [these lines](https://github.com/thelastpickle/cassandra-reaper/blob/master/.github/scripts/run-tests.sh#L121-L122) to start the mgmt api server with TLS enabled.
Then modify the Reaper configuration for the [http api reaper config](https://github.com/thelastpickle/cassandra-reaper/blob/master/src/server/src/test/resources/cassandra-reaper-http-at.yaml) to set up TLS in Reaper for these tests.
We'll create a subsequent ticket to add a set of test to the matrix without TLS enabled (out of scope for this ticket then).

cc @Miles-Garnsey

adejanovski:
@Miles-Garnsey this is ready for another review.
Thanks!

src/server/checkstyle.xml
```diff
@@ -83,10 +83,6 @@
           <property name="format" value="[\t]"/>
           <property name="message" value="Tabs in java file"/>
         </module>
-        <module name="RegexpSinglelineJava">
```
Miles-Garnsey, line 86: Suggestion: Not sure this is intended to be here, but I personally find this checkstyle rule frustrating so I'm down with removing it if everyone else is.

src/server/src/test/resources/cassandra-reaper-https-at.yaml
```diff
@@ -0,0 +1,105 @@
+# Copyright 2017-2017 Spotify AB
```
Miles-Garnsey, line 1: Suggestion: Looks like cassandra-reaper-http-at.yaml is no longer used anywhere but still exists. Should we remove it?

src/server/src/test/resources/generate-certs.sh
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env bash
```
Miles-Garnsey, line 1: Question: looks like we commit in encryption materials above, but generate them here. Any reason for having both approaches?

src/server/src/test/resources/openssl.cnf
```diff
@@ -0,0 +1,147 @@
+# Copyright 2023-2023 DataStax Inc.
```
Miles-Garnsey, line 1: Question: why do we need a whole openssl configuration here? Is there something that we can't control from the command line flags?

Miles-Garnsey (changes requested):
I've taken a cursory look, this is quite big. 

Looks like there's a test failing. I've re-run, let's see if it is just a flake.

src/server/src/test/resources/generate-certs.sh
burmanm: This is to regenerate the provided encryption materials.

src/server/checkstyle.xml
burmanm: It is intended to be here (I talked with Alex before pushing it)

src/server/src/test/resources/openssl.cnf
burmanm: Do we want to control it with command line flags instead of files? The openssl.cnf was from management-api repository, I wanted to ensure we generate the certificates the same way.

src/server/checkstyle.xml
Miles-Garnsey: Coolio.

src/server/src/test/resources/generate-certs.sh
Miles-Garnsey: Right, why do we want to regenerate them? This is probably fine, just making sure I understand what you're doing.

src/server/src/test/resources/openssl.cnf
Miles-Garnsey: I'm actually OK with this I think, leave it as is.

Miles-Garnsey:
We still have a failing test on this one.

src/server/src/test/resources/generate-certs.sh
burmanm: Hopefully we wouldn't (or we could modify it later to regenerate on every CI run), but I like to have generators of stuff we use in the repository, in case someone would for example run into an issue on their refactoring - and wonder why something doesn't work. Having the script gives a clear indication how the certs were generated and be easier to debug - instead of having bunch of encrypted things that were done "somehow". In the same way, if this works, but integration from somewhere else doesn't - this could give a clue.

adejanovski (approved):

Issue #1424, opened 2023-10-04
Docker config for HTTP management interface


Miles-Garnsey:
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=40647548)

The Docker config file needs to be updated so that an environment variable can be used to set whether the HTTP management interface is enabled (which replaces the JMX interface).

Miles-Garnsey:
Merged to main [here](https://github.com/thelastpickle/cassandra-reaper/commit/8a26f9d858ba9ee52fc55d8e2d343d1e69845a66).

PR #1425, opened 2023-10-04
Docker config for HTTP management interface

Miles-Garnsey:
The Docker config file needs to be updated so that an environment variable can be used to set whether the HTTP management interface is enabled (which replaces the JMX interface)

**Fixes issue**

#1424


adejanovski (approved):
This looks good, but you'll need to rebase on top of the integration branch first it seems. There are a lot of commits here that don't belong to this PR.

Miles-Garnsey:
> This looks good, but you'll need to rebase on top of the integration branch first it seems. There are a lot of commits here that don't belong to this PR.

Done, not sure what happened there sorry!

PR #1431, opened 2023-10-20
Add http setup logging

emerkle826:


emerkle826:
Closing this for now. We can revisit if we need it later.

PR #1433, opened 2023-10-21
Filter out Stargate/coordinator nodes from calls to get live nodes or list tokens

### LINKED ISSUE #1432: Exceptions are being raised when using LOCAL mode with the mgmt-api
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=42255620)

When using Reaper in LOCAL dc Availability mode and the management api used for communications, exceptions are being raised and the app cannot work properly.
This is due to a wrong implementation of getUntranslatedHost() which returns the local endpoint whereas it should return the target connection host without any address translation.

adejanovski:
Fixes #1432

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
+++ b/src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
@@ -154,8 +154,11 @@ public class HttpCassandraManagementProxy implements ICassandraManagementProxy {
       EndpointStates endpointStates = apiClient.getEndpointStates();
       List<BigInteger> tokenList = new ArrayList<>();
       endpointStates.getEntity().forEach((Map<String, String> states) -> {
-        for (String token : states.get("TOKENS").split(",")) {
-          tokenList.add(new BigInteger(token));
+        // Stargate nodes are part of the endpoint states but do not have tokens
+        if (!isCoordinatorNode(states)) {
+          for (String token : states.get("TOKENS").split(",")) {
+            tokenList.add(new BigInteger(token));
+          }
         }
       });
       // sort the list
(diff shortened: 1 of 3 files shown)


src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
```diff
@@ -405,7 +408,9 @@ public List<String> getLiveNodes() throws ReaperException {
       List<String> liveNodes = new ArrayList<>();
       EndpointStates endpoints = apiClient.getEndpointStates();
       for (Map<String, String> states : endpoints.getEntity()) {
-        if (states.containsKey("IS_ALIVE") && Boolean.parseBoolean(states.get("IS_ALIVE"))) {
+        if (states.containsKey("IS_ALIVE")
+            && Boolean.parseBoolean(states.get("IS_ALIVE"))
+            && !isCoordinatorNode(states)) {
```
emerkle826: suggestion:
I think it would be a small optimization to check if it's a coordinator node first as the states should almost always have an "IS_ALIVE" key. For coordinator-only nodes(i.e. Stargate), changing the order of the checks here will reduce the number of map reads I think.
```suggestion
        if (!isCoordinatorNode(states)
            && states.containsKey("IS_ALIVE")
            && Boolean.parseBoolean(states.get("IS_ALIVE"))) {
```

emerkle826 (approved):
small nit.

src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
adejanovski: addressed 👍

PR #1443, opened 2023-11-29
Fix incorrect callback function

### LINKED ISSUE #1442: Bug: webui will generate a huge number of requests if `_refreshClusterStatus` keep getting error responses
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=46036709)

https://github.com/thelastpickle/cassandra-reaper/blob/master/src/ui/app/jsx/cluster-list.jsx#L52-L71

According to JQuery doc (https://api.jquery.com/jQuery.ajax/), `complete` function will be called whether the ajax request succeeds or fails. As a result, `_refreshClusterStatus` will set two new scheduled calls if the ajax request gets an error response (the first one by `error` and the second one by `complete`).

If `_refreshClusterStatus` keeps getting error responses (e.g. token has expired, the ajax request always gets 304 response), it will generate a huge number of requests in 10min (doubles every 30s)

upsilon2gamma:
Fixes #1442

According to JQuery doc (https://api.jquery.com/jQuery.ajax/), `complete` function will be called whether the ajax request succeeds or fails. As a result, `_refreshClusterStatus` will set two new scheduled calls if the ajax request gets an error response (the first one by `error` and the second one by `complete`).

If `_refreshClusterStatus` keeps getting error responses (e.g. token has expired, the ajax request always gets 304 response), it will generate a huge number of requests in 10min (doubles every 30s)

### DIFF
--- a/src/ui/app/jsx/cluster-list.jsx
+++ b/src/ui/app/jsx/cluster-list.jsx
@@ -64,8 +64,6 @@ const Cluster = CreateReactClass({
           },
           error: function(data) {
             console.error(this.component.props.name + " failed.");
-            this.component.setState({clusterStatuses: setTimeout(this.component._refreshClusterStatus, 30000)});
-
           }
       });
   },


adejanovski (changes requested):
Hi @upsilon2gamma,

you're right, in case of error there's some duplication of timers that will happen.
May I suggest that we keep the `complete:` callback and modify the `error:` one by removing the `this.component.setState(...` line, keeping just the console logging?
This way we'll remove some unnecessary lines of code.

Thanks!

upsilon2gamma:
@adejanovski Updated

adejanovski (approved):
Looks good, thanks!

Issue #1192, opened 2022-04-28
segmentCountPerNode inconsistent across configs/docs


SesquipedalianDefenestrator:
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=13938551)

This seems to be set to 16 or 64 in various places.  The docs say the default is 16, but when I set up a new cluster I got a default of 64 unless I explicitly set it.

16 in docs:
src/docs/content/docs/configuration/reaper_specific.md#L355 
docs/docs/configuration/reaper_specific/index.html#L526 

Looks like a default is set to 16 here: src/server/src/main/java/io/cassandrareaper/service/RepairRunService.java#L62
but here it's set to 64 and that seems to be what's actually applied:
src/server/src/main/java/io/cassandrareaper/ReaperApplicationConfiguration.java#L47

(and src/docs/content/docs/configuration/docker_vars.md seems to claim a default of 200?)

The various .yaml example docs seem to be split fairly evenly between 16 and 64

I'm not sure what the actual correct value should be for these.  The 64s seem to generally be newer than the 16s, but with the new adaptive feature lower probably makes more sense?

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/REAP-92) by [Unito](https://www.unito.io)
┆Issue Number: REAP-92

adejanovski:
Reaper's default [changed a while ago](https://github.com/thelastpickle/cassandra-reaper/commit/3c46245385db54c5f0567080014a27f11e08ddf7) as we were seeing issues with 16 segments per node, which tends to generate timeouts. 64 was considered a more conservative value over 16.
There's no "right" value for this settings as it depends on many factors, and you're right that the adaptive feature makes that less problematic.

Still, the "real" default is 64, which apparently wasn't updated everywhere in the docs or the code 😅 And this shows, if needed, that a default shouldn't be defined in multiple places...

I would be good to harmonize the value across the code and docs 👍

PR #1477, opened 2024-03-09
Set segment count per node to consistent value across defaults and docs

SesquipedalianDefenestrator:
Per https://github.com/thelastpickle/cassandra-reaper/issues/1192 the default should be 64.  So, set that everywhere, including docs


adejanovski (approved):
Thanks @SesquipedalianDefenestrator

PR #1478, opened 2024-03-11
Ignore repair_run_by_cluster_v2 rows with no corresponding repair

### LINKED ISSUE #1463: NPE when listing repairs
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=51261893)

Without an understanding of what's going on so far, I'm seeing the following error popping up on a reaper instance running in k8ssandra-operator:

```
ERROR  [2024-01-26 10:42:02,249] [dw-3200 - GET /repair_run?cluster_name=all&limit=10] i.d.j.e.LoggingExceptionMapper - Error handling a request: bf338ca94d1b3864 
java.lang.NullPointerException: null
	at io.cassandrareaper.storage.repairrun.CassandraRepairRunDao.lambda$getRepairRunsForClusterPrioritiseRunning$3(CassandraRepairRunDao.java:404)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195)
	at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1655)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:913)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:578)
	at io.cassandrareaper.storage.repairrun.CassandraRepairRunDao.getRepairRunsForClusterPrioritiseRunning(CassandraRepairRunDao.java:406)
	at io.cassandrareaper.storage.repairrun.CassandraRepairRunDao.getRepairRunsForClusterPrioritiseRunning(CassandraRepairRunDao.java:58)
	at io.cassandrareaper.resources.RepairRunResource.lambda$listRepairRuns$1(RepairRunResource.java:726)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1541)
	at java.base/java.util.Collections$UnmodifiableCollection.forEach(Collections.java:1085)
	at io.cassandrareaper.resources.RepairRunResource.listRepairRuns(RepairRunResource.java:725)
	at jdk.internal.reflect.GeneratedMethodAccessor70.invoke(Unknown Source)
```

While we need to investigate the underlying issue, we should protect ourselves from the NPE and return a proper error message while letting the lambda pursue its work.

The fact that we're [invoking](https://github.com/thelastpickle/cassandra-reaper/blob/master/src/server/src/main/java/io/cassandrareaper/storage/repairrun/CassandraRepairRunDao.java#L403) `row.getUninterruptibly().one()` without any prior check that a row exists is a bet that we may often lose, like we currently are:

```
    return repairRunFutures
        .stream()
        .map(
            row -> {
              Row extractedRow = row.getUninterruptibly().one();
              return buildRepairRunFromRow(extractedRow, extractedRow.getUUID("id"));
            }
        ).collect(Collectors.toList());
```

SesquipedalianDefenestrator:
When bad things (like Reaper filling the drive with logs) happen, it's possible to end up with repair_run_by_cluster_v2 rows with no corresponding repair row, which breaks Reaper.  So, just skip them.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/storage/repairrun/CassandraRepairRunDao.java
+++ b/src/server/src/main/java/io/cassandrareaper/storage/repairrun/CassandraRepairRunDao.java
@@ -30,6 +30,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Date;
 import java.util.List;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.SortedSet;
@@ -399,14 +400,10 @@ public class CassandraRepairRunDao implements IRepairRunDao {
     // Defuture the repair_run rows and build the strongly typed RepairRun objects from the contents.
     return repairRunFutures
         .stream()
-        .map(
-            row -> {
-              Row extractedRow = row.getUninterruptibly().one();
-              return (extractedRow == null)
-                      ? (null)
-                      : (buildRepairRunFromRow(extractedRow, extractedRow.getUUID("id")));
-            }
-        ).filter(x -> x != null).collect(Collectors.toList());
+        .map(row -> row.getUninterruptibly().one())
+        .filter(Objects::nonNull)
+        .map(extractedRow -> buildRepairRunFromRow(extractedRow, extractedRow.getUUID("id")))
+        .collect(Collectors.toList());
   }
 
   @Override
--- a/src/server/src/main/java/io/cassandrareaper/storage/repairrun/CassandraRepairRunDao.java
+++ b/src/server/src/main/java/io/cassandrareaper/storage/repairrun/CassandraRepairRunDao.java
@@ -395,15 +395,18 @@ public class CassandraRepairRunDao implements IRepairRunDao {
                 )
         )
     );
+
     // Defuture the repair_run rows and build the strongly typed RepairRun objects from the contents.
     return repairRunFutures
         .stream()
         .map(
             row -> {
               Row extractedRow = row.getUninterruptibly().one();
-              return buildRepairRunFromRow(extractedRow, extractedRow.getUUID("id"));
+              return (extractedRow == null)
+                      ? (null)
+                      : (buildRepairRunFromRow(extractedRow, extractedRow.getUUID("id")));
             }
-        ).collect(Collectors.toList());
+        ).filter(x -> x != null).collect(Collectors.toList());
   }
 
   @Override


SesquipedalianDefenestrator:
I'm not actually sure if anything will clean up the repair_run_by_cluster_v2 rows, but I'm not sure it's a problem.  The mismatch between repair_run and repair_run_by_cluster_v2 is frequent enough to be a pain (basically, a percentage of the time when anything goes wrong with the cluster and Reaper logs until the drive fills), but not often enough to generate measurable load.

src/server/src/main/java/io/cassandrareaper/storage/repairrun/CassandraRepairRunDao.java
```diff
@@ -395,15 +395,18 @@ public List<RepairRun> getRepairRunsForClusterPrioritiseRunning(String clusterNa
...
     );
+
     // Defuture the repair_run rows and build the strongly typed RepairRun objects from the contents.
     return repairRunFutures
         .stream()
         .map(
             row -> {
               Row extractedRow = row.getUninterruptibly().one();
-              return buildRepairRunFromRow(extractedRow, extractedRow.getUUID("id"));
+              return (extractedRow == null)
```
FieteO: I would suggest to rewrite this to do less in the `map`. That would make it more readable. Something like
```java
.stream()
.map(row -> row.getUninterruptibly().one())
.filter(Objects::nonNull)
.map(extractedRow -> buildRepairRunFromRow(extractedRow, extractedRow.getUUID("id"))
.collect(Collectors.toList());
```

SesquipedalianDefenestrator: Thanks, that's definitely cleaner.

SesquipedalianDefenestrator:
Looks like there's an issue for this already as well:  https://github.com/thelastpickle/cassandra-reaper/issues/1463

Miles-Garnsey:
I'm looking at this, having some issues with my local setup as far as testing goes so bear with me while I try to work around them.

Miles-Garnsey (approved):
In terms of testing, I've followed the procedure below:

Try to repro the issue from `main` by:
1. First commenting out the `testSSLHotReload` (which is a problem to run locally) then building the repo into a docker image, `kind load`ing it into a `kind` cluster, spinning up a K8ssandraCluster with the new Reaper image.
2. Creating a new repair via the UI.
3. Confirming the repair exists in both `repair_run_by_cluster_v2` and `repair_run` 
4. Deleting from `repair_run` via `delete from repair_run where id = d03c24a0-e59e-11ee-9a67-5dd395ae716d`

When reloading the UI I then see:

image

Which I think is what we want based on the error repro'd [here](https://github.com/thelastpickle/cassandra-reaper/issues/1463#issue-2102057828).

I then go and rebuild the image and re-deploy everything using the branch from this PR. Following the same procedure, I no longer see the NPE visible in Reaper's logs. Instead, the UI simply returns no results for repair runs in the DONE or RUNNING state. 

I'm going to chat to @adejanovski about whether we need some more work to ensure that repair runs are purged from `repair_run_by_cluster_v2`, and whether this data inconsistency should be logged as a warning. But I'll approve this for now on the basis that it eliminates the NPE.

PR #1509, opened 2024-06-21
Subrange incremental repair

### LINKED ISSUE #1502: Segment count for incremental repairs remains 1 per node regardless of the segment count specified (3.2.0 + )
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=61610099)

When specifying segment count for incremental repairs in repair, the value seems to be ignored (including any defaults set in config)

Unsure if this is expected behaviour, as segment count/node can be specified for full repairs. 

This causes failures and backoffs during incrementals on large segments.

Screenshots of launching incremental repair and specifying segment count/node and the resulting segment count of the running repair (3 node cluster 1 per node only, instead of 64 per node)

reaper3 2 0-incremental-repair-config
reaper3 2 0-incremental-repair-running

adejanovski:
Fixes #1502 

Before reviewing, please ensure you're aware of [our PR review guidelines](https://docs.k8ssandra.io/contribute/pr_review/).

Here are the key changes in this PR:

### Dependency Update
- **pom.xml**:
  - The Checkstyle version is updated from `8.19` to `10.17.0`. This might include enhancements, bug fixes, or new features provided by the newer Checkstyle version.

### New Configuration Options for Incremental Repair
- **api.md**:
  - The `incrementalRepair` option is expanded to clarify that it performs incremental repair on all tokens of each node at once.
  - A new option `subrangeIncrementalRepair` is introduced. This allows incremental repair to be done in subrange mode, targeting discrete token ranges. This option is available starting from Cassandra 4.0.
  
- **docker_vars.md** and **reaper_specific.md**:
  - Added `REAPER_SUBRANGE_INCREMENTAL` environment variable to support subrange incremental repairs.
  - Documentation updates to explain the purpose and usage of the `subrangeIncrementalRepair` setting.

### Configuration Changes
- **Dockerfile and Environment Variables**:
  - Environment variable `REAPER_SUBRANGE_INCREMENTAL` is set to `false` by default in Docker configurations, allowing users to enable it as needed.

### Checkstyle Configuration Update
- **checkstyle.xml**:
  - Updated to the new Checkstyle DTD to `https://checkstyle.org/dtds/configuration_1_3.dtd`.
  - Some reorganization and cleanup in the Checkstyle configuration for better readability and alignment with the new Checkstyle version.

### CLI Enhancements
- **spreaper**:
  - Added a new argument `--subrange-incremental` to the CLI tool, allowing users to specify whether to perform subrange incremental repairs directly from the command line.

### Code Adjustments for Repair Functionality
- **ReaperApplication.java**:
  - Added logging for the new `subrangeIncrementalRepair` configuration to ensure it is correctly read and applied.
  
- **AppContext.java** and **ClusterFacade.java**:
  - Adjusted the classes to include support for subrange incremental repairs.

- **RepairUnit.java**:
  - Updated to include a new boolean field `subrangeIncrementalRepair`.
  - Adjusted constructor and builder pattern to accommodate the new field.
  
- **RepairRunService.java**, **RepairManager.java**, **RepairRunner.java**, and **SegmentRunner.java**:
  - Various methods updated to handle subrange incremental repairs. This includes segment generation, segment repair initiation, and leader election for repair segments.
  - Subrange incremental repair segments are treated differently from regular incremental repair segments, ensuring proper handling and preventing overlapping repairs.

- **RepairType.java**:
  - A new enum to clearly define and distinguish between different repair types: `SUBRANGE_FULL`, `INCREMENTAL`, and `SUBRANGE_INCREMENTAL`.

### Database Interaction
- **RepairUnitService.java**:
  - Added checks to ensure `subrangeIncrementalRepair` is only enabled for Cassandra versions 4.0 and above.
  - Validation added to prevent enabling both `incrementalRepair` and `subrangeIncrementalRepair` in incompatible versions of Cassandra.

### Miscellaneous Adjustments
- **RepairRunResource.java** and **RepairScheduleResource.java**:
  - Updated to accept and validate the new `subrangeIncrementalRepair` parameter.
  - Ensured compatibility and correct application of the new setting during repair run and schedule creation.

By introducing `subrangeIncrementalRepair`, this PR enhances the repair flexibility for large clusters, particularly those running on Cassandra 4.0 or later. This mode allows for more efficient repairs by dividing the repair process into smaller, manageable subranges, potentially improving performance and reducing repair time.


src/server/src/main/java/io/cassandrareaper/core/RepairType.java
```diff
@@ -0,0 +1,43 @@
...
+    SUBRANGE_INCREMENTAL("subrange_incremental");
+
+  private final String repairType;
+
+  RepairType(String repairType) {
+    this.repairType = repairType;
+  }
+
+  public String getRepairType() {
+    return repairType;
```
michaelsembwever:
```suggestion
    return name().toLowercase();
```
(and remove the field) …? 


and fromName just
```
public static RepairType fromName(String name) {
    return valueOf(name.toUppercase());
}
```

src/server/src/main/java/io/cassandrareaper/core/RepairUnit.java
```diff
@@ -42,7 +43,9 @@ private RepairUnit(Builder builder, UUID id) {
     this.clusterName = builder.clusterName;
     this.keyspaceName = builder.keyspaceName;
     this.columnFamilies = builder.columnFamilies;
-    this.incrementalRepair = builder.incrementalRepair;
+    // If subrange incremental repair is enabled, incremental repair must also be enabled
+    this.incrementalRepair = builder.incrementalRepair || builder.subrangeIncrementalRepair;
```
michaelsembwever: comment doesn't match.

can this be enforced at the config layer ?

src/server/src/main/java/io/cassandrareaper/management/jmx/JmxCassandraManagementProxy.java
```diff
@@ -617,14 +618,14 @@ private int triggerRepairPost2dot2(
     Map<String, String> options = new HashMap<>();
 
     options.put(RepairOption.PARALLELISM_KEY, repairParallelism.getName());
-    options.put(RepairOption.INCREMENTAL_KEY, Boolean.toString(!fullRepair));
+    options.put(RepairOption.INCREMENTAL_KEY, Boolean.toString(repairType != RepairType.SUBRANGE_FULL));
```
michaelsembwever: nit: for readability maybe add methods to enum
`repairType.isFull()`
and
`repairType.isIncrement()`
(and `repairType.isSubrange()`)

src/server/src/main/java/io/cassandrareaper/service/RepairManager.java
```diff
@@ -256,7 +256,7 @@ private void resumeUnknownPausedRepairRuns(Collection<RepairRun> pausedRepairRun
 
   private void abortSegmentsWithNoLeader(RepairRun repairRun, Collection<RepairSegment> runningSegments) {
     RepairUnit repairUnit = context.storage.getRepairUnitDao().getRepairUnit(repairRun.getRepairUnitId());
-    if (repairUnit.getIncrementalRepair()) {
+    if (repairUnit.getIncrementalRepair() && !repairUnit.getSubrangeIncrementalRepair()) {
```
michaelsembwever: same nit: add a single method for readability

michaelsembwever:
ton of work here, lgtm, just some comments wrt readability and safety.

what does concern me is if this addresses the primary reason (as far as we know) to want to use subrange IR – to break up and pace out the IR+anti-compactions to a rate and load that stabilises the cluster… 

how do we know subranges are optimal for this ?

src/server/src/main/java/io/cassandrareaper/core/RepairUnit.java
adejanovski: I'll update the comment, I ended up making it seamless. You can enable either or any of these flags, it will run an incremental repair.

adejanovski: Also, from the UI perspective this is transparent as we now select a "repair type" which can be either of the three types. No need to select both incremental and subrange incremental there.

src/server/src/main/java/io/cassandrareaper/core/RepairType.java
adejanovski: sounds good 👍

src/server/src/main/java/io/cassandrareaper/management/jmx/JmxCassandraManagementProxy.java
adejanovski: yep, good idea 👍

adejanovski:
> what does concern me is if this addresses the primary reason (as far as we know) to want to use subrange IR – to break up and pace out the IR+anti-compactions to a rate and load that stabilises the cluster…
how do we know subranges are optimal for this ?

We don't, and I guess it depends on factors that we don't control.
Since this has been requested and is allowed by Cassandra 4, I think it's ok to give the option and let users find the best solution for their clusters. The good part here is that the number of segments is tunable, so one can find a sweet spot even with large numbers of vnodes.

adejanovski:
Comments addressed, ready for another review @michaelsembwever

michaelsembwever (approved):

PR #1524, opened 2024-09-20
REAP-10 Allow setting up per-cluster TLS connections

rzvoncek:
This PR allows Reaper to use cluster-specific trust/key-stores when using HTTP instead of JMX (meaning Reaper is talking to [mgmt-api](https://github.com/k8ssandra/management-api-for-apache-cassandra)).

 The Reaper's yaml file can now have a new `httpManagement.truststoresDir` field:
 ```
httpManagement:
  enabled: ${REAPER_HTTP_MANAGEMENT_ENABLE}
  keystore: ${REAPER_HTTP_MANAGEMENT_KEYSTORE_PATH}
  truststore: ${REAPER_HTTP_MANAGEMENT_TRUSTSTORE_PATH}
  truststoresDir: ${REAPER_HTTP_MANAGEMENT_TRUSTSTORES_DIR}
```

This should point to a directory where per-cluster certificates will appear. The certs are expected to have format of `clusterName-truststore|keystore.jks`.

When adding a cluster, it is now necessary to specify the cluster name via the seed host, so the seed host becomes `seedHost@clusterName`. The `clusterName` should match the prefix of in the cert names in `truststoresDir`. If the cluster name is not provided, Reaper falls back to using the global keystores. 

Hot reloading of the cluster-specific stores works the same way as for the global ones.

### CODE REVIEW

File: src/server/src/main/java/io/cassandrareaper/ReaperApplication.java [hunk 1]
```diff
@@ -427,6 +430,19 @@ private void checkConfiguration(ReaperApplicationConfiguration config) {
     LOG.debug("repairParallelism: {}", config.getRepairParallelism());
     LOG.debug("hangingRepairTimeoutMins: {}", config.getHangingRepairTimeoutMins());
     LOG.debug("jmxPorts: {}", config.getJmxPorts());
+
+    if (config.getHttpManagement() != null) {
+      if (config.getHttpManagement().isEnabled()) {
+        if (config.getHttpManagement().getTruststoresDir() != null) {
+          if (!Files.exists(Paths.get(config.getHttpManagement().getTruststoresDir()))) {
+            throw new RuntimeException(String.format(
```
burmanm, line 438: issue: There's odd logic here. If getTrustStoresDir() is null, then we won't have any RuntimeException.

burmanm: Or do you rely on the truststore value to cause the RuntimeException somewhere else?

rzvoncek: Yeah I've added it here, in a `checkConfiguration()`, so that it fails early. Then the stacktrace is smaller and more obvious.

Otherwise, the RuntimeException would happen later, when it's trying to setup the watchers for changes, which in turn happens when actually adding a cluster. The stack traces at this point are incredibly long and nested.

.github/workflows/ci.yaml
```diff
@@ -507,6 +507,58 @@ jobs:
 
       - uses: codecov/codecov-action@v1
 
+  snapshot-release:
```
burmanm: discussion/issue: I'm not a huge fan of releasing every PR's every commit as image. It really pollutes the Docker Hub. Merging on main is fine, but do we really need these for every PR commit? Especially since this will fail the PR build when a user uses a fork (which would be the default for most users).

If you just want the built image available as part of the PR, then you could use the upload-action so it would be available as download.

rzvoncek: I agree that poluting Dcker Hub is no good. I've got my image out and the related k8ssandra-operator PR is passing. I'm ok with dropping this entirely (pushed a commit doing so)

src/server/src/main/java/io/cassandrareaper/management/http/HttpManagementConnectionFactory.java
```diff
@@ -95,13 +100,18 @@ public HttpManagementConnectionFactory(AppContext context, ScheduledExecutorServ
...
+
+    String ts = context.config.getHttpManagement().getTruststore();
+    boolean watchTruststore = ts != null && !ts.isEmpty();
+    String ks = context.config.getHttpManagement().getKeystore();
+    boolean watchKeystore = ks != null && !ks.isEmpty();
+    String tsd = context.config.getHttpManagement().getTruststoresDir();
+    boolean watchTruststoreDir = tsd != null && !tsd.isEmpty() && Files.isDirectory(Paths.get(tsd));
+
+    try {
+      createSslWatcher(watchTruststore, watchKeystore, watchTruststoreDir);
```
burmanm: issue: There's no if here to prevent creating ssl watcher. We would still always create the thread and use Filewatchers, just we wouldn't actually watch any events. If all of these are false, then we shouldn't call createSslWatcher() at all.

burmanm (changes requested):

burmanm (approved):

rzvoncek:
Squashed locally, merging.

PR #1533, opened 2024-12-12
Memory store leader election - REAP-2

### LINKED ISSUE #1519: The memory store allows multiple segments to run per node for the same repair
[Project board link](https://github.com/orgs/k8ssandra/projects/8/views/1?pane=issue&itemId=76126870)

It looks like the memory store doesn't honor the guarantee of scheduling a single segment per node per repair.
Instead, it's the maxParallelRepair concurrency number that seem to be applied as two repairs seem run at once.

At any time, one replica should run at most one segment per allowed repair, and never multiple segments for the same repair.
In the Cassandra storage implementation, we use LWTs to guarantee that.

### Definition of Done
* [ ] The memory storage implementation should allow no more than one segment per node to be scheduled for a given repair run

┆Issue is synchronized with this [Jira Story](https://datastax.jira.com/browse/REAP-2) by [Unito](https://www.unito.io)
┆Issue Number: REAP-2

adejanovski:
Fixes #1519 

The memory storage backend didn't implement any if the `IDistributedStorage` leader election methods.
This was ok until we started using leader election for segment scheduling, which doesn't require to run in a distributed mode.
As a consequence, Reaper using the mem store would schedule one new segment at each poll, even if the replicas were already busy processing another segment.

This PR introduces a class which manages locks on replicas for segments and moves the required methods from `IDistributedStorage` to `IStorage` so they can be implemented in the memory storage implementation.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/storage/IStorageDao.java
+++ b/src/server/src/main/java/io/cassandrareaper/storage/IStorageDao.java
@@ -26,6 +26,9 @@ import io.cassandrareaper.storage.repairsegment.IRepairSegmentDao;
 import io.cassandrareaper.storage.repairunit.IRepairUnitDao;
 import io.cassandrareaper.storage.snapshot.ISnapshotDao;
 
+import java.util.Set;
+import java.util.UUID;
+
 import io.dropwizard.lifecycle.Managed;
 
 /**
@@ -33,6 +36,23 @@ import io.dropwizard.lifecycle.Managed;
  */
 public interface IStorageDao extends Managed,
     IMetricsDao {
+
+  boolean lockRunningRepairsForNodes(
+      UUID repairId,
+      UUID segmentId,
+      Set<String> replicas);
+
+  boolean renewRunningRepairsForNodes(
+      UUID repairId,
+      UUID segmentId,
+      Set<String> replicas);
+
+  boolean releaseRunningRepairsForNodes(
+      UUID repairId,
+      UUID segmentId,
+      Set<String> replicas);
+
+  Set<UUID> getLockedSegmentsForRun(UUID runId);
 
   boolean isStorageConnected();
 
--- a/src/server/src/main/java/io/cassandrareaper/storage/MemoryStorageFacade.java
+++ b/src/server/src/main/java/io/cassandrareaper/storage/MemoryStorageFacade.java
@@ -63,9 +63,8 @@ import org.slf4j.LoggerFactory;
  */
 public final class MemoryStorageFacade implements IStorageDao {
 
-  private static final long LEAD_TIME = 90;
+  private static final long DEFAULT_LEAD_TIME = 90;
   private static final Logger LOG = LoggerFactory.getLogger(MemoryStorageFacade.class);
-  private static final ReplicaLockManagerWithTtl REPAIR_RUN_LOCK_MANAGER = new ReplicaLockManagerWithTtl(LEAD_TIME);
   /** Field evaluator to find transient attributes. This is needed to deal with persisting Guava collections objects
    * that sometimes use the transient keyword for some of their implementation's backing stores**/
   private static final PersistenceFieldEvaluator TRANSIENT_FIELD_EVALUATOR =
@@ -89,8 +88,9 @@ public final class MemoryStorageFacade implements IStorageDao {
   );
   private final MemorySnapshotDao memSnapshotDao = new MemorySnapshotDao();
   private final MemoryMetricsDao memMetricsDao = new MemoryMetricsDao();
+  private final ReplicaLockManagerWithTtl repairRunLockManager;
 
-  public MemoryStorageFacade(String persistenceStoragePath) {
+  public MemoryStorageFacade(String persistenceStoragePath, long leadTime) {
     LOG.info("Using memory storage backend. Persistence storage path: {}", persistenceStoragePath);
     this.embeddedStorage = EmbeddedStorage.Foundation(Paths.get(persistenceStoragePath))
         .onConnectionFoundation(
@@ -107,10 +107,19 @@ public final class MemoryStorageFacade implements IStorageDao {
       LOG.info("Loading existing data from persistence storage");
       this.memoryStorageRoot = (MemoryStorageRoot) this.embeddedStorage.root();
     }
+    this.repairRunLockManager = new ReplicaLockManagerWithTtl(leadTime);
   }
 
   public MemoryStorageFacade() {
-    this("/tmp/" + UUID.randomUUID().toString());
+    this("/tmp/" + UUID.randomUUID().toString(), DEFAULT_LEAD_TIME);
+  }
+
+  public MemoryStorageFacade(String persistenceStoragePath) {
+    this(persistenceStoragePath, DEFAULT_LEAD_TIME);
+  }
+
+  public MemoryStorageFacade(long leadTime) {
+    this("/tmp/" + UUID.randomUUID().toString(), leadTime);
   }
 
   @Override
@@ -303,21 +312,21 @@ public final class MemoryStorageFacade implements IStorageDao {
 
   @Override
   public boolean lockRunningRepairsForNodes(UUID runId, UUID segmentId, Set<String> replicas) {
-    return REPAIR_RUN_LOCK_MANAGER.lockRunningRepairsForNodes(runId, segmentId, replicas);
+    return repairRunLockManager.lockRunningRepairsForNodes(runId, segmentId, replicas);
   }
 
   @Override
   public boolean renewRunningRepairsForNodes(UUID runId, UUID segmentId, Set<String> replicas) {
-    return REPAIR_RUN_LOCK_MANAGER.renewRunningRepairsForNodes(runId, segmentId, replicas);
+    return repairRunLockManager.renewRunningRepairsForNodes(runId, segmentId, replicas);
   }
 
   @Override
   public boolean releaseRunningRepairsForNodes(UUID runId, UUID segmentId, Set<String> replicas) {
-    return REPAIR_RUN_LOCK_MANAGER.releaseRunningRepairsForNodes(runId, segmentId, replicas);
+    return repairRunLockManager.releaseRunningRepairsForNodes(runId, segmentId, replicas);
   }
 
   @Override
   public Set<UUID> getLockedSegmentsForRun(UUID runId) {
-    return REPAIR_RUN_LOCK_MANAGER.getLockedSegmentsForRun(runId);
+    return repairRunLockManager.getLockedSegmentsForRun(runId);
   }
 }
--- a/src/server/src/main/java/io/cassandrareaper/storage/MemoryStorageFacade.java
+++ b/src/server/src/main/java/io/cassandrareaper/storage/MemoryStorageFacade.java
@@ -29,6 +29,7 @@ import io.cassandrareaper.storage.cluster.MemoryClusterDao;
 import io.cassandrareaper.storage.events.IEventsDao;
 import io.cassandrareaper.storage.events.MemoryEventsDao;
 import io.cassandrareaper.storage.memory.MemoryStorageRoot;
+import io.cassandrareaper.storage.memory.ReplicaLockManagerWithTtl;
 import io.cassandrareaper.storage.metrics.MemoryMetricsDao;
 import io.cassandrareaper.storage.repairrun.IRepairRunDao;
 import io.cassandrareaper.storage.repairrun.MemoryRepairRunDao;
@@ -46,6 +47,7 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.UUID;
 import java.util.stream.Collectors;
 
@@ -61,12 +63,14 @@ import org.slf4j.LoggerFactory;
  */
 public final class MemoryStorageFacade implements IStorageDao {
 
+  private static final long LEAD_TIME = 90;
   private static final Logger LOG = LoggerFactory.getLogger(MemoryStorageFacade.class);
-
+  private static final ReplicaLockManagerWithTtl REPAIR_RUN_LOCK_MANAGER = new ReplicaLockManagerWithTtl(LEAD_TIME);
   /** Field evaluator to find transient attributes. This is needed to deal with persisting Guava collections objects
    * that sometimes use the transient keyword for some of their implementation's backing stores**/
   private static final PersistenceFieldEvaluator TRANSIENT_FIELD_EVALUATOR =
       (clazz, field) -> !field.getName().startsWith("_");
+  private static final UUID REAPER_INSTANCE_ID = UUID.randomUUID();
 
   private final EmbeddedStorageManager embeddedStorage;
   private final MemoryStorageRoot memoryStorageRoot;
@@ -295,5 +299,25 @@ public final class MemoryStorageFacade implements IStorageDao {
   // RepairSubscription operations
   public Map<UUID, DiagEventSubscription> getSubscriptionsById() {
     return this.memoryStorageRoot.getSubscriptionsById();
+  }
+
+  @Override
+  public boolean lockRunningRepairsForNodes(UUID runId, UUID segmentId, Set<String> replicas) {
+    return REPAIR_RUN_LOCK_MANAGER.lockRunningRepairsForNodes(runId, segmentId, replicas);
+  }
+
+  @Override
+  public boolean renewRunningRepairsForNodes(UUID runId, UUID segmentId, Set<String> replicas) {
+    return REPAIR_RUN_LOCK_MANAGER.renewRunningRepairsForNodes(runId, segmentId, replicas);
+  }
+
+  @Override
+  public boolean releaseRunningRepairsForNodes(UUID runId, UUID segmentId, Set<String> replicas) {
+    return REPAIR_RUN_LOCK_MANAGER.releaseRunningRepairsForNodes(runId, segmentId, replicas);
+  }
+
+  @Override
+  public Set<UUID> getLockedSegmentsForRun(UUID runId) {
+    return REPAIR_RUN_LOCK_MANAGER.getLockedSegmentsForRun(runId);
   }
 }
--- a/src/server/src/main/java/io/cassandrareaper/storage/MemoryStorageFacade.java
+++ b/src/server/src/main/java/io/cassandrareaper/storage/MemoryStorageFacade.java
@@ -51,6 +51,7 @@ import java.util.Set;
 import java.util.UUID;
 import java.util.stream.Collectors;
 
+import com.google.common.io.Files;
 import org.eclipse.serializer.persistence.types.PersistenceFieldEvaluator;
 import org.eclipse.store.storage.embedded.types.EmbeddedStorage;
 import org.eclipse.store.storage.embedded.types.EmbeddedStorageManager;
@@ -111,7 +112,7 @@ public final class MemoryStorageFacade implements IStorageDao {
   }
 
   public MemoryStorageFacade() {
-    this("/tmp/" + UUID.randomUUID().toString(), DEFAULT_LEAD_TIME);
+    this(Files.createTempDir().getAbsolutePath(), DEFAULT_LEAD_TIME);
   }
 
   public MemoryStorageFacade(String persistenceStoragePath) {
@@ -119,7 +120,7 @@ public final class MemoryStorageFacade implements IStorageDao {
   }
 
   public MemoryStorageFacade(long leadTime) {
-    this("/tmp/" + UUID.randomUUID().toString(), leadTime);
+    this(Files.createTempDir().getAbsolutePath(), leadTime);
   }
 
   @Override
--- a/src/server/src/main/java/io/cassandrareaper/storage/MemoryStorageFacade.java
+++ b/src/server/src/main/java/io/cassandrareaper/storage/MemoryStorageFacade.java
@@ -64,13 +64,13 @@ import org.slf4j.LoggerFactory;
  */
 public final class MemoryStorageFacade implements IStorageDao {
 
-  private static final long DEFAULT_LEAD_TIME = 90;
+  // Default time to live of leads taken on a segment
+  private static final long DEFAULT_LEAD_TTL = 90_000;
   private static final Logger LOG = LoggerFactory.getLogger(MemoryStorageFacade.class);
   /** Field evaluator to find transient attributes. This is needed to deal with persisting Guava collections objects
    * that sometimes use the transient keyword for some of their implementation's backing stores**/
   private static final PersistenceFieldEvaluator TRANSIENT_FIELD_EVALUATOR =
       (clazz, field) -> !field.getName().startsWith("_");
-  private static final UUID REAPER_INSTANCE_ID = UUID.randomUUID();
 
   private final EmbeddedStorageManager embeddedStorage;
   private final MemoryStorageRoot memoryStorageRoot;
@@ -89,7 +89,7 @@ public final class MemoryStorageFacade implements IStorageDao {
   );
   private final MemorySnapshotDao memSnapshotDao = new MemorySnapshotDao();
   private final MemoryMetricsDao memMetricsDao = new MemoryMetricsDao();
-  private final ReplicaLockManagerWithTtl repairRunLockManager;
+  private final ReplicaLockManagerWithTtl replicaLockManagerWithTtl;
 
   public MemoryStorageFacade(String persistenceStoragePath, long leadTime) {
     LOG.info("Using memory storage backend. Persistence storage path: {}", persistenceStoragePath);
@@ -108,15 +108,15 @@ public final class MemoryStorageFacade implements IStorageDao {
       LOG.info("Loading existing data from persistence storage");
       this.memoryStorageRoot = (MemoryStorageRoot) this.embeddedStorage.root();
     }
-    this.repairRunLockManager = new ReplicaLockManagerWithTtl(leadTime);
+    this.replicaLockManagerWithTtl = new ReplicaLockManagerWithTtl(leadTime);
   }
 
   public MemoryStorageFacade() {
-    this(Files.createTempDir().getAbsolutePath(), DEFAULT_LEAD_TIME);
+    this(Files.createTempDir().getAbsolutePath(), DEFAULT_LEAD_TTL);
   }
 
   public MemoryStorageFacade(String persistenceStoragePath) {
-    this(persistenceStoragePath, DEFAULT_LEAD_TIME);
+    this(persistenceStoragePath, DEFAULT_LEAD_TTL);
   }
 
   public MemoryStorageFacade(long leadTime) {
@@ -313,22 +313,22 @@ public final class MemoryStorageFacade implements IStorageDao {
 
   @Override
   public boolean lockRunningRepairsForNodes(UUID runId, UUID segmentId, Set<String> replicas) {
-    return repairRunLockManager.lockRunningRepairsForNodes(runId, segmentId, replicas);
+    return replicaLockManagerWithTtl.lockRunningRepairsForNodes(runId, segmentId, replicas);
   }
 
   @Override
   public boolean renewRunningRepairsForNodes(UUID runId, UUID segmentId, Set<String> replicas) {
-    return repairRunLockManager.renewRunningRepairsForNodes(runId, segmentId, replicas);
+    return replicaLockManagerWithTtl.renewRunningRepairsForNodes(runId, segmentId, replicas);
   }
 
   @Override
   public boolean releaseRunningRepairsForNodes(UUID runId, UUID segmentId, Set<String> replicas) {
     LOG.info("Releasing locks for runId: {}, segmentId: {}, replicas: {}", runId, segmentId, replicas);
-    return repairRunLockManager.releaseRunningRepairsForNodes(runId, segmentId, replicas);
+    return replicaLockManagerWithTtl.releaseRunningRepairsForNodes(runId, segmentId, replicas);
   }
 
   @Override
   public Set<UUID> getLockedSegmentsForRun(UUID runId) {
-    return repairRunLockManager.getLockedSegmentsForRun(runId);
+    return replicaLockManagerWithTtl.getLockedSegmentsForRun(runId);
   }
 }
--- a/src/server/src/main/java/io/cassandrareaper/storage/MemoryStorageFacade.java
+++ b/src/server/src/main/java/io/cassandrareaper/storage/MemoryStorageFacade.java
@@ -322,6 +322,7 @@ public final class MemoryStorageFacade implements IStorageDao {
 
   @Override
   public boolean releaseRunningRepairsForNodes(UUID runId, UUID segmentId, Set<String> replicas) {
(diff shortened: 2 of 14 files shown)


Miles-Garnsey:
@adejanovski this is very tasty but your CI is failing at the moment, and can you remind me how to manually test this :)

src/server/src/main/java/io/cassandrareaper/storage/IStorageDao.java
```diff
@@ -34,6 +37,23 @@
 public interface IStorageDao extends Managed,
     IMetricsDao {
 
+  boolean lockRunningRepairsForNodes(
```
Miles-Garnsey, line 40: Suggestion: given we've been trying to break down the IStorageDao over the past year or two, would it be better to put this stuff into an IConcurrencyDao class? I think that would be more in keeping with the path we're on.

src/server/src/main/java/io/cassandrareaper/storage/MemoryStorageFacade.java
```diff
@@ -85,8 +89,9 @@ public final class MemoryStorageFacade implements IStorageDao {
   );
   private final MemorySnapshotDao memSnapshotDao = new MemorySnapshotDao();
   private final MemoryMetricsDao memMetricsDao = new MemoryMetricsDao();
+  private final ReplicaLockManagerWithTtl repairRunLockManager;
```
Miles-Garnsey: Nit: this naming is a bit incconsistent, I think it would be clearer if the variable was called `replicaLockManagerWithTtl`

src/server/src/main/java/io/cassandrareaper/storage/memory/ReplicaLockManagerWithTtl.java
```diff
@@ -0,0 +1,169 @@
...
+  private String getReplicaLockKey(String replica, UUID runId) {
+    return replica + runId;
+  }
+
+  public boolean lockRunningRepairsForNodes(UUID runId, UUID segmentId, Set<String> replicas) {
+    lock.lock();
+    try {
+      long currentTime = System.currentTimeMillis();
+      // Check if any replica is already locked by another runId
+      for (String replica : replicas) {
```
Miles-Garnsey: Nit: I think the pattern in this codebase is to the use the streams API instead of for loops where possible?

src/server/src/main/java/io/cassandrareaper/storage/memory/ReplicaLockManagerWithTtl.java
```diff
@@ -0,0 +1,169 @@
...
+        LockInfo lockInfo = replicaLocks.get(getReplicaLockKey(replica, runId));
+        if (lockInfo != null && lockInfo.expirationTime > currentTime && lockInfo.runId.equals(runId)) {
+          return false; // Replica is locked by another runId and not expired
+        }
+      }
+
+      // Lock the replicas for the given runId and segmentId
+      long expirationTime = currentTime + (ttlSeconds * 1000);
```
Miles-Garnsey: Question: why do we multiply ttlSeconds by 1000?

src/server/src/main/java/io/cassandrareaper/storage/memory/ReplicaLockManagerWithTtl.java
```diff
@@ -0,0 +1,169 @@
...
+    // Schedule cleanup of expired locks
+    scheduler.scheduleAtFixedRate(this::cleanupExpiredLocks, 1, 1, TimeUnit.SECONDS);
+  }
+
+  private String getReplicaLockKey(String replica, UUID runId) {
+    return replica + runId;
+  }
+
+  public boolean lockRunningRepairsForNodes(UUID runId, UUID segmentId, Set<String> replicas) {
+    lock.lock();
```
Miles-Garnsey, line 51: Just so I'm clear on the language, I presume that there's some N sized threadpool which is running N `RepairRuns`, and those repairs are attempting to repair `Segments` on specific `Replicas`. The intention here is to limit repair cadence so that only one `RepairRun` can run on one `Replica` at any given time?

src/server/src/main/java/io/cassandrareaper/storage/memory/ReplicaLockManagerWithTtl.java
```diff
@@ -0,0 +1,169 @@
...
+import java.util.concurrent.locks.ReentrantLock;
+
+import com.google.common.annotations.VisibleForTesting;
+
+public class ReplicaLockManagerWithTtl {
+
+  private final ConcurrentHashMap<String, LockInfo> replicaLocks = new ConcurrentHashMap<>();
+  private final ConcurrentHashMap<UUID, Set<UUID>> runToSegmentLocks = new ConcurrentHashMap<>();
+  private final Lock lock = new ReentrantLock();
+  private final ScheduledExecutorService scheduler = Executors.newScheduledThreadPool(1);
```
Miles-Garnsey: Explanatory note to self: the ttl cleanup logic runs in a single threaded executor which runs once every second and cleans up those locks whose TTL has expired.

src/server/src/main/java/io/cassandrareaper/storage/memory/ReplicaLockManagerWithTtl.java
```diff
@@ -0,0 +1,169 @@
...
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+
+import com.google.common.annotations.VisibleForTesting;
+
+public class ReplicaLockManagerWithTtl {
+
+  private final ConcurrentHashMap<String, LockInfo> replicaLocks = new ConcurrentHashMap<>();
+  private final ConcurrentHashMap<UUID, Set<UUID>> runToSegmentLocks = new ConcurrentHashMap<>();
+  private final Lock lock = new ReentrantLock();
```
Miles-Garnsey, line 35: Issue: I have a minor concern on the use of this lock because it seems that it might suffer a fairly high degree of contention in larger Reaper setups and won't exhibit good scaling behaviour. 

A few suggestions to resolve this better:
1. It looks like the data structures protected by this mutex are concurrency safe to start with - what is the reason in terms of consistency between or within the two data structures (`replicaLocks`, `runToSegmentLocks`) which necessitates this lock? What behaviour are we protecting from? Can we re-design the logic to be more concurrency safe and lock-free?
2. This lock appears to be global across the entire application. This will result in uneccesary contention in Reaper instances managing large numbers of clusters and replicas. AFAICT, the lock could apply to individual clusters, or (even better) individual replicas within a specific cluster. This would reduce contention significantly.
3. Given all of the above, I am concerned about fairness in scheduling for the individual functions relying on acquiring this lock. 
4. I am also concerned about retry storms given the lack of jitter and backoff in the callers, this is another area where improvement could be had.

src/server/src/main/java/io/cassandrareaper/storage/memory/ReplicaLockManagerWithTtl.java
```diff
@@ -0,0 +1,169 @@
...
+    this.ttlSeconds = ttlSeconds;
+    // Schedule cleanup of expired locks
+    scheduler.scheduleAtFixedRate(this::cleanupExpiredLocks, 1, 1, TimeUnit.SECONDS);
+  }
+
+  private String getReplicaLockKey(String replica, UUID runId) {
+    return replica + runId;
+  }
+
+  public boolean lockRunningRepairsForNodes(UUID runId, UUID segmentId, Set<String> replicas) {
```
Miles-Garnsey, line 50: Nit: runiD - is this repairRunID? Request we be clear on the language since there are a lot of different concepts here for folks like me who don't work on this codebase too often.

src/server/src/main/java/io/cassandrareaper/storage/memory/ReplicaLockManagerWithTtl.java
```diff
@@ -0,0 +1,169 @@
...
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+
+import com.google.common.annotations.VisibleForTesting;
+
+public class ReplicaLockManagerWithTtl {
+
+  private final ConcurrentHashMap<String, LockInfo> replicaLocks = new ConcurrentHashMap<>();
```
Miles-Garnsey, line 33: Nit: more naming. If a `Run` is a `RepairRun` can we please spell that out for clarity.

src/server/src/main/java/io/cassandrareaper/storage/memory/ReplicaLockManagerWithTtl.java
Miles-Garnsey: Nit: more naming - is the intention of this method to be more like `addRepairRunLockForReplica`? Consistent language would be a big help.

src/server/src/main/java/io/cassandrareaper/storage/memory/ReplicaLockManagerWithTtl.java
```diff
@@ -0,0 +1,169 @@
...
+    }
+  }
+
+  public boolean releaseRunningRepairsForNodes(UUID runId, UUID segmentId, Set<String> replicas) {
+    lock.lock();
+    try {
+      // Remove the lock for replicas
+      for (String replica : replicas) {
+        LockInfo lockInfo = replicaLocks.get(getReplicaLockKey(replica, runId));
+        if (lockInfo != null && lockInfo.runId.equals(runId)) {
```
Miles-Garnsey: Suggestion: the only reason I can see to use a lock here is to maintain consistency between `replicaLocks` and `runToSegmentLocks`. If that's the case, you don't need to lock until you've actually found the lock (i.e. after `replicaLocks.get`). If the lock is found, then at that point you can call `lock.lock();` to maintain consistency between the two maps. 

I think similar logic might apply to all of these methods. Locking at entry to the method is probably overkill.

src/server/src/test/java/io/cassandrareaper/service/RepairRunnerHangingTest.java
```diff
@@ -242,7 +243,7 @@ public void testHangingRepair() throws InterruptedException, ReaperException, JM
     final double intensity = 0.5f;
     final int repairThreadCount = 1;
     final int segmentTimeout = 1;
-    final IStorageDao storage = new MemoryStorageFacade();
+    final IStorageDao storage = new MemoryStorageFacade(LEAD_TIME);
```
Miles-Garnsey: Issue: Can we please replace `LEAD_TIME` with ttl or something that describes what it does. This was a bit confusing.

Miles-Garnsey (changes requested):
I've put a bunch of feedback as a first review. I'll likely want to consider this more - this isn't an easy PR so nice work on getting something together. It feels like its very close.

I have a bunch of comments around naming since I had a bit of a tricky time understanding some of what was going on at first. These are always nits, but given the functionality is complex I'd love to see some of them straightened out for future readers.

I also think we could handle the locking we're doing much better, and suggest we revisit that now given some of the use cases we want to put this into. For a first cut this is great, but I think a bit of improvement there might do us a world of good. Improvements might include:

1. Better segmentation of what needs to be locked, since the global lock you're using isn't scalable.
2. Being more discerning about where you're locking. You probably only need to lock on writes, but you're often acquiring the lock before reads at entry into the method.

Again, this is kind of a minor tweak but I think it'll avoid headaches down the road.

src/server/src/main/java/io/cassandrareaper/storage/MemoryStorageFacade.java
adejanovski: done

src/server/src/main/java/io/cassandrareaper/storage/memory/ReplicaLockManagerWithTtl.java
adejanovski: to convert it to milliseconds? :)

src/server/src/main/java/io/cassandrareaper/storage/IStorageDao.java
adejanovski: My personal opinion would be that it's probably too much breaking up and that could end up making the code less easy to grasp. Since all storage backends need to implement these methods, it might be better to keep them in the IStorageDao.
I appreciate the design concern behind your comment though.

src/server/src/main/java/io/cassandrareaper/storage/memory/ReplicaLockManagerWithTtl.java
adejanovski: A run or runId always refer to a repairRun across the codebase (while we have a SegmentRunner class, we don't have the notion of segmentRun).
I'll rename in this class to address your comment nonetheless.

src/server/src/main/java/io/cassandrareaper/storage/memory/ReplicaLockManagerWithTtl.java
adejanovski: > It looks like the data structures protected by this mutex are concurrency safe to start with - what is the reason in terms of consistency between or within the two data structures (replicaLocks, runToSegmentLocks) which necessitates this lock? What behaviour are we protecting from? Can we re-design the logic to be more concurrency safe and lock-free?

While the data structures we use are thread safe, we have to place multiple locks (one per replica) by first checking if the locks exist and then place the lock if all replicas are free. That's not going to work if we allow multiple threads to try locking replicas concurrently. Between the moment when they check if replicas are locked and the moment they lock these replicas, there could be a race condition.

> This lock appears to be global across the entire application. This will result in unnecessary contention in Reaper instances managing large numbers of clusters and replicas. AFAICT, the lock could apply to individual clusters, or (even better) individual replicas within a specific cluster. This would reduce contention significantly.

I hear that, but keep in mind that we're trying to schedule segments every 30 seconds by default (which can be lowered to 1s at a minimum), and there's no networking involved in these methods. The execution should be quite fast then and I don't think this contention is really an issue.
Using a more specific lock could introduce some complexity that would be a premature optimization.
I briefly tried to introduce a repair run specific lock, but managing the cleanup of these locks so they don't pile up indefinitely turned out to create more complexity than needed.

> Given all of the above, I am concerned about fairness in scheduling for the individual functions relying on acquiring this lock.
I am also concerned about retry storms given the lack of jitter and backoff in the callers, this is another area where improvement could be had.

We're talking about 2 allowed concurrent repair runners per cluster, which can only try once at a time to get such a lock every 30 seconds.
I understand the concern, but I'm not seeing evidence of this in my testing.
I do see flakes, but can't prove yet if they're due to the presence of a global lock. Some of the integration tests which are failing use the cassandra storage, which seems to point elsewhere.
If I manage to get some evidence that the global lock is causing issues, I'll consider some additional granularity.

src/server/src/main/java/io/cassandrareaper/storage/memory/ReplicaLockManagerWithTtl.java
adejanovski: yeah, I'll change the name, that's not very explanatory 😅

src/server/src/main/java/io/cassandrareaper/storage/memory/ReplicaLockManagerWithTtl.java
adejanovski: I'm using the same method names as the ones from the IStorage interface. At least that brings some consistency.
We can revisit this in a refactoring PR.

src/server/src/main/java/io/cassandrareaper/storage/memory/ReplicaLockManagerWithTtl.java
adejanovski: I could rephrase this to: for a given `RepairRun`, two `Segments` should never involve the same `Replica` concurrently.

You can have multiple repair runs running (rate limiting for repair runs happens elsewhere) in a way that segments from different repair runs can involve the same replicas concurrently.

src/server/src/main/java/io/cassandrareaper/storage/memory/ReplicaLockManagerWithTtl.java
adejanovski: fair point, I'm refactoring this to use streams.

src/server/src/main/java/io/cassandrareaper/storage/memory/ReplicaLockManagerWithTtl.java
adejanovski: I disagree, since we're trying to lock multiple replicas at the same time, we need to protect the code from race conditions between checking the locks and taking the locks.
The alternative would be to check without a lock and then risk a failure in placing the replica lock (and deal with the exception).
Given that everything in the method happens in memory, it's going to be a quick lock.

src/server/src/test/java/io/cassandrareaper/service/RepairRunnerHangingTest.java
adejanovski: sure, I'll switch to LEAD_TTL everywhere.

src/server/src/main/java/io/cassandrareaper/storage/memory/ReplicaLockManagerWithTtl.java
Miles-Garnsey: I hear you on the premature optimisation angle so I'm happy to run with what you have here if you feel it's best. I do note that the cleanups are happening every second, which is where most of my concern comes from.

src/server/src/main/java/io/cassandrareaper/storage/memory/ReplicaLockManagerWithTtl.java
Miles-Garnsey: In the event of failure to acquire the lock, isn't it just a retry anyway? Again, I'm not committed to this, it is just a suggestion.

src/server/src/main/java/io/cassandrareaper/storage/IStorageDao.java
Miles-Garnsey: OK, only a suggestion, resolving.

src/server/src/main/java/io/cassandrareaper/storage/memory/ReplicaLockManagerWithTtl.java
Miles-Garnsey: I think you might have missed this one in your latest commit?

src/server/src/main/java/io/cassandrareaper/storage/memory/ReplicaLockManagerWithTtl.java
Miles-Garnsey: I kind of got that... I think the point I was trying to make (badly) was: Can I suggest we either use proper time time-based types here instead of longs (ideal) and drop the unit from all variable names, or that we include a unit in all variable names, AND that we use the same units for all variables so we aren't doing conversions like this?

src/server/src/main/java/io/cassandrareaper/storage/memory/ReplicaLockManagerWithTtl.java
Miles-Garnsey: Looks like the name there is still the same but this comment is marked as outdated - I can't see what changed? :|

src/server/src/main/java/io/cassandrareaper/storage/memory/ReplicaLockManagerWithTtl.java
Miles-Garnsey: > Between the moment when they check if replicas are locked and the moment they lock these replicas, there could be a race condition.

Roger, I've been mulling this over and I can see what you're saying, although I still think you could make better uses of the concurrent map to avoid locking so early using the external lock. 

I think I'm getting a bit confused between the `Lock` in the Java sense in one place and talking about our own `Locks` which exist in a map. In the first case, it is an atomic CAS operation but in the latter it is a read then a write from the Java side. I can see that in that latter case you might end up with an overwrite using your current logic. 

That said, I think a revision could be made to use a CAS type operation just using the features of the `ConcurrentHashMap`. For example, `putIfAbsent` would probably enable an atomic read then write operation without recourse to the external lock.

I suspect (although I have not worked it through entirely, so I could be wrong) that you could do this to avoid using an external lock entirely, but I'm open to being wrong there and understand if you want to move on without exploring that.

src/server/src/main/java/io/cassandrareaper/storage/memory/ReplicaLockManagerWithTtl.java
Miles-Garnsey: This is only six clicks of the mouse and the typing of three words I'm pretty sure - but it is also just a suggestion ;)

src/server/src/main/java/io/cassandrareaper/storage/memory/ReplicaLockManagerWithTtl.java
Miles-Garnsey: Let me try to provide an example later on today. This requires some thinking through.

src/server/src/main/java/io/cassandrareaper/storage/memory/ReplicaLockManagerWithTtl.java
```diff
@@ -0,0 +1,179 @@
...
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+
+import com.google.common.annotations.VisibleForTesting;
+
+public class ReplicaLockManagerWithTtl {
+
+  private final ConcurrentHashMap<String, LockInfo> replicaLocks = new ConcurrentHashMap<>();
+  private final ConcurrentHashMap<UUID, Set<UUID>> repairRunToSegmentLocks = new ConcurrentHashMap<>();
+  private final ConcurrentHashMap<UUID, ReentrantLock> runIdLocks = new ConcurrentHashMap<>();
```
Miles-Garnsey: Issue: I thought you said you weren't going to divide the locks up by runID? 😅

src/server/src/main/java/io/cassandrareaper/storage/memory/ReplicaLockManagerWithTtl.java
```diff
@@ -0,0 +1,179 @@
...
+  private Lock getLockForRunId(UUID runId) {
+    return runIdLocks.computeIfAbsent(runId, k -> new ReentrantLock());
+  }
+
+  public boolean lockRunningRepairsForNodes(UUID runId, UUID segmentId, Set<String> replicas) {
+    Lock lock = getLockForRunId(runId);
+    lock.lock();
+    try {
+      long currentTime = System.currentTimeMillis();
+      // Check if any replica is already locked by another runId
```
Miles-Garnsey, line 54: Issue: This is confusing. I think this is checking if each replica is locked and then taking action per replica. It isn't checking if ANY replica is locked.

Miles-Garnsey (changes requested):

src/server/src/main/java/io/cassandrareaper/storage/memory/ReplicaLockManagerWithTtl.java
```diff
@@ -0,0 +1,179 @@
...
+      boolean anyReplicaLocked = replicas.stream()
+          .map(replica -> replicaLocks.get(getReplicaLockKey(replica, runId)))
+          .anyMatch(lockInfo -> lockInfo != null
+            && lockInfo.expirationTime > currentTime && lockInfo.runId.equals(runId));
+
```
Miles-Garnsey, line 59: Issue: This looks wrong, you're saying to return `false` only if the value of this entry has `runId` equal to the `runId` you're trying to lock. Don't you want to return `false` regardless of who holds the lock?

Miles-Garnsey:
I'm starting to struggle with the length of this review, so let's propose a sketch to do just one method without external locks:

```
  public boolean lockRunningRepairsForNodes(UUID runId, UUID segmentId, Set<String> replicas) {
      // For each replica, check if a lock is held by another RepairRun.
      for (String replica : replicas) {
        replicaLocks.compute(getReplicaLockKey(replica, runId), (k,lockInfo) -> {
          long currentTime = System.currentTimeMillis();
          if (lockInfo != null && lockInfo.expirationTime > currentTime && lockInfo.runId.equals(runId)) { // lockInfo.runId.equals(runId) looks wrong to me?
            return lockInfo; // Replica is locked by another runId and not expired, return existing value.
          }
          long expirationTime = currentTime + (ttlSeconds * 1000);
          // Lock the replicas for the given runId and segmentId
          LockInfo newLockInfo = new LockInfo(runId, expirationTime);
          newLockInfo.segmentLocks.add(segmentId);
          return newLockInfo;
        });
      }
      return true;
  }
```

This also requires an update:

```
private static class LockInfo {
    UUID runId;
    long expirationTime;
    Set<UUID> segmentLocks;
    LockInfo(UUID runId, long expirationTime) {
      this.runId = runId;
      this.expirationTime = expirationTime;
      this.segmentLocks = new HashSet<>();
    }
  }
```

What this does:

1. Uses the ConcurrentHashMaps internal locking (which I think should be per key) instead of a global lock. 
2. Moves the `segmentLocks` into  `LockInfo` which makes clear that they exist within a given `RunID` (I hope I've understood this correctly, otherwise, I'll have broken things).
3. Eliminates all external locking requirements. This will just iterate through the replicaLocks independently and update them if possible.  It might be good to have a logic branch here for the case `lockInfo.expirationTime < currentTime` so that this can be removed without waiting for the TTL cleanup to happen, but maybe that can be handled with a simple retry after the other threadpool cleans the entry up.

src/server/src/main/java/io/cassandrareaper/storage/memory/ReplicaLockManagerWithTtl.java
adejanovski: yeah, I tried but then rolled back. Not far enough it seems :)

Miles-Garnsey:
From discussions with Alex earlier: 

1. The proposal I have above where we move `Set<UUID> segmentLocks` inside `LockInfo` won't work because segments are not unique to replicas. 
2. This means that the same segment can be running on multiple replicas (in fact, in RF 3 it will be running on all three), and all replicas involved in a particular segment need to be locked.
3. We discussed the possibility of keying a map by segment with replica ID as the value instead, but Alex pointed out that even then, you may be able to lock one replica, but if you fail to lock all three then you need to roll back any locked ones before proceeding. 
4. Because the Hashmap's mutex can't be used to lock all three replicas at the same time, the unlocking process on one replica then might also fail, which could lead to deadlocks.
5. Effectively, what we'd need is a way to do a lookup for the three replicas the segment is supposed to run on, then lock all three atomically. I don't think any Map implementation gives us the ability to lock multiple entries at the same time, which is the critical issue here.

I feel like there's some concurrent bimap that might resolve this whole state, but given that I've been reminded that we will currently only be managing one cluster per instance in the proposed use case we have, we can probably defer this more complicated implementation for later.

src/server/src/main/java/io/cassandrareaper/storage/memory/ReplicaLockManagerWithTtl.java
adejanovski: We're good now, no more loop.

src/server/src/main/java/io/cassandrareaper/storage/memory/ReplicaLockManagerWithTtl.java
adejanovski: I moved to milliseconds to get more control over this.
Let's keep any change into a follow up refactoring on this.

src/server/src/main/java/io/cassandrareaper/storage/memory/ReplicaLockManagerWithTtl.java
adejanovski: name was changed to `lockCleanupScheduler`, and it's no longer a class member, but just a variable in the constructor.

src/server/src/main/java/io/cassandrareaper/storage/memory/ReplicaLockManagerWithTtl.java
adejanovski: As discussed we won't have much contention anyway, everything happens in memory and we have a safe implementation for setting our distributed lock the way it's implemented here.
It'll work fine even with many competing threads (which we won't have).

src/server/src/main/java/io/cassandrareaper/storage/memory/ReplicaLockManagerWithTtl.java
adejanovski: I don't follow your reasoning. We won't be able to run the segment if "any" of the replicas is already locked. Not if each of them is locked. That would mean we would be allowed to run if some of them are locked and no others, that's not the case.

Miles-Garnsey: Right, I think this came about from me misunderstanding what you were trying to do here. You basically want to check if a given replica for a specific segment is locked, not if any replica for any segment is locked.

src/server/src/main/java/io/cassandrareaper/storage/memory/ReplicaLockManagerWithTtl.java
Miles-Garnsey: Did you want to roll this back? I thought we were reverting to using a global lock.

src/server/src/main/java/io/cassandrareaper/storage/memory/ReplicaLockManagerWithTtl.java
Miles-Garnsey: Still issues with naming here IMO, but I think we agreed that we wouldn't necessarily fix all of them in this PR.

src/server/src/main/java/io/cassandrareaper/storage/memory/ReplicaLockManagerWithTtl.java
Miles-Garnsey: As discussed elsewhere, the fact that a segment can run on multiple replicas probably we need a more complex data structure than either a one way map or nested maps. We would investigate this later if required.

src/server/src/main/java/io/cassandrareaper/storage/memory/ReplicaLockManagerWithTtl.java
Miles-Garnsey: Looks like there is still a loop here `for (String replica : replicas) {` was this something you wanted to change?

Miles-Garnsey (approved):
I've put more comments as it seems you've said that you've made certain updates in comments but then they aren't all reflected in the code. You may have rolled some back accidentally while rolling back other changes (esp the use of a map of locks).

I'm approving anyway since the tests pass and I don't want you to be blocked from merging pending my review. Whether you make the rest of the changes we've discussed is up to you, since I think they are all non-blocking and this is functionally correct AFAICT.

src/server/src/main/java/io/cassandrareaper/storage/memory/ReplicaLockManagerWithTtl.java
adejanovski: no, we allow concurrency on replicas across repair runs. Its within a repair run that we disallow it.

PR #1535, opened 2024-12-18
Fix bug in replica lock release in the mem store

adejanovski:
There's currently a bug in the release lock code that doesn't actually releases the lock. As a consequence we need to wait for the lock to expire before being able to schedule a new segment.

This PR fixes this by properly cleaning up replica locks.


Miles-Garnsey (approved):
Approving because tests pass - some explanation still going on offline but this is non blocking.

PR #1538, opened 2025-01-28
Build a slim image without Python nor spreaper

adejanovski:
This will limit the number of CVEs in our Docker images.


src/server/src/main/docker/Dockerfile
```diff
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM amazoncorretto:11.0.25-alpine
+FROM amazoncorretto:11.0.26-alpine
```
burmanm: suggestion: Wouldn't it make sense to use 11-alpine instead? That way it would be the newest image everytime we build.

burmanm (approved):

src/server/src/main/docker/Dockerfile
adejanovski: done 👍

PR #1554, opened 2025-03-25
Add clusterName to JMX connection logs

andresbeckruiz:
To enhance observability of JMX connection failures and simplify troubleshooting, this patch introduces a new `clusterName` parameter in `JmxCassandraManagementProxy.connect` so that logs can include the cluster name of the node that Reaper is attempting to connect to. 

Compiled and successfully ran all server tests:

Screenshot 2025-03-25 at 3 43 14 PM


adejanovski (approved):
Looks good, and it should be very helpful for troubleshooting indeed 👍

PR #1556, opened 2025-04-03
Upgrade Dropwizard and dropwizard-cassandra to v3

adejanovski:
This PR upgrades the base version of Dropwizard to v3 which is not EOL.
It also switches the cassandra bundle to the official dropwizard-cassandra one, upgrading the underlying Cassandra java driver to v4 and bringing support for Cassandra 5.


src/server/src/main/docker/Dockerfile
```diff
@@ -58,14 +58,16 @@ ENV REAPER_SEGMENT_COUNT_PER_NODE=64 \
     REAPER_SERVER_ADMIN_BIND_HOST="0.0.0.0" \
     REAPER_CASS_ACTIVATE_QUERY_LOGGER=false \
     REAPER_CASS_CLUSTER_NAME="clustername" \
-    REAPER_CASS_CONTACT_POINTS="[]" \
-    REAPER_CASS_PORT=9042 \
+    REAPER_CASS_CONTACT_POINTS="{\"host\": \"127.0.0.1\", \"port\": \"9042\"}, {\"host\": \"127.0.0.2\", \"port\": \"9042\"}" \
```
burmanm: Why is Dockerfile hardcoding these?

src/server/src/main/java/io/cassandrareaper/service/SchedulingManager.java
```diff
@@ -35,8 +32,11 @@
 import java.util.UUID;
 import java.util.concurrent.ThreadLocalRandom;
 
-import com.datastax.driver.core.exceptions.DriverException;
-import com.datastax.driver.core.exceptions.DriverInternalError;
+import org.joda.time.DateTime;
+import org.slf4j.Logger;
```
burmanm: Double import from L43. Same for LoggerFactory and DateTime.

src/server/src/main/docker/Dockerfile
adejanovski: Right, I think it was meant as an example but I'd need to set it to an equivalent than previously.

src/server/src/main/java/io/cassandrareaper/storage/operations/CassandraOperationsDao.java
```diff
@@ -1,92 +1,81 @@
...
-        return row.getString("data");
+    List<CompletionStage<AsyncResultSet>> futures = Lists.newArrayList();
+    futures.add(session.executeAsync(listOperationsForNodePrepStmt.bind(clusterName,
+        operationType.getName(), DateTime.now().toString(TIME_BUCKET_FORMATTER), host)));
+    futures.add(session
+        .executeAsync(listOperationsForNodePrepStmt.bind(clusterName, operationType.getName(),
+            DateTime.now().minusMinutes(1).toString(TIME_BUCKET_FORMATTER), host)));
+    for (CompletionStage<AsyncResultSet> future : futures) {
+      AsyncResultSet results = future.toCompletableFuture().join();
+      while (true) {
```
burmanm: This confuses me, why do we need while(true) ? Shouldn't CompletableFuture.join() block the execution until the resultSet is available? That's how the CompletableFuture should work.

adejanovski: right, the while loop can probably be removed

adejanovski: actually it's needed it seems with the new paging model of the API.
But in this specific case, we're not going through pages and returning the first row we find, so we should be ok.

src/server/src/main/java/io/cassandrareaper/storage/repairrun/CassandraRepairRunDao.java
```diff
@@ -461,19 +515,24 @@ private Collection<? extends RepairRun> getRepairRunsWithStateForCluster(
...
-      ResultSet repairRunResult = future.getUninterruptibly();
-      for (Row row : repairRunResult) {
-        repairRuns.add(buildRepairRunFromRow(row, row.getUUID("id")));
+    for (CompletionStage<AsyncResultSet> future : futures) {
+      AsyncResultSet results = future.toCompletableFuture().join();
+      while (true) {
+        for (Row row : results.currentPage()) {
+          repairRuns.add(buildRepairRunFromRow(row, row.getUuid("id")));
+        }
+        if (!results.hasMorePages()) {
```
burmanm, line 462: This is just nitpicking, but wouldn't this code be equal to

```java
do {
    for(Row row ...) { .. }
    results = results.fetchNextPage().toCompletableFuture().join();
} while(results.hasMorePages());
```

src/server/src/main/java/io/cassandrareaper/storage/snapshot/CassandraSnapshotDao.java
```diff
@@ -71,11 +74,11 @@ public Snapshot getSnapshot(String clusterName, String snapshotName) {
     Snapshot.Builder snapshotBuilder = Snapshot.builder().withClusterName(clusterName).withName(snapshotName);
 
     ResultSet result = session.execute(getSnapshotPrepStmt.bind(clusterName, snapshotName));
-    for (Row row : result) {
+    for (Row row : result.all() ) {
```
burmanm: Probably sane reason for this, but in some parts you modified the code to do session.executeAsync() and then blocked with join() and here you use the .execute() with .all() for the same end result basically. Why?

src/server/src/main/java/io/cassandrareaper/storage/MemoryStorageFacade.java
```diff
@@ -90,37 +105,48 @@ public final class MemoryStorageFacade implements IStorageDao {
...
+        root = new MemoryStorageRoot();
+        storage.setRoot(root);
+      } else {
+        LOG.info("Loading existing data from persistence storage");
+        root = (MemoryStorageRoot) storage.root();
+      }
     } else {
-      LOG.info("Loading existing data from persistence storage");
-      this.memoryStorageRoot = (MemoryStorageRoot) this.embeddedStorage.root();
+      root = new MemoryStorageRoot();
```
burmanm, line 128: On L127 this explicitly sets the storage.setRoot(root);, but not in this branch? Both are creating a new MemoryStorageRoot() so I don't see where the functional difference comes from.

burmanm: Also, if this path is taken, the storage variable remains null? That would cause NPE in stop().

burmanm: And would leave embeddedStorage as null also.

src/server/src/main/java/io/cassandrareaper/storage/operations/CassandraOperationsDao.java
burmanm: Yeah, that was different from the other examples, that's why it caught my eye.

src/server/src/main/java/io/cassandrareaper/storage/MemoryStorageFacade.java
adejanovski: This path avoids persisting data to disk. It's used for testing where persistence is disabled.
The `stop()` method has a null check.

src/server/src/main/java/io/cassandrareaper/storage/snapshot/CassandraSnapshotDao.java
adejanovski: I used a simpler form because we're getting a single snapshot and pagination isn't expected.
Usually `executeAsync()` is used when there are multiple queries which need to be sent concurrently. It could be that I've used it in some cases where it wasn't absolutely necessary? It's ok to use it even in this case, but I may have some inconsistent code style then.

src/server/src/main/java/io/cassandrareaper/storage/repairrun/CassandraRepairRunDao.java
adejanovski: Sure, I guess we should refactor those before cutting out a final v4.0.0 to use the driver consistently across all DAOs.

burmanm (approved):

PR #1612, opened 2025-09-29
Fix the intensity rule which wasn't properly applied

### LINKED ISSUE #1613: Fix the intensity feature
The intensity feature doesn't seem to work anymore in Reaper due to changes that have been applied to segment orchestration in the past few years.
We need to bring it back to a working state.

The way it was previously done, Reaper would wait for a time depending on the previous repair duration before trying to schedule a new segment.
This cannot work anymore given we allow multiple segments to run concurrently if they belong to different repairs.
Instead, we should keep the lock on the replicas involved in the segment that just finished, for the duration that the intensity imposes as a pause. This would forcefully prevent any other segment from running on these replicas, creating the desired behavior.

adejanovski:
The intensity was so far applied by making the SegmentRunner thread sleep for the computed duration. This worked fine as long as this would prevent other segments from being started, which hasn't been the case since the redesign of segment orchestration (and locks) in v2.

This PR leverages the locks we put on Cassandra nodes and renews them for the duration of the intensity pause. This way no other segment can be started for the same repair that would involve one of these nodes, effectively giving them the expected break.


adejanovski:
### :tada: **Snyk checks have passed. No issues have been found so far.**

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/adejanovski/pr-checks/0a5fb3b4-24a1-4a21-8f98-72a05c0bc36d?refs=53eb42bf4e90116dd6419117dff6006098734d92&source=prChecksComment)

:white_check_mark: **code/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/adejanovski/project/92329191-e09b-4a92-bcbd-6fdcb09d13ca/pr-check/67d29508-8570-4361-9afe-1e3c2cba5778?refs=53eb42bf4e90116dd6419117dff6006098734d92&source=prChecksComment)

src/server/src/main/java/io/cassandrareaper/service/SegmentRunner.java
```diff
@@ -790,9 +805,7 @@ long intensityBasedDelayMillis(double intensity) {
       long repairDuration = Math.max(1, repairEnd - repairStart);
       long delay = (long) (repairDuration / intensity - repairDuration);
       LOG.debug("Scheduling next runner run() with delay {} ms", delay);
-      int nbRunningReapers = countRunningReapers();
-      LOG.debug("Concurrent reaper instances : {}", nbRunningReapers);
-      return delay * nbRunningReapers;
+      return delay;
```
adejanovski, line 808: The number of concurrent reaper instances isn't relevant anymore (if it ever was). We're now keeping the lock on cluster nodes which gives us the expected behavior.

burmanm (approved):

PR #1623, opened 2025-10-09
Add a new unfulfilledRepairSchedule metric to track efficiently repairs complete in due time

### LINKED ISSUE #1610: Create a metric that detects repairs not running in due time
We currently publish a metric that gives us the time since the last repair, which is handy to set up alerting but is inconvenient in many cases:
- If the repair never ran fully, it won't provide any value
- If the repair cycle is different between two schedules, the monitoring system has to know what the cycle in days is to set differentiated alerting (which is tricky)
- We can't easily track individual schedules in the metrics, although it should be our scope for monitoring

Instead (or on top) of this we should publish a new metric that will detect schedule violations in a binary way (can take 0 or 1 as value).
That metric would be set to 1 if we haven't had a completed repair for a schedule within its cycle time.
We would have to leave some room for variance in execution times, and deal with edge cases (rebuilding the set of metrics after a restart for example would be a good thing, we just start empty on startup).

We could allow for 10 to 20% of overage compared to the last repair (which could be configurable as a global setting), and compute it based on the end time for the previous repair run of that schedule.

## Definition of Done
- [ ] A new metric named repairCycleViolation is created that will be set to 0 by default for every schedule, and jumps to 1 if we don't have a completed repair since last finished repair (or schedule activation time) + x% (x being configurable)

adejanovski:
Fixes #1610

The current metric we use for this is `millisSinceLastRepair`, which doesn't account for the schedule repair interval.
This PR introduces a new metric named `io_cassandrareaper_service_RepairScheduleService_unfulfilledRepairSchedule` which will get a value of 1 if a repair hasn't completed within the interval defined for the schedule (and 0 if the last repair completed within that cycle time).

It will detect if a repair hasn't started within 10% over the planned activation time (which will happen if the previous repair is still running at the next planned activation time) and also if the last repair end time is over the schedule interval.
The metric will account for the state of the schedule and won't alert on PAUSED schedules. If a schedule gets unpaused and was paused for longer than the cycle time, the metric will be set to 1 though.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/service/RepairScheduleService.java
+++ b/src/server/src/main/java/io/cassandrareaper/service/RepairScheduleService.java
@@ -34,6 +34,7 @@ import static io.cassandrareaper.metrics.MetricNameUtils.cleanName;
 
 import com.codahale.metrics.Gauge;
 import com.codahale.metrics.MetricRegistry;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 import org.apache.cassandra.repair.RepairParallelism;
 import org.joda.time.DateTime;
@@ -42,6 +43,7 @@ public final class RepairScheduleService {
 
   public static final String MILLIS_SINCE_LAST_REPAIR_METRIC_NAME =
       "millisSinceLastRepairForSchedule";
+  public static final String UNFULFILLED_REPAIR_SCHEDULE_METRIC_NAME = "unfulfilledRepairSchedule";
   private final AppContext context;
   private final RepairUnitService repairUnitService;
 
@@ -176,6 +178,18 @@ public final class RepairScheduleService {
       context.metricRegistry.register(
           metricName, getMillisSinceLastRepairForSchedule(schedule.getId()));
     }
+
+    String unfulfilledRepairScheduleMetricName =
+        metricName(
+            UNFULFILLED_REPAIR_SCHEDULE_METRIC_NAME,
+            repairUnit.getClusterName(),
+            repairUnit.getKeyspaceName(),
+            schedule.getId());
+
+    if (!context.metricRegistry.getMetrics().containsKey(unfulfilledRepairScheduleMetricName)) {
+      context.metricRegistry.register(
+          unfulfilledRepairScheduleMetricName, getUnfulfilledRepairSchedule(schedule.getId()));
+    }
   }
 
   private void unregisterScheduleMetrics(UUID repairScheduleId) {
@@ -194,6 +208,20 @@ public final class RepairScheduleService {
 
           if (context.metricRegistry.getMetrics().containsKey(metricName)) {
             context.metricRegistry.remove(metricName);
+          }
+
+          String unfulfilledRepairScheduleMetricName =
+              metricName(
+                  UNFULFILLED_REPAIR_SCHEDULE_METRIC_NAME,
+                  repairUnit.getClusterName(),
+                  repairUnit.getKeyspaceName(),
+                  sched.getId());
+
+          if (context
+              .metricRegistry
+              .getMetrics()
+              .containsKey(unfulfilledRepairScheduleMetricName)) {
+            context.metricRegistry.remove(unfulfilledRepairScheduleMetricName);
           }
         });
   }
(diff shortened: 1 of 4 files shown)


adejanovski:
### :tada: **Snyk checks have passed. No issues have been found so far.**

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/adejanovski/pr-checks/06bab0f6-9c80-4833-9228-a5e633d4883e?refs=2ba8e22b53a6d1d105ae23f3a81f14d2487bfe40&source=prChecksComment)

:white_check_mark: **code/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/adejanovski/project/92329191-e09b-4a92-bcbd-6fdcb09d13ca/pr-check/adaa658e-db30-40e0-b211-d8f7a50e884d?refs=2ba8e22b53a6d1d105ae23f3a81f14d2487bfe40&source=prChecksComment)

src/server/src/main/java/io/cassandrareaper/service/RepairScheduleService.java
```diff
@@ -224,6 +252,56 @@ private Gauge<Long> getMillisSinceLastRepairForSchedule(UUID repairSchedule) {
     };
   }
 
+  @VisibleForTesting
+  Gauge<Long> getUnfulfilledRepairSchedule(UUID repairSchedule) {
```
gmanojDS: Noting that Prometheus handles both Integer and Long, but for binary 0/1 gauges the usual convention is to stick with Integer. Should we consider using Integer here for consistency, or is there a reason to prefer Long instead? just Curious

src/server/src/main/java/io/cassandrareaper/service/RepairScheduleService.java
```diff
@@ -224,6 +252,56 @@ private Gauge<Long> getMillisSinceLastRepairForSchedule(UUID repairSchedule) {
...
+              .filter(Optional::isPresent)
+              .map(Optional::get)
+              .filter(repairRun -> repairRun.getRunState() == RepairRun.RunState.DONE)
+              .map(RepairRun::getEndTime)
+              .filter(Objects::nonNull)
+              .map(dateTime -> DateTime.now().getMillis() - dateTime.getMillis());
+
+      if (millisSinceLastRepair.isPresent()) {
+        // we have a last repair that has finished successfully
+        if (millisSinceLastRepair.get() > schedule.get().getDaysBetween() * 24 * 60 * 60 * 1000) {
```
gmanojDS: minor: Would it be cleaner to compute schedule.get().getDaysBetween() * 24 * 60 * 60 * 1000 once and reuse it?

src/server/src/main/java/io/cassandrareaper/service/RepairScheduleService.java
```diff
@@ -224,6 +252,56 @@ private Gauge<Long> getMillisSinceLastRepairForSchedule(UUID repairSchedule) {
...
+                  .getLastRun());
+
+      // Cases where it is not relevant to check if the repair schedule is unfulfilled
+      if (schedule.get().getDaysBetween() == 0
+          || schedule.get().getState() == RepairSchedule.State.PAUSED
+          || schedule.get().getState() == RepairSchedule.State.DELETED) {
+        return 0l;
+      }
+
+      // Fail fast if we're past the next activation time by 10% of the cycle time
```
gmanojDS, line 275: Should we add a defensive check for nextActivation == null here, just to avoid assuming it always exists?

src/server/src/main/java/io/cassandrareaper/service/RepairScheduleService.java
adejanovski: No good reason to prefer the `Long` type, I'll switch to `Integer`

src/server/src/main/java/io/cassandrareaper/service/RepairScheduleService.java
adejanovski: It would, most definitely :) I'll do that.

src/server/src/main/java/io/cassandrareaper/service/RepairScheduleService.java
adejanovski: sure thing 👍

adejanovski:
@gmanojDS, could you please do another round of review?
Thanks!

gmanojDS (approved):

PR #1637, opened 2026-01-02
Add TLS configuration abilities to the server for Docker container

### LINKED ISSUE #1636: Add ability to configure server with TLS in the Docker container
The Dockerfiles and their configuration scripts do not allow setting the server itself to be behind a TLS connection. This needs to be configurable so that the TLS server setting is available through some env variables (which themselves are a bad thing, but that's the way currently..).

burmanm:
Fixes #1636 

WIP: We need to test this somehow, but that means more changes for the SimpleClient. For now, this PR verifies that I didn't break anything..


burmanm:
Since this can't be tested here, I've added deployment to the https://github.com/k8ssandra/k8ssandra-operator/pull/1662

rzvoncek (approved):

PR #1649, opened 2026-02-17
Change compaction strategy to SizeTieredCompactionStrategy

zencircle:
Support 3.11.x cassandra


adejanovski:
Hi @zencircle, thanks for sending this PR but we're not supporting Cassandra 3.x in newer versions of Reaper I'm afraid.
I'm also unsure how that related to a switch from LCS to STCS.

PR #1650, opened 2026-02-18
Fix NPE in RetryPolicyImpl caused by null request and nullable Request.isIdempotent()

emptyhammond:
This is the driver 4.x counterpart to the null-safety fix in #229 (driver 3.x).

When Reaper was migrated from driver 3.x to 4.x, the defensive null guards from #229 were not carried forward. The 4.x RetryPolicy callbacks can receive null parameters in failure paths (e.g. connection-level failures, abortAllInFlight), and Request.isIdempotent() returns Boolean (nullable). Using that value in a boolean context auto-unboxes null and throws an NPE. The driver treats the exception as a fatal error, closes the connection, and never retries. Under load this cascades into pool exhaustion, AllNodesFailedException, and a CPU-bound retry loop.

Changes:
- Restore #229-style null guards: Each callback (onReadTimeout, onWriteTimeout, onUnavailable, onRequestAborted, onErrorResponse) now checks for request == null at the top and returns RetryDecision.RETHROW with a warning log instead of dereferencing.
- Null-safe isIdempotent helper: Replace all 5 direct calls to request.isIdempotent() with a private helper that uses Boolean.TRUE.equals(), so null idempotency is treated as non-idempotent (rethrow, no retry) instead of throwing.
- No behaviour change when request is non-null and idempotency is explicitly set.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/storage/cassandra/RetryPolicyImpl.java
+++ b/src/server/src/main/java/io/cassandrareaper/storage/cassandra/RetryPolicyImpl.java
@@ -34,22 +34,26 @@ public final class RetryPolicyImpl implements RetryPolicy {
     return Math.min(delay, MAX_DELAY_MS);
   }
 
-  @Override
-  public RetryDecision onReadTimeout(
-      @NotNull Request request,
-      @NotNull ConsistencyLevel cl,
-      int required,
-      int received,
-      boolean retrieved,
-      int retry) {
+  private boolean isIdempotent(Request request) {
+    return Boolean.TRUE.equals(request.isIdempotent());
+  }
 
-    if (retry >= MAX_READ_RETRIES) {
-      LOG.warn(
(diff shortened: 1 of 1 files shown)


emptyhammond:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=thelastpickle_cassandra-reaper&pullRequest=1650) **Quality Gate failed**
> Failed conditions  [73.5% Duplication on New Code](https://sonarcloud.io/component_measures?id=thelastpickle_cassandra-reaper&pullRequest=1650&metric=new_duplicated_lines_density&view=list) (required ≤ 3%)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=thelastpickle_cassandra-reaper&pullRequest=1650)

Changes pushed to address this.

adejanovski (approved):
Looks good, thanks for the fixes!

PR #1652, opened 2026-02-22
Fix concurrent schema migration race condition in sidecar mode

kswia:
When multiple Reaper instances start simultaneously (e.g. sidecar mode), all instances race to perform Cassandra schema migration concurrently, causing stuck migrations due to schema disagreements.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/storage/cassandra/MigrationManager.java
+++ b/src/server/src/main/java/io/cassandrareaper/storage/cassandra/MigrationManager.java
@@ -45,6 +45,12 @@ final class MigrationManager {
 
   private static final Logger LOG = LoggerFactory.getLogger(MigrationManager.class);
 
+  /** Maximum time to wait for schema agreement after DDL statements. */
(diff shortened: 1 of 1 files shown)


adejanovski:
Hi @kswia, 

We solved this a while ago by [implementing leader election](https://github.com/patka/cassandra-migration/commit/bbe40159d0375d65f55729326509a648192b21e7) in the migration library itself to prevent concurrent schema upgrades.
I'm curious as to how we could still get disagreements. I guess there was still a window since a schema leader table is created upon init which will then be used for leader election. In practice, multiple instances can try creating that table concurrently, leading to schema disagreements.
But your implementation will suffer the same issue since you also create the `schema_migration_leader` table as part of the init.

Am I missing something?

kswia:
Hi @adejanovski,
you're absolutely right. I completely missed that the library already handles leader election via withConsensus=true in the MigrationTask constructor.
I've reworked the fix based on your feedback. Now it's much simpler and targets the actual root cause you identified.  By waiting for schema agreement right after that DDL, the library's built-in LWT leader election can then work reliably.

Issue #1658, opened 2026-03-17
NPE in class io.cassandrareaper.storage.cassandra.RetryPolicyImpl


he0871:
`Unexpected error while invoking the retry policy` caused by `NullPointerException` in `RetryPolicyImpl` when using Cassandra as backend. 

as request.isIdempotent() cloud be null  at if (request.isIdempotent()).

Reaper version: 4.0.1
Cassandra version: 5.0.3

adejanovski:
Fixed in v4.2.1

PR #1660, opened 2026-03-17
Fix NPE at RetryPolicyImp.java issue 1658

he0871:
Unexpected error while invoking the retry policy caused by NullPointerException in RetryPolicyImpl when using Cassandra as backend.

As request.isIdempotent() cloud be null at if (request.isIdempotent()).

Reaper version: 4.0.1
Cassandra version: 5.0.3

https://github.com/thelastpickle/cassandra-reaper/issues/1658

----------

First time to create a PR here. Do apologize if I did something wrong.


adejanovski (approved):
Nice fix, thanks!

PR #1661, opened 2026-03-19
 Fix metric registry leak in CompactionProxy due to unique executor names

a-pashchenko:
**Problem**:
CompactionProxy created an InstrumentedExecutorService with a unique name on every invocation ("CompactionProxy-" + Uuids.random()). Although the executor was stored in a static AtomicReference, the if (null == EXECUTOR.get()) check had a race condition — multiple threads could pass the null check simultaneously and create multiple executors. But even without the race, the main issue was different: each time an InstrumentedExecutorService was created with a new unique name, a new set of metrics was registered in the MetricRegistry and never removed. This led to unbounded metric growth (metric leak).

**Fix**:
- Replaced the if (null == EXECUTOR.get()) { EXECUTOR.set(...) } condition with atomic creation under a lock, which eliminates the race condition and guarantees a one-time initialization of the executor.
- Removed Uuids.random() from the executor name — now uses a fixed name "CompactionProxy", since the executor is created exactly once and a unique suffix is unnecessary.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/management/CompactionProxy.java
+++ b/src/server/src/main/java/io/cassandrareaper/management/CompactionProxy.java
@@ -47,7 +47,7 @@ public final class CompactionProxy {
 
   private CompactionProxy(ICassandraManagementProxy proxy, MetricRegistry metrics) {
     this.proxy = proxy;
-    if (EXECUTOR.get() != null) {
+    if (EXECUTOR.get() == null) {
       synchronized (EXECUTOR) {
         if (EXECUTOR.get() != null) {
           EXECUTOR.set(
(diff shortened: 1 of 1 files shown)


src/server/src/main/java/io/cassandrareaper/management/CompactionProxy.java
```diff
@@ -26,32 +26,40 @@
...
 
   private final ICassandraManagementProxy proxy;
 
   private CompactionProxy(ICassandraManagementProxy proxy, MetricRegistry metrics) {
     this.proxy = proxy;
-    if (null == EXECUTOR.get()) {
-      EXECUTOR.set(
-          new InstrumentedExecutorService(
-              Executors.newCachedThreadPool(), metrics, "CompactionProxy-" + Uuids.random()));
+    if (EXECUTOR != null) {
```
adejanovski: **Suggestion**: instead of using a lock, wouldn't we simplify the code by using a synchronized section instead, and keep the AtomicReference?

a-pashchenko: Done! Replaced the lock with a synchronized block while keeping the AtomicReference.

adejanovski (approved):
Thanks a lot! LGTM

PR #1668, opened 2026-04-09
fix: remove JVM exit on SchedulingManager errors

### LINKED ISSUE #1645: Exiting the JVM in case of SchedulingManager failure is a little too extreme
In a scenario where one out of 2 DCs go down, Reaper will fail running Quorum queries that it needs for schedule management.
As such, if assertions are enabled, it will shut down entirely.

While that's manageable in a kubernetes environment where Reaper will get restarted automatically, it is a more challenging situation in non k8s environments.

We should probably make that an opt in feature rather than an opt out one and favor the stability of Reaper in such a case.

artsiom-bukhautsou:
Removes JVM exit on SchedulingManager errors

Fixes: #1645

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/service/SchedulingManager.java
+++ b/src/server/src/main/java/io/cassandrareaper/service/SchedulingManager.java
@@ -154,9 +154,6 @@ public final class SchedulingManager extends TimerTask {
         } else {
           LOG.error("Failed managing repair schedule with id '{}'", lastId, ex);
         }
-        if (context.isRunning.get()) {
-          LOG.error("SchedulingManager encountered an error but will continue running.");
-        }
       }
     }
   }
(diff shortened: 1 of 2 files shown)


src/server/src/main/java/io/cassandrareaper/service/SchedulingManager.java
```diff
@@ -154,13 +154,8 @@ public void run() {
...
         }
-        try {
-          assert false : "if assertions are enabled then exit the jvm";
-        } catch (AssertionError ae) {
-          if (context.isRunning.get()) {
-            LOG.error("SchedulingManager failed. Exiting JVM.");
-            System.exit(1);
-          }
+        if (context.isRunning.get()) {
+          LOG.error("SchedulingManager encountered an error but will continue running.");
```
adejanovski: **todo:** I think we can fully remove this block. We already have exceptions logged above and this one doesn't really add anything useful for our users.

adejanovski (changes requested):

src/server/src/main/java/io/cassandrareaper/service/SchedulingManager.java
artsiom-bukhautsou: agree, fixed

adejanovski (approved):

PR #1670, opened 2026-05-06
Fix: Support configurable serial consistency for LWT queries in multi…

n4u71lu5-j3d1:
## Context

Adds support for a configurable serial consistency level for LWT (Light Weight Transaction) queries in Cassandra Reaper. The immediate problem is that in multi-DC Cassandra deployments, using the default SERIAL consistency for LWT operations (leader election, running-repairs tracking) causes WriteTimeoutException because Paxos must reach quorum across ALL datacenters. Setting LOCAL_SERIAL restricts the Paxos quorum to the local DC, resolving the issue.

## Changes in the Commit

```
┌─────────────────────────────────┬────────────────────────────────────────────────────────────────────────────────────────────────────┐
│              File               │                                            What changed                                            │
├─────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ CassandraConcurrencyDao.java    │ Added serialConsistencyLevel field; updated constructor; applied it to all 4 LWT statements        │
│                                 │ (takeLead, renewLead, releaseLead, setRunningRepairs)                                              │
├─────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ CassandraStorageFacade.java     │ Reads requestSerialConsistency from dropwizard-cassandra RequestOptionsFactory; defaults to SERIAL │
│                                 │  if absent                                                                                         │
├─────────────────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────┤
│ cassandra-reaper-cassandra.yaml │ Added commented-out requestSerialConsistency: LOCAL_SERIAL example with explanatory comments       │
└─────────────────────────────────┴────────────────────────────────────────────────────────────────────────────────────────────────────┘
```

## Problem

In multi-DC Cassandra deployments, Reaper's LWT (Light Weight Transaction) operations (leader election and running-repairs tracking) use `SERIAL` consistency by default. `SERIAL` requires a Paxos quorum across **all** datacenters, which can trigger `WriteTimeoutException` when cross-DC latency is high or a remote DC is degraded.

## Solution

Make the serial consistency level for LWT prepared statements configurable via the existing `requestOptionsFactory` block in `cassandra-reaper-cassandra.yaml`. Operators running multi-DC clusters can now set `LOCAL_SERIAL` to restrict the Paxos quorum to the local DC.

### Configuration (multi-DC clusters)

```yaml
cassandra:
requestOptionsFactory:
    requestTimeout: 60s
    requestDefaultIdempotence: true
    requestSerialConsistency: LOCAL_SERIAL   # restricts Paxos quorum to local DC
```

Leave the option commented out (or omit it entirely) to retain the existing SERIAL
behavior — no breaking change for current deployments.

### DIFF
--- a/src/packaging/resource/cassandra-reaper-cassandra.yaml
+++ b/src/packaging/resource/cassandra-reaper-cassandra.yaml
@@ -117,6 +117,9 @@ cassandra:
   requestOptionsFactory:
     requestTimeout: 60s
     requestDefaultIdempotence: true
+    # Use LOCAL_SERIAL for multi-DC deployments to restrict LWT Paxos quorum to the local DC.
+    # Resolves WriteTimeoutException at consistency SERIAL in multi-DC setups.
+    # requestSerialConsistency: LOCAL_SERIAL
   authProvider:
     type: plain-text
     username: cassandra
--- a/src/server/src/main/java/io/cassandrareaper/storage/cassandra/CassandraConcurrencyDao.java
+++ b/src/server/src/main/java/io/cassandrareaper/storage/cassandra/CassandraConcurrencyDao.java
@@ -50,6 +50,7 @@ public class CassandraConcurrencyDao {
   private final Version version;
   private final UUID reaperInstanceId;
   private final CqlSession session;
+  private final ConsistencyLevel serialConsistencyLevel;
   private PreparedStatement takeLeadPrepStmt;
   private PreparedStatement renewLeadPrepStmt;
   private PreparedStatement releaseLeadPrepStmt;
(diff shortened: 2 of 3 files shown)


adejanovski:
Hi @n4u71lu5-j3d1,

the purpose of these LWTs is to have a leader election system so that Reaper instances running concurrently can safely pick up segments to run without any risk of having them processed twice.
It also allows ensuring that only one segment from each repair run would involve a given replica, and spread the load evenly across the cluster.
Lowering the CL to the local DC breaks those guarantees.
I would say that if you plan on running multiple Reaper instances working together on the same cluster(s), you need to keep the SERIAL CL for LWTs. If you only have a single instance, I'd recommend the badly named memory storage which now uses sqlite and can persist data locally: https://cassandra-reaper.io/docs/backends/memory/
It will be much more efficient than the cassandra storage, but won't allow coordination between multiple reaper instances.

PR #1679, opened 2026-05-19
Enable metrics collection through https when hitting TLS endpoints

### LINKED ISSUE #1680: Use https as scheme when the metrics client has TLS enabled
When support was added for pulling metrics from the Cassandra nodes with TLS enabled, one part was forgotten: using the https scheme for the calls.
As a result, metrics pulls are failing when TLS is enabled in the metrics endpoint of the management API.

adejanovski:
Fixes #1680
While we were building a TLS enabled http client, we were always using the `http` scheme.
This PR makes it so that https is used when TLS is enabled in the metrics client.


rzvoncek (approved):
Manually tested, works as expected.

Issue #1673, opened 2026-05-14
Codebase update to use Dropwizard 5.*


tengu-alt:
Hello all, and thank you for your work!

The next Dropwizard release should contain a fix for [CVE-2026-2332](https://github.com/advisories/GHSA-355h-qmc2-wpwf) (which is displayed as Critical on our security scans), and it is crucial to us (and I'm sure - for other folks) to resolve this issue as fast as possible.

The current Dropwizard version that is used in cassandra-reaper is 4.0.15, which is officially [EOL](https://github.com/dropwizard/dropwizard/discussions/10799) since January of 2026. 

 In order to resolve this vulnerability, Cassandra-Reaper should be able to support newer 5.* versions of Dropwizard when it will be released with bumped version of Jetty (vulnerable component).

As I understand, some codebase updates should be done in order to support newer versions.
When can we expect such updates?
Best regads!

mattsheppard-instaclustr:
Dropwizard 5.0.2, which actually addresses that mentioned CVE, is released now (https://github.com/dropwizard/dropwizard/releases/tag/v5.0.2), so it would be nice to try to move this forward.

From a brief look the code changes required here are not complex - They come mostly from the upgrade to Jetty 12, but since Jetty 12 requires JDK 17, that would force cassandra-reaper to follow along and start requiring JDK17.

I'm not sure how the project would feel about changing to require JDK17 though - Any input from current contributors?

(I presume a switch to JDK17 is unavoidable unless the plan here is to somehow move away from Dropwizard and Jetty entirely to something that continues to be supported on JDK11)

tengu-alt:
Great news!
Let's discuss this code update so Reaper can support never versions.
@adejanovski What do you think?

adejanovski:
Hi folks, 

this is definitely something we want to do. There are quite a few changes indeed, and no way around moving to JDK 17.
It matches what Cassandra 5 supports as well, which work for me.
I tend to have a strong focus on Kubernetes where switching JDKs is not a problem thanks to isolation. I can understand that folks not running Reaper in a container could have different constraints, but despite that we need to move forward.

There's no plan to move away from Dropwizard, especially if that's motivated by sticking with an older version of Java.

Contributions are welcome, if someone feels like performing the upgrade (might not be trivial), otherwise our team will probably tackle that in the next couple months.

tengu-alt:
> Hi folks,

> this is definitely something we want to do...

[...]

PR #1684, opened 2026-06-02
fix: Fix CVEs in cassandra reaper

ShreyashD333:
Fixed CVE
image

Note: Cannot use jetty-http, jetty-io v12.x package as its supported from Java 17+

CVEs present in OS of Docker image
image

### DIFF
--- a/src/server/pom.xml
+++ b/src/server/pom.xml
@@ -190,7 +190,7 @@
         <dependency>
             <groupId>org.apache.commons</groupId>
             <artifactId>commons-configuration2</artifactId>
-            <version>2.12.0</version>
+            <version>2.15.0</version>
         </dependency>
         <dependency>
             <groupId>io.jsonwebtoken</groupId>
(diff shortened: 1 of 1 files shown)


adejanovski (approved):

mattsheppard-instaclustr:
For my understanding @ShreyashD333 - Do you believe this fixes [CVE-2026-2332](https://www.cve.org/CVERecord?id=CVE-2026-2332)?

I understood it to be fixed only in Jetty 12.x because 11.x is 'end of life' from the Jetty project's perspective (and hence no expectation it would ever be fixed there in the open source version). https://mvnrepository.com/artifact/org.eclipse.jetty/jetty-http/11.0.26 still reports that CVE against the set version here, but I notice your table above lists it as fixed, so perhaps you know something I don't.

I was expecting it would only be fixed by something like #1673.

tengu-alt:
> For my understanding @ShreyashD333 - Do you believe this fixes [CVE-2026-2332](https://www.cve.org/CVERecord?id=CVE-2026-2332)?

Agree with @mattsheppard-instaclustr. [Advisory](https://github.com/advisories/GHSA-355h-qmc2-wpwf) states that 11.0.26 is in range of the affected versions for [CVE-2026-2332](https://github.com/advisories/GHSA-355h-qmc2-wpwf)

ShreyashD333:
Hey, Thanks for pointing it out, the CVE-2026-2332 is not fixed with this PR. I have updated the description with fixed CVE in the PR. In upcoming release we will try to fix it.
Thanks.

PR #1686, opened 2026-06-03
fix: scheduled repairs never re-fire after a completed run (#1683)

### LINKED ISSUE #1683: Schedule never re-fires after first successful run until purge deletes the previous run (regression in 4.2.2)
Hi,

I just updated from 4.0.1 to 4.2.3 and found some buggy behavior. My scheduled weekly runs were not activating and clicking "Run now" on the schedule also didn't fire.:
```
INFO   [2026-05-29 09:57:49,839] [SchedulingManagerTimer] i.c.s.SchedulingManager - repair unit '811a8fc0-eb09-11f0-952b-0dcbc7b3f437' should be repaired based on RepairSchedule with id '794c7500-59e6-11f1-817f-a147de180cb5'
INFO   [2026-05-29 09:57:49,846] [SchedulingManagerTimer] i.c.s.SchedulingManager - there is repair (id df2b1270-59e9-11f1-a3a3-6599ab37b5a8) in state 'DONE' for repair unit '811a8fc0-eb09-11f0-952b-0dcbc7b3f437', postponing current schedule trigger until next scheduling
```

Deleting the run history fixed this. I got Claude to write this up for me and obviously verified its finding, it appears correct to me. For now I've lowered `purgeRecordsAfterInDays` as a workaround.

**Affected versions:** 4.2.2, 4.2.3 (master). Not present in 4.0.x, 4.1.x, 4.2.0, 4.2.1.

**Summary**

Since 4.2.2, an active repair schedule fires its first scheduled run successfully, then refuses to fire again until the previous `DONE` (or `ERROR`/`ABORTED`) run is removed from storage by `PurgeService`. With default `purgeRecordsAfterInDays: 30`, a `scheduled_days_between: 7` schedule effectively runs every ~30 days, not 7. Clicking "Run now" in the UI does not help - it advances `next_activation` to now, the scheduler bumps it forward by `days_between`, then refuses to fire due to the same check, pushing the next visible activation a full cycle into the future.

**Reproduction**

1. Run Reaper 4.2.2 or 4.2.3 with default config (`purgeRecordsAfterInDays: 30`).
2. Create a repair schedule with `scheduled_days_between: 7`. Let it fire once and complete (state `DONE`).
3. Wait for the next activation, or click "Run now" in the UI to advance `next_activation`.
4. Observe in the logs:

   ```
   SchedulingManager - repair unit '<unit-id>' should be repaired based on RepairSchedule with id '<schedule-id>'
   SchedulingManager - there is repair (id <previous-done-id>) in state 'DONE' for repair unit '<unit-id>', postponing current schedule trigger until next scheduling
   ```

5. The previously-DONE run referenced is the one from the previous scheduled execution. No new run is created. `next_activation` has been advanced by `scheduled_days_between` regardless.

The schedule will keep postponing every cycle until `PurgeService` deletes the prior `DONE` run (default: after 30 days).

**Root cause**

PR #1672 ("Populate metrics on startup", merged 2026-05-13, commit `5752a6ad`) removed the state filter from `repairRunComesFromSchedule` in `SchedulingManager.java`:

```diff
   private static boolean repairRunComesFromSchedule(RepairRun repairRun, RepairSchedule schedule) {
-    return repairRun.getRunState().isActive()
-        || (RepairRun.RunState.NOT_STARTED == repairRun.getRunState()
-            && repairRun.getCause().equals(getCauseName(schedule)));
+    return repairRun.getCause().equals(getCauseName(schedule));
   }
```

`repairRunAlreadyScheduled` (called from `manageSchedule`) iterates **all** historical runs for the unit returned by `getRepairRunsForUnit`, with no state or age filter, and returns true on the first cause-string match. With the state filter removed, terminated runs (`DONE`, `ERROR`, `ABORTED`) that came from the schedule now block all future automatic firings of that schedule until those rows are deleted from `repair_run_by_unit` (which has no TTL - `default_time_to_live = 0` in the schema).

The PR was primarily about metrics; the `SchedulingManager` change appears to be an unintentional simplification - the original two-clause expression encoded a deliberate "ignore terminated runs" semantic that the new one-liner drops. No tests in the PR cover the schedule re-firing path.

**Why this slipped in**

The same PR added a new method `RepairScheduleService.findLastCompletedRepairRun` that calls `getRepairRunsForUnit` and explicitly filters by `DONE`:

```java
repairRuns.stream()
    .filter(run -> run.getRunState() == RepairRun.RunState.DONE)
    .filter(run -> run.getEndTime() != null)
    .min(RepairRun::compareTo)
    .map(RepairRun::getId);
```

So the need to filter results from that DAO was clearly on the author's mind - the same filter just didn't make it into the four-line change in `SchedulingManager`.

Likely how it happened: `repairRunComesFromSchedule` was previously misnamed - it returned true for "any active run, or NOT_STARTED with matching cause," which is really a "should we hold off firing this schedule" check rather than literally "did this run come from this schedule." Simplifying it to cause equality reads as a cleanup that matches the method name, but the only caller (`repairRunAlreadyScheduled`) was relying on the broader semantics. The new tests all target `getUnfulfilledRepairSchedule`; nothing exercises the `manageSchedule` → `repairRunAlreadyScheduled` path.

**Expected behavior**

A schedule with `scheduled_days_between: N` should fire approximately every N days (modulo run duration), independent of how many days of run history are retained. Terminated runs from prior cycles should not block the next cycle's firing.

**Suggested fix**

Restore the state filter:

```java
private static boolean repairRunComesFromSchedule(RepairRun repairRun, RepairSchedule schedule) {
  return repairRun.getRunState().isActive()
      || (RepairRun.RunState.NOT_STARTED == repairRun.getRunState()
          && repairRun.getCause().equals(getCauseName(schedule)));
}
```

And ideally add a test in `SchedulingManagerTest` that:
1. Creates a schedule and lets it complete one run to `DONE`.
2. Advances `next_activation` past now.
3. Calls `manageSchedule` again.
4. Asserts a new `RepairRun` is created.

**Workarounds**

- Downgrade to 4.2.1.
- Lower `purgeRecordsAfterInDays` to less than `scheduled_days_between` so the purge keeps up (tradeoff: less run history).
- Manually `DELETE /repair_run/<id>` the most recent scheduled `DONE` run after each cycle (operationally painful).

onnos:
Fixes #1683.

PR #1672 reduced `repairRunComesFromSchedule()` to a pure cause-string match. That made `repairRunAlreadyScheduled()` treat terminated runs (DONE/ERROR/ABORTED) carrying the schedule's cause as still blocking, so a schedule stayedpostponed until the previous run was purged from storage - effectively running on the purge interval (default 30 days) instead of its configured cadence.
Regression in 4.2.2; not present in 4.0.x–4.2.1.

Rather than reverting #1672 (which would re-couple `maybeRegisterRepairRunCompleted()` to the run's transient state), this restores the pre-#1672 postponement predicate in a dedicated `repairRunBlocksSchedule()` helper - any active run on the unit, or a NOT_STARTED run from this schedule - while keeping the cause-only match for the schedule-metrics path added in #1672.

Adds `SchedulingManagerTest` coverage for re-firing after a completed run and for postponing while a run from the same schedule is still in flight.

Tested this against a .deb built with the same process as in CI and seems to work as expected.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/service/SchedulingManager.java
+++ b/src/server/src/main/java/io/cassandrareaper/service/SchedulingManager.java
@@ -78,14 +78,14 @@ public final class SchedulingManager extends TimerTask {
   }
 
   /**
-   * Postpone a schedule only while a conflicting run is still in flight: any active run on the
-   * unit, or a not-yet-started run from this schedule. Terminated runs must not block it, else the
+   * Postpone a schedule only while one of its own runs is still in flight: an active or
+   * not-yet-started run that comes from this schedule. Terminated runs must not block it, else the
    * schedule stays stuck until the completed run is purged.
    */
   private static boolean repairRunBlocksSchedule(RepairRun repairRun, RepairSchedule schedule) {
-    return repairRun.getRunState().isActive()
-        || (RepairRun.RunState.NOT_STARTED == repairRun.getRunState()
-            && repairRunComesFromSchedule(repairRun, schedule));
+    return (repairRun.getRunState().isActive()
+            || RepairRun.RunState.NOT_STARTED == repairRun.getRunState())
+        && repairRunComesFromSchedule(repairRun, schedule);
   }
 
   private static String getCauseName(RepairSchedule schedule) {
--- a/src/server/src/main/java/io/cassandrareaper/service/SchedulingManager.java
+++ b/src/server/src/main/java/io/cassandrareaper/service/SchedulingManager.java
@@ -77,6 +77,17 @@ public final class SchedulingManager extends TimerTask {
     return repairRun.getCause().equals(getCauseName(schedule));
   }
 
+  /**
+   * Postpone a schedule only while a conflicting run is still in flight: any active run on the
+   * unit, or a not-yet-started run from this schedule. Terminated runs must not block it, else the
+   * schedule stays stuck until the completed run is purged.
+   */
+  private static boolean repairRunBlocksSchedule(RepairRun repairRun, RepairSchedule schedule) {
+    return repairRun.getRunState().isActive()
+        || (RepairRun.RunState.NOT_STARTED == repairRun.getRunState()
+            && repairRunComesFromSchedule(repairRun, schedule));
+  }
+
   private static String getCauseName(RepairSchedule schedule) {
     return "scheduled run (schedule id " + schedule.getId().toString() + ')';
   }
@@ -253,7 +264,7 @@ public final class SchedulingManager extends TimerTask {
     Collection<RepairRun> repairRuns =
         repairRunDao.getRepairRunsForUnit(schedule.getRepairUnitId());
     for (RepairRun repairRun : repairRuns) {
-      if (repairRunComesFromSchedule(repairRun, schedule)) {
+      if (repairRunBlocksSchedule(repairRun, schedule)) {
         LOG.info(
             "there is repair (id {}) in state '{}' for repair unit '{}', "
                 + "postponing current schedule trigger until next scheduling",
--- a/src/server/src/test/java/io/cassandrareaper/service/SchedulingManagerTest.java
+++ b/src/server/src/test/java/io/cassandrareaper/service/SchedulingManagerTest.java
@@ -438,6 +438,106 @@ public final class SchedulingManagerTest {
     Mockito.verify(context.repairManager, Mockito.times(0)).startRepairRun(any());
   }
 
+  // A terminated run (DONE/ERROR/ABORTED) must not block the schedule from firing again.
+  @Test
+  public void manageScheduleStartsNewRunWhenPreviousRunIsDone() throws ReaperException {
+    RepairUnit unit = testUnit();
+    UUID scheduleId = Uuids.timeBased();
+    RepairSchedule schedule = dueSchedule(unit.getId(), scheduleId);
+    AppContext context =
+        contextWithExistingRuns(unit, scheduledRun(unit.getId(), scheduleId, RunState.DONE));
+    when(context.storage.getClusterDao()).thenReturn(mock(IClusterDao.class));
+
+    RepairRunService repairRunService = mock(RepairRunService.class);
+    when(repairRunService.registerRepairRun(any(), any(), any(), any(), any(), any(), any(), any()))
+        .thenReturn(scheduledRun(unit.getId(), scheduleId, RunState.NOT_STARTED));
+    SchedulingManager schedulingManager =
+        SchedulingManager.create(
+            context, () -> repairRunService, context.storage.getRepairRunDao());
+
+    assertTrue(
+        "A completed previous run must not block the schedule from firing again",
+        schedulingManager.manageSchedule(schedule));
+    Mockito.verify(context.repairManager).startRepairRun(any());
+  }
+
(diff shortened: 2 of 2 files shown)


onnos:
@adejanovski could I ask for your review on this one? Seems like a pretty bad bug in the sense that it surfaces in the default configuration and users may not notice it for a while, which means nothing gets repaired for a month.

adejanovski:
Sure thing, that's a bad bug indeed.
Let me review that tomorrow morning and circle back.

src/server/src/main/java/io/cassandrareaper/service/SchedulingManager.java
```diff
@@ -77,6 +77,17 @@ private static boolean repairRunComesFromSchedule(RepairRun repairRun, RepairSch
...
 
+  /**
+   * Postpone a schedule only while a conflicting run is still in flight: any active run on the
+   * unit, or a not-yet-started run from this schedule. Terminated runs must not block it, else the
+   * schedule stays stuck until the completed run is purged.
+   */
+  private static boolean repairRunBlocksSchedule(RepairRun repairRun, RepairSchedule schedule) {
+    return repairRun.getRunState().isActive()
+        || (RepairRun.RunState.NOT_STARTED == repairRun.getRunState()
+            && repairRunComesFromSchedule(repairRun, schedule));
```
adejanovski: **Issue:** this condition was wrong initially already, because it returns false if the tested repair run is active, regardless of the schedule it's coming from, but checks the not started ones if they're associated to the schedule.
If we take this method in isolation, it's incorrect because it only works well if the tested repair run is guaranteed to be part of the schedule.

There are two ways to solve this:
- either we always test the association between the schedule and the repair run
- or we don't test that and just check the state.

I'd be keen to do the former:


```suggestion
    return (repairRun.getRunState().isActive()
        || (RepairRun.RunState.NOT_STARTED == repairRun.getRunState())
        && repairRunComesFromSchedule(repairRun, schedule);
```

src/server/src/main/java/io/cassandrareaper/service/SchedulingManager.java
```diff
@@ -253,7 +264,7 @@ private boolean repairRunAlreadyScheduled(RepairSchedule schedule, RepairUnit re
     Collection<RepairRun> repairRuns =
         repairRunDao.getRepairRunsForUnit(schedule.getRepairUnitId());
     for (RepairRun repairRun : repairRuns) {
-      if (repairRunComesFromSchedule(repairRun, schedule)) {
+      if (repairRunBlocksSchedule(repairRun, schedule)) {
```
adejanovski, line 267: Yes, good catch. `repairRunComesFromSchedule` was used in two different contexts which I've obviously missed, introducing this bug.
It's better to have a separate method to handle this.

adejanovski (changes requested):
Thanks for pushing a fix for the bug I introduced 😅 
There's just one small change I'd like to make to avoid future issues depending on how `repairRunBlocksSchedule` could be used in the future.

src/server/src/main/java/io/cassandrareaper/service/SchedulingManager.java
onnos: Thanks, makes sense! I applied your suggestion (just with a small parenthesis fix) and added a small test for this scenario.

adejanovski (approved):

PR #1690, opened 2026-06-11
Handle topology changes by creating replacement repair segments

gmanojDS:
## Summary

When a repair is running and the cluster topology changes (for example during a scale-out operation), some existing repair segments may no longer represent valid token ranges under the new topology.

Today, if Reaper attempts to process one of these outdated segments, coordinator selection can fail with a "no potential coordinator" error, causing the repair to fail.

This change introduces topology-aware segment recovery. Instead of failing the repair immediately, Reaper detects when a segment has become invalid due to a topology change, generates valid replacement segments based on the current token ownership, and allows the repair to continue.

### RepairRunner

* Added topology change detection in the outdated segment failure path.
* Attempts topology-aware recovery before failing the repair.
* Computes replacement segments using the latest token boundaries.
* Verifies complete coverage of the original token range before retiring the original segment.

### Segment Recovery

* Uses a create-first approach:

  1. Compute replacement segments.
  2. Create any missing replacement segments.
  3. Verify full token range coverage.
  4. Retire the original segment.

* Prevents token range gaps if Reaper crashes or a storage operation fails during recovery.

### Storage Layer

Added support for:

* Looking up repair segments by token range.
* Conditional segment state updates.
* Batch creation of replacement segments through the storage abstraction.

Implemented for:

* Memory storage
* Cassandra storage

### Idempotency

The recovery flow is safe to retry:

* Existing replacement segments are detected and reused.
* Only missing replacement segments are created.
* The original segment is retired only after replacement coverage is verified.

## Testing

Added unit tests covering:

* Segment splitting logic.
* Coverage verification.
* Token range lookup.
* Conditional segment retirement.
* Replacement segment creation behavior.
* Edge cases around invalid and overlapping ranges.

## Example

Before topology change:

```text
Segment: 1 -> 10
```

After cluster expansion:

```text
1 -> 4
5 -> 10
```

Previous behavior:

```text
No coordinator found
→ Repair fails
```

New behavior:

```text
No coordinator found
→ Detect topology change
→ Create replacement segments:
   1 -> 4
   5 -> 10
→ Retire original segment
→ Continue repair
```


gmanojDS:
Manual validation completed on CDEV cluster 411cd32d-5bd8-4ee4-8331-b161696550bd.

Steps:
- Created 3-node C40i cluster and loaded data into testks.keyvalue.
- Deployed custom Reaper image: manojtg/cassandra-reaper:topology-fix.
- Created repair while cluster was 3 nodes and left it NOT_STARTED.
- Scaled CassandraDatacenter dc-1 from 3 to 6 nodes.
- Waited until all 6 nodes were UN.
- Activated the repair.

Result:
- Reaper hit the "no potential coordinators" path.
- Topology change was detected.
- Replacement segments were created.
- Coverage verification succeeded.
- Original invalid segments were retired.
- Repair continued and completed successfully.

Evidence:
- Logs show “Successfully handled topology change for segment ..., repair will continue”.
- Logs show replacement segments completing with “returned with state DONE”.
- Logs show “Repairs for repair run #7e57c490-68b0-11f1-b1b7-eff092f54e43 done”.

gmanojDS:
### Topology change detected and handled
```
INFO  [2026-06-15 12:21:59,232] [caas-cluster:7e57c490-68b0-11f1-b1b7-eff092f54e43]
i.c.s.RepairRunner - Successfully handled topology change for segment #7e5ad1d0-68b0-11f1-b1b7-eff092f54e43, repair will continue

INFO  [2026-06-15 12:22:09,280] [caas-cluster:7e57c490-68b0-11f1-b1b7-eff092f54e43]
i.c.s.RepairRunner - Successfully handled topology change for segment #7e5c3160-68b0-11f1-b1b7-eff092f54e43, repair will continue

INFO  [2026-06-15 12:22:19,327] [caas-cluster:7e57c490-68b0-11f1-b1b7-eff092f54e43]
i.c.s.RepairRunner - Successfully handled topology change for segment #7e594b30-68b0-11f1-b1b7-eff092f54e43, repair will continue
```

### Replacement segments executed successfully
```
INFO ... SegmentRunner - Repair command 1 on segment da6a7d50-68b4-11f1-b1b7-eff092f54e43 returned with state DONE
INFO ... SegmentRunner - Repair command 2 on segment d46ea8e0-68b4-11f1-b1b7-eff092f54e43 returned with state DONE
INFO ... SegmentRunner - Repair command 155 on segment fea98300-68b4-11f1-b1b7-eff092f54e43 returned with state DONE
```

### Repair completed
```
INFO [2026-06-15 12:24:00,462]
i.c.s.RepairRunner - Repair amount done 12.0

INFO [2026-06-15 12:24:00,462]
i.c.s.RepairRunner - Repairs for repair run #7e57c490-68b0-11f1-b1b7-eff092f54e43 done
```

src/server/src/main/java/io/cassandrareaper/storage/repairsegment/IRepairSegmentDao.java
```diff
@@ -51,4 +52,58 @@ public interface IRepairSegmentDao {
...
+   * @return true if the segment was updated (condition matched), false otherwise
+   */
+  default boolean updateRepairSegmentStateConditional(
+      UUID runId,
+      UUID segmentId,
+      RepairSegment.State newState,
+      RepairSegment.State expectedCurrentState) {
+    // Default implementation delegates to the version without runId
+    return updateRepairSegmentStateConditional(segmentId, newState, expectedCurrentState);
+  }
```
adejanovski: **Issue:** I'm confused with the presence of two interface methods, which seem to accommodate each of the implementation classes differently.
If that's the case, then we have a design issue has the interface should provide unique method signatures that the implementations should override. Creating one that calls the other, with the original one throwing an unsupported operation exception doesn't look right.
What's the rationale behind this? Is there a reason why the version with the runId isn't the only one we define?

src/server/src/main/java/io/cassandrareaper/storage/repairsegment/CassandraRepairSegmentDao.java
```diff
@@ -367,4 +375,211 @@ public Collection<RepairSegment> getSegmentsWithState(
...
+  @Override
+  public boolean updateRepairSegmentStateConditional(
+      UUID segmentId, RepairSegment.State newState, RepairSegment.State expectedCurrentState) {
+    // For Cassandra backend, we need to use lightweight transactions (LWT) for conditional updates
+    // CQL: UPDATE repair_run SET segment_state = ? WHERE id = ? AND segment_id = ? IF segment_state
+    // = ?
+
+    // This method requires runId context which is not available with just segmentId.
+    // Use the overloaded version that accepts runId parameter.
+    throw new UnsupportedOperationException(
```
adejanovski: **Issue:** we should throw a ReaperException (which can wrap the UnsupportedOperationException), and adjust the interface method signature to add `throws ReaperException`.
But let's wait before doing anything and assess if we really need two versions of that method.

src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
```diff
@@ -923,6 +942,408 @@ private synchronized void failRepairDueToOutdatedSegment(UUID segmentId, Segment
...
+      // Only retire if it's still in NOT_STARTED state
+      boolean retired =
+          context
+              .storage
+              .getRepairSegmentDao()
+              .updateRepairSegmentStateConditional(
+                  originalSegment.getRunId(),
+                  originalSegment.getId(),
+                  RepairSegment.State.DONE,
+                  RepairSegment.State.NOT_STARTED);
```
adejanovski, line 1330: **question:** could there be a race condition between multiple reaper instances trying to make changes on this segment or are we covered by leader election?
What happens if that operation fails, will it get retried?

adejanovski (changes requested):

src/server/src/main/java/io/cassandrareaper/storage/repairsegment/IRepairSegmentDao.java
gmanojDS: The intent was to support Cassandra’s need for runId while keeping the simpler signature for other storage implementations, but that makes the interface awkward. I’ll simplify this to expose only the runId version and update all implementations to override that.

src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
gmanojDS: Leader election should prevent multiple Reaper instances from processing the same segment concurrently. The conditional update is an extra safeguard so we only retire the original segment if it is still NOT_STARTED.

If that update fails, topology recovery returns false and the repair follows the existing error path. Replacement segment creation is idempotent, so a later attempt can reuse existing replacements instead of creating duplicates.

src/server/src/main/java/io/cassandrareaper/storage/repairsegment/CassandraRepairSegmentDao.java
gmanojDS: If we remove the non-runId method, this becomes moot because there won’t be an unsupported method anymore. If we did keep it for any reason, it should throw ReaperException, but I think the cleaner fix is to remove that method entirely.

src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
```diff
@@ -923,6 +942,408 @@ private synchronized void failRepairDueToOutdatedSegment(UUID segmentId, Segment
...
+      if (ringTokens.isEmpty()) {
+        LOG.warn("No ring tokens available from cluster");
+        return replacements;
+      }
+
+      // Sort ring tokens
+      Collections.sort(ringTokens);
+
+      BigInteger startToken = segmentTokenRange.getBaseRange().getStart();
+      BigInteger endToken = segmentTokenRange.getBaseRange().getEnd();
```
adejanovski, line 1082: **Issue:** I hadn't spotted that before, but a segment can contain multiple token ranges.
The base range itself is not enough, you need to cycle through the token ranges which can be accessed through `segmentTokenRange.getTokenRanges()`.
So if you're splitting the base range and then deleting the segment, you'll be skipping some tokens from those ranges which are coalesced in the segment.

adejanovski: I think we need to treat each range that's part of a segment separately in their own new segment, and disable the coalescing in that case. Managing coalescing of the newly generated splits seems too complex for that first iteration.

adejanovski (changes requested):

gmanojDS:
## Manual Topology Change Validation

Validated the topology-change segment splitting/recovery flow end-to-end against a live AMC Cassandra cluster using Cassandra-backed Reaper storage.

### Test Objective

Validate that when Cassandra topology changes after repair segments have already been generated, Reaper:

1. Detects stale repair segments that no longer have valid coordinators.
2. Handles coalesced segments containing multiple token ranges.
3. Splits **all constituent token ranges**, rather than only the base/first range.
4. Creates replacement segments based on the current topology.
5. Verifies complete coverage of every original token range.
6. Handles wrap-around token ranges correctly.
7. Retires the original segment only after replacement coverage is verified.
8. Continues repairing the replacement segments.
9. Completes the repair successfully.


Repair run:

```text
d2ab5db0-b2a5-11f1-b424-ab31acb48e6a
```

### Test Procedure

1. Started with a healthy 3-node Cassandra cluster.
2. Captured the generated repair segments.
3. Confirmed that all 4 generated segments were coalesced/multi-range segments.
4. Expanded Cassandra from 3 nodes to 6 nodes.
5. Waited until all 6 Cassandra nodes were healthy.
6. Confirmed that the repair was still `NOT_STARTED`, meaning the persisted segments were still based on the original 3-node topology.
7. Started the stale repair against the new 6-node topology.
8. Observed Reaper detect invalid segments and enter topology-change recovery.
9. Allowed Reaper to create and execute the replacement segments until the entire repair completed.

### Original Segments

The repair was initially generated with **4 segments**, and all of them contained multiple token ranges:

```text
d2ababd0-b2a5-11f1-b424-ab31acb48e6a ranges=8
d2abf9f0-b2a5-11f1-b424-ab31acb48e6a ranges=7
d2ac2100-b2a5-11f1-b424-ab31acb48e6a ranges=7
d2ac4810-b2a5-11f1-b424-ab31acb48e6a ranges=2
```

### Topology Change Detection

After starting the stale 3-node repair against the 6-node ring, Reaper encountered segments for which the persisted token range could no longer find a valid coordinator and entered the topology-change recovery path.

Example:

```text
Segment #d2abf9f0-b2a5-11f1-b424-ab31acb48e6a is faulty, no potential coordinators for range...
```

Instead of permanently failing the repair, Reaper detected the topology change and generated replacement segments using the current ring.

---

### Multi-Range Recovery: 7 Original Ranges

One of the original coalesced segments contained **7 token ranges**.

Reaper generated **14 replacement segments across all 7 original ranges**, verified complete coverage, and only then retired the original segment.

```text
Verified complete coverage with 14 replacement segments across 7 original token range(s)

Retiring original segment #d2abf9f0-b2a5-11f1-b424-ab31acb48e6a

Successfully retired original segment #d2abf9f0-b2a5-11f1-b424-ab31acb48e6a

Successfully retired original segment #d2abf9f0-b2a5-11f1-b424-ab31acb48e6a after topology change

Successfully handled topology change for segment #d2abf9f0-b2a5-11f1-b424-ab31acb48e6a, repair will continue
```

This confirms that recovery was not limited to only one token range from the coalesced segment.

---

### Multi-Range Recovery: 8 Original Ranges

A second original segment contained **8 token ranges**.

Reaper generated **17 replacement segments across all 8 ranges**:

```text
Verified complete coverage with 17 replacement segments across 8 original token range(s)

Retiring original segment #d2ababd0-b2a5-11f1-b424-ab31acb48e6a

Successfully retired original segment #d2ababd0-b2a5-11f1-b424-ab31acb48e6a

Successfully retired original segment #d2ababd0-b2a5-11f1-b424-ab31acb48e6a after topology change

Successfully handled topology change for segment #d2ababd0-b2a5-11f1-b424-ab31acb48e6a, repair will continue
```

This is the strongest validation of the coalesced/multi-range fix: a single stale repair segment containing 8 ranges was expanded into 17 valid replacement segments while preserving complete coverage.

---

### Additional 7-Range Recovery

Another original 7-range segment also entered topology recovery successfully:

```text
Verified complete coverage with 14 replacement segments across 7 original token range(s)

Retiring original segment #d2ac2100-b2a5-11f1-b424-ab31acb48e6a

Successfully retired original segment #d2ac2100-b2a5-11f1-b424-ab31acb48e6a

Successfully retired original segment #d2ac2100-b2a5-11f1-b424-ab31acb48e6a after topology change

Successfully handled topology change for segment #d2ac2100-b2a5-11f1-b424-ab31acb48e6a, repair will continue
```

So the topology recovery behavior was reproduced across multiple original multi-range segments during the same repair run.

---

### Per-Range Coverage Verification

Coverage was verified independently for the constituent ranges before the original segment was retired.

Examples:

```text
Verified complete coverage for range [3939437640763611309, 4881313981049535108) with 2 replacement segments

Verified complete coverage for range [7004039717411533086, 8126610941490667320) with 2 replacement segments

Verified complete coverage for range [-109199958328730541, 593978172128338129) with 3 replacement segments

Verified complete coverage for range [-5509783172976984861, -4598799676847958705) with 2 replacement segments

Verified complete coverage for range [1281100629711040537, 1728553392004483838) with 1 replacement segments
```

This verifies that replacement coverage is checked against each original range rather than treating a coalesced segment as a single base range.

### Wrap-Around Range Validation

The test also exercised a Murmur3 wrap-around token range.

Reaper successfully verified its replacement coverage:

```text
Verified complete coverage for range [9170529409608998445, -8602360958978690837) with 3 replacement segments
```

This confirms that coverage verification also works when the original range crosses the Murmur3 ring boundary.

---

### Replacement Segments Continued Repairing

After the original stale segments were retired, Reaper continued selecting and repairing the dynamically generated replacement segments.

Examples from the replacement repair execution:

```text
Repair command 5 on segment bbb11620-b2ac-11f1-bdc2-f1b3dba43525 returned with state DONE

Repair command 6 on segment 347bf020-b2ad-11f1-bdc2-f1b3dba43525 returned with state DONE

Repair command 7 on segment 348258c0-b2ad-11f1-bdc2-f1b3dba43525 returned with state DONE

Repair command 94 on segment ec2235a0-b2ac-11f1-bdc2-f1b3dba43525 returned with state DONE

Repair command 79 on segment 3487fe10-b2ad-11f1-bdc2-f1b3dba43525 returned with state DONE

Repair command 80 on segment 34800ed0-b2ad-11f1-bdc2-f1b3dba43525 returned with state DONE
```

Additional replacement segments continued completing successfully:

```text
Repair command 8 on segment 347cda80-b2ad-11f1-bdc2-f1b3dba43525 returned with state DONE

Repair command 5 on segment 3479cd40-b2ad-11f1-bdc2-f1b3dba43525 returned with state DONE

Repair command 6 on segment 347debf0-b2ad-11f1-bdc2-f1b3dba43525 returned with state DONE

Repair command 95 on segment ec1eda40-b2ac-11f1-bdc2-f1b3dba43525 returned with state DONE

Repair command 5 on segment ec247f90-b2ac-11f1-bdc2-f1b3dba43525 returned with state DONE

Repair command 6 on segment bbacd060-b2ac-11f1-bdc2-f1b3dba43525 returned with state DONE

Repair command 7 on segment ec236e20-b2ac-11f1-bdc2-f1b3dba43525 returned with state DONE

Repair command 81 on segment ec26f090-b2ac-11f1-bdc2-f1b3dba43525 returned with state DONE

Repair command 9 on segment bbb24ea0-b2ac-11f1-bdc2-f1b3dba43525 returned with state DONE

Repair command 8 on segment 3484a2b0-b2ad-11f1-bdc2-f1b3dba43525 returned with state DONE

Repair command 7 on segment 34839140-b2ad-11f1-bdc2-f1b3dba43525 returned with state DONE
```

This confirms that the replacement segments were not merely inserted into storage; they were subsequently picked up and successfully repaired.

---

### End-to-End Repair Completion

At the end of the run, Reaper reported:

```text
Repair amount done 49.0
Repairs for repair run #d2ab5db0-b2a5-11f1-b424-ab31acb48e6a done
```

Final repair state from the API:

```json
{
  "id": "d2ab5db0-b2a5-11f1-b424-ab31acb48e6a",
  "state": "DONE",
  "segments_repaired": 49,
  "total_segments": 4,
  "end_time": "2026-09-17T15:39:03Z"
}
```

The repair originally contained 4 coalesced segments. During topology recovery, those stale segments were replaced by new segments matching the updated ring, resulting in 49 repaired segments by completion.

src/server/src/main/java/io/cassandrareaper/service/RepairRunner.java
gmanojDS: made the required changes and have attached the testing result.

PR #1704, opened 2026-07-23
Prevent poller termination during concurrent repair cleanup

### LINKED ISSUE #1665: Reaper loses track of repair notifications with the management api
After some time, it looks like Reaper often loses track of repair completion notifications when using the management API.
We can see in the Cassandra logs that the repairs complete correctly, but debugging logs don't show that Reaper is getting notified of completions.
This usually leads to segments timing out and being retried, without any further progress.

Restarting Reaper fixes the issue and ongoing repairs make progress again up to completion.

gmanojDS:
### Summary

Fix a race condition in the Reaper Management API notification poller that could terminate the notificationsTracker task during concurrent repair cleanup.

### Problem
removeRepairStatusHandler() removed the repair executor before removing the corresponding jobTracker entry. If notificationsTracker() was iterating jobTracker during this window, it could observe a live job entry whose executor had already been removed. This resulted in a NullPointerException when calling executor.submit(...).

Since notificationsTracker() is scheduled using ScheduledExecutorService.scheduleWithFixedDelay(), any uncaught exception permanently suppresses future executions of the poller, causing subsequent repairs to complete only via timeout.

### Changes
- Reordered cleanup in removeRepairStatusHandler() to remove the jobTracker entry before removing the handler and executor.
- Hardened notificationsTracker() by:
    - isolating failures on a per-job basis,
    - handling transient getJobStatus() failures,
    - guarding against concurrently removed executors and handlers,
    - preventing exceptions from escaping the scheduled poller,
    - capturing the handler before submitting work to avoid an additional race.
- Added a deterministic unit test that reproduces the concurrent cleanup race using CountDownLatch and verifies that notificationsTracker().run() completes without throwing.

### Testing
- Added deterministic race-condition unit test covering concurrent cleanup.
- Existing project tests compile successfully (environment-specific Mockito limitations under Java 25 remain unchanged).

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
+++ b/src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
@@ -634,11 +634,21 @@ public class HttpCassandraManagementProxy implements ICassandraManagementProxy {
         try {
           final String jobId = entry.getKey();
           Job job = getJobStatus(jobId);
+          if (job.getStatusChanges() == null) {
+            LOG.warn("Job {} returned null statusChanges in notificationsTracker, skipping", jobId);
+            continue;
+          }
           int availableNotifications = job.getStatusChanges().size();
           int currentNotificationCount = entry.getValue().latestNotificationCount.get();
           if (currentNotificationCount < availableNotifications) {
+            String rawId = job.getId();
+            if (rawId == null || !rawId.startsWith("repair-") || rawId.length() <= 7) {
+              LOG.warn(
+                  "Job {} has malformed id '{}' in notificationsTracker, skipping", jobId, rawId);
+              continue;
+            }
             // remove "repair-" prefix once per job, not per notification
-            int repairNo = Integer.parseInt(job.getId().substring(7));
+            int repairNo = Integer.parseInt(rawId.substring(7));
             for (int i = currentNotificationCount; i < availableNotifications; i++) {
               dispatchNotification(repairNo, jobId, i, job.getStatusChanges().get(i));
               entry.getValue().latestNotificationCount.incrementAndGet();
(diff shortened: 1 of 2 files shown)


src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
```diff
@@ -627,41 +630,99 @@ private void scheduleJobPoller(int pollInterval) {
...
-                      });
-
-              // Update the count as we process them
-              entry.getValue().latestNotificationCount.incrementAndGet();
-            }
-          }
+      for (Map.Entry<String, JobStatusTracker> entry : jobTracker.entrySet()) {
+        try {
+          processJobEntry(entry);
+        } catch (Exception e) {
```
adejanovski, line 636: **issue:** catch `RuntimeException` instead. As a convention in Reaper we never catch `Exceptions` because it will swallow checked exceptions.
Java's checked/unchecked split exists for a reason. A checked exception (IOException, InterruptedException, etc.) is one the compiler forces you to acknowledge because it represents an expected recoverable condition. If catch (Exception e) swallows one silently, you've hidden a bug the compiler would otherwise have surfaced.

The most dangerous case here is InterruptedException. If the thread running the notificationsTracker() runnable is interrupted (e.g., during JVM shutdown or a test teardown), catch (Exception e) will:
- eat the interrupt,
- log "will retry on next poll",
- and loop on without restoring Thread.interrupted().
This is the classic "swallowed interrupt" anti-pattern. The thread will keep running when it should be stopping.

src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
```diff
@@ -627,41 +630,99 @@ private void scheduleJobPoller(int pollInterval) {
...
+          jobId,
+          index);
+      return;
+    }
+
+    try {
+      executor.submit(
+          () ->
+              handler.handle(repairNo, Optional.of(progressType), statusChange.getMessage(), this));
+    } catch (Exception e) {
```
adejanovski, line 714: **Issue:** same as before, we should never catch `Exceptions` but catch `RuntimeExceptions` to get the widest scope of exceptions to catch.

adejanovski (changes requested):

adejanovski (approved):

PR #1710, opened 2026-07-28
Guard notificationsTracker against poller death in HTTP management mode

### LINKED ISSUE #1709: HTTP management mode: repair job-status poller dies permanently on first failed poll, wedging all repairs until pod restart
## Project board / version info

- Reaper: 4.2.5 (defect also verified present in 4.0.1 and current `master` @ bd718f65)
- Cassandra: 4.1.3, managed by k8ssandra-operator, management-API (HTTP) mode
- Storage backend: Cassandra (`reaper_db`)
- Deployment: Kubernetes, one Reaper per cluster

## Symptom

Repair runs randomly get stuck in `RUNNING` forever with
`last_event = "All nodes are busy or have too many pending compactions for the remaining candidate segments."`
while the Cassandra cluster is completely idle (`compactionstats` → 0 pending, `nodetool repair_admin list` → no sessions, all nodes UP/NORMAL).

The stuck run holds one or more zombie segments in state `STARTED` with `endTime = null`. Reaper's in-memory `currentlyRunningSegments` treats those segments' replicas as busy, so every remaining candidate segment is postponed and the whole repair queue (with `maxParallelRepairs=1`: all runs) jams behind it. `hangingRepairTimeoutMins` does not recover it. The only recovery is restarting the Reaper pod, after which everything works again — until the next occurrence.

We hit this repeatedly over 3 months (2026-04-11, 05-09, 05-21, 06-16, 06-18, 07-09), on different clusters, different keyspaces, and three different coordinator nodes, which sent us down several wrong paths (node-specific degradation, mgmt-api heap starvation, replication factor) before we found the actual cause.

## Root cause

`HttpCassandraManagementProxy` learns about repair completion exclusively through a background poller:

```java
private void scheduleJobPoller(int pollInterval) {
    statusTracker.scheduleWithFixedDelay(
        notificationsTracker(), pollInterval * 2, pollInterval, TimeUnit.MILLISECONDS);
}
```

`ScheduledExecutorService.scheduleWithFixedDelay` **permanently and silently cancels the task if the runnable throws** (the exception is only surfaced through the never-read `ScheduledFuture`). The runnable has no exception handling:

```java
Runnable notificationsTracker() {
    return () -> {
      if (jobTracker.size() > 0) {
        for (Map.Entry<String, JobStatusTracker> entry : jobTracker.entrySet()) {
          Job job = getJobStatus(entry.getKey());     // throws RuntimeException on any ApiException
          ...
          int repairNo = Integer.parseInt(job.getId().substring(7));                 // can throw
          ProgressEventType progressType = ProgressEventType.valueOf(statusChange.getStatus()); // can throw
          repairStatusExecutors.get(repairNo).submit(...);                           // NPE if handler already removed
          ...
```

```java
private Job getJobStatus(String id) {
    try {
      return apiClient.getJobStatus(id);
    } catch (ApiException e) {
      throw new RuntimeException(e);    // ← one transient mgmt-api error kills the poller forever
    }
}
```

So a single failed poll — one transient management-API timeout, connection reset, or a malformed/unknown status — permanently kills the poller **for the lifetime of the pod**. From that moment on:

1. `triggerRepair` still works (`POST /api/v2/repairs` → `202 Accepted`), Cassandra runs and completes the repair,
2. but no one ever polls `/api/v0/ops/executor/job`, so the completion notification never reaches `SegmentRunner`,
3. the segment waits out the full status-wait timeout, gets cancelled/postponed/re-triggered in an endless loop, and stays `STARTED` in storage → zombie → wedge.

This explains every property of the failure mode we and others have observed: pod restart always fixes it (fresh poller thread), the "bad coordinator" appears random (it's whichever poll happened to fail), it is keyspace- and cluster-agnostic, and Cassandra logs show the repair as finished while Reaper's segment stays `STARTED`.

## Evidence from our production incident (2026-07-09 → 07-14)

Counting Reaper's requests in the management-API access logs (`Cli.java` lines) across all 6 Cassandra nodes:

| date | `GET /api/v0/ops/executor/job` calls |
|---|---|
| 07-07 | 37,919 |
| 07-08 | 79,510 |
| 07-09 | 11,547 (last one at **04:00:41**, then nothing) |
| 07-10 → 07-14 | **0** |

The active run froze at exactly that moment (460/478 segments) and 25 more runs queued up behind it over the following days. Repair triggers kept flowing after the poller died — the coordinator's access log shows the re-trigger pair with nothing in between and nothing after:

```
08:15:25,430 Cli.java:672 - address=/10.244.94.180:53774 url=/api/v2/repairs status=202 Accepted   ← abort/re-trigger
08:15:34,997 Cli.java:672 - address=/10.244.94.180:53774 url=/api/v2/repairs status=202 Accepted   ← new attempt
(only k8s probe traffic afterwards — no job status poll ever again)
```

while Reaper (DEBUG) shows the segment loop waiting on a notification that can never arrive:

```
08:15:25,426 SegmentRunner - Repair command 1306 on segment 8553c4a5-… returned with state STARTED
08:15:25,426 SegmentRunner - Repair command 1306 on segment 8553c4a5-… has been cancelled while running
08:15:25,426 SegmentRunner - Postponing segment 8553c4a5-…
08:15:25,427 SegmentRunner - Aborting repair on segment with id 8553c4a5-… on coordinator 10.244.107.29
08:15:34,999 SegmentRunner - Repair for segment 8553c4a5-… started, status wait will timeout in 54000000 millis
08:15:34,999 RepairRunner  - Triggered repair of segment 8553c4a5-… via host 10.244.107.29
```

## Proposed fix

Make the scheduled runnable exception-safe: catch per job entry (so one bad job doesn't block status processing of the others) and add an outer `Throwable` guard so nothing can escape and cancel the task; log and retry on the next tick.

We have been running exactly that patch in production (built on top of 4.2.5), together with a regression test (`testNotificationsTrackerSurvivesPollingFailure`: first poll throws `ApiException`, assert nothing propagates and the next poll still delivers the notification — fails on unpatched 4.2.5).

## Validation from two weeks running the patch in production

Since deploying the patched 4.2.5 (2026-07-14) we have had zero wedges across several full schedule cycles. The patch also survived a real sustained failure: on 2026-07-26 a tracked job's coordinator pod IP became permanently unreachable (`java.net.NoRouteToHostException` on every poll). The guard logged "Failed to poll status of job repair-0, will retry on the next poll" ~49,000 times over the following three days while the poller stayed alive and kept delivering status for all other jobs — on unpatched code the **first** of those failures would have silently killed the poller and wedged every subsequent repair.

(That incident also shows a pre-existing, separate issue this fix merely makes visible and survivable: `jobTracker` entries for jobs whose coordinator is gone are never evicted, so a dead host produces one WARN per poll tick until restart. Out of scope here.)

## Steps to reproduce

1. Run Reaper in HTTP management mode with an active repair run.
2. Make one job-status poll fail once (e.g. drop/timeout a single `GET /api/v0/ops/executor/job` response, or briefly restart the management API on the coordinator).
3. All subsequent repairs are triggered but never observed to complete; segments pile up in `STARTED`; runs eventually show "All nodes are busy…" while Cassandra is idle. Only a Reaper restart recovers.

zehweh:
Fixes #1709

## Problem

In HTTP management mode, `HttpCassandraManagementProxy` learns about repair completion exclusively from a job-status poller scheduled with `scheduleWithFixedDelay(notificationsTracker(), …)`. `scheduleWithFixedDelay` permanently and silently cancels the task on the first uncaught exception, `notificationsTracker()` has no exception handling, and `getJobStatus()` rethrows any `ApiException` as `RuntimeException`. One transient management-API error (timeout, connection reset, unreachable coordinator) therefore kills the poller for the lifetime of the pod. Every later repair completes invisibly, segments stay `STARTED` forever, and runs wedge with "All nodes are busy or have too many pending compactions for the remaining candidate segments." while the cluster is idle. Only a Reaper restart recovers. Full analysis and production evidence in #1709.

## Fix

Make the scheduled runnable exception-safe:

- per-job-entry `catch (RuntimeException)` — one failing job skips only that job; the other tracked jobs still get their status processed in the same tick;
- outer `catch (Throwable)` — nothing whatsoever can escape the runnable;
- both paths log and the poller retries on the next tick.

The outer guard catches `Exception` (per SonarCloud rule S1181). An escaping JVM `Error` would still cancel the task, but swallowing `Error`s tends to hide conditions where failing loudly is preferable. Log volume is naturally bounded by the poll interval (one line per failing job per tick).

Secondary throw points covered by the same guards: `Integer.parseInt` on the job id, `ProgressEventType.valueOf` on an unknown status, and a potential NPE from `repairStatusExecutors.get(repairNo)` when a handler was already removed.

## Regression test

`testNotificationsTrackerSurvivesPollingFailure`: the first poll throws `ApiException`, the test asserts nothing propagates and that the next poll still delivers the notification. It fails on unpatched source (the first `run()` throws) and passes with the fix. `HttpCassandraManagementProxyTest` is green on this branch (31/31).

## Production validation

We have been running this patch (applied to 4.2.5) in production since 2026-07-14: zero wedges since, after six wedges in the preceding three months. The guard also survived a real sustained failure — a tracked job's coordinator pod IP became permanently unreachable and the poller logged "will retry on the next poll" ~49,000 times over three days while continuing to deliver status for all other jobs. On unpatched code the first of those failures would have wedged every subsequent repair.

### DIFF
--- a/src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
+++ b/src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
@@ -632,50 +632,45 @@ public class HttpCassandraManagementProxy implements ICassandraManagementProxy {
       // repair completions are never observed again and segments stay STARTED
       // forever. Nothing thrown here may escape.
       try {
-        if (jobTracker.size() > 0) {
-          for (Map.Entry<String, JobStatusTracker> entry : jobTracker.entrySet()) {
-            try {
-              Job job = getJobStatus(entry.getKey());
-              int availableNotifications = job.getStatusChanges().size();
-              int currentNotificationCount = entry.getValue().latestNotificationCount.get();
-
-              if (currentNotificationCount < availableNotifications) {
-                // We need to process the new ones
-                for (int i = currentNotificationCount; i < availableNotifications; i++) {
-                  StatusChange statusChange = job.getStatusChanges().get(i);
-                  // remove "repair-" prefix
-                  int repairNo = Integer.parseInt(job.getId().substring(7));
-                  ProgressEventType progressType =
-                      ProgressEventType.valueOf(statusChange.getStatus());
-                  repairStatusExecutors
-                      .get(repairNo)
-                      .submit(
-                          () -> {
-                            repairStatusHandlers
-                                .get(repairNo)
-                                .handle(
-                                    repairNo,
-                                    Optional.of(progressType),
-                                    statusChange.getMessage(),
-                                    this);
-                          });
-
-                  // Update the count as we process them
-                  entry.getValue().latestNotificationCount.incrementAndGet();
-                }
-              }
-            } catch (RuntimeException e) {
-              LOG.warn(
-                  "Failed to poll status of job {}, will retry on the next poll",
-                  entry.getKey(),
-                  e);
-            }
+        for (Map.Entry<String, JobStatusTracker> entry : jobTracker.entrySet()) {
+          try {
+            processJobNotifications(entry);
+          } catch (RuntimeException e) {
+            LOG.warn(
+                "Failed to poll status of job {}, will retry on the next poll", entry.getKey(), e);
           }
         }
       } catch (Throwable t) {
         LOG.error("Job status poller iteration failed, will retry on the next poll", t);
       }
     };
+  }
+
+  private void processJobNotifications(Map.Entry<String, JobStatusTracker> entry) {
+    Job job = getJobStatus(entry.getKey());
+    int availableNotifications = job.getStatusChanges().size();
+    int currentNotificationCount = entry.getValue().latestNotificationCount.get();
+
+    if (currentNotificationCount < availableNotifications) {
+      // We need to process the new ones
+      for (int i = currentNotificationCount; i < availableNotifications; i++) {
+        StatusChange statusChange = job.getStatusChanges().get(i);
+        // remove "repair-" prefix
+        int repairNo = Integer.parseInt(job.getId().substring(7));
+        ProgressEventType progressType = ProgressEventType.valueOf(statusChange.getStatus());
+        repairStatusExecutors
+            .get(repairNo)
+            .submit(
+                () -> {
+                  repairStatusHandlers
+                      .get(repairNo)
+                      .handle(repairNo, Optional.of(progressType), statusChange.getMessage(), this);
+                });
+
+        // Update the count as we process them
+        entry.getValue().latestNotificationCount.incrementAndGet();
+      }
+    }
   }
 
   // Coordinator nodes such as Stargate instances do not have tokens
--- a/src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
+++ b/src/server/src/main/java/io/cassandrareaper/management/http/HttpCassandraManagementProxy.java
@@ -633,8 +633,8 @@ public class HttpCassandraManagementProxy implements ICassandraManagementProxy {
       // forever. Nothing thrown here may escape.
       try {
         pollTrackedJobs();
-      } catch (Throwable t) {
-        LOG.error("Job status poller iteration failed, will retry on the next poll", t);
+      } catch (Exception e) {
+        LOG.error("Job status poller iteration failed, will retry on the next poll", e);
       }
     };
   }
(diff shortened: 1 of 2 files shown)


zehweh:
Note on the SonarCloud quality gate: the smells introduced by this branch (S3776 cognitive complexity, S1141 nested try, S1181 catch-Throwable) have all been addressed in follow-up commits. The two remaining counted issues are `java:S107` (constructors with 9 and 8 parameters, lines 96/122) — they pre-date this PR (SonarCloud creation date 2025-06-04) and this diff does not touch those lines, but the PR analysis attributes them to new code anyway. They weigh 40min of debt against this PR's ~510 LoC development cost, which caps the new-code maintainability rating at B, so the gate cannot reach the required A from this diff without an out-of-scope refactor of those constructors.

adejanovski:
Hi @zehweh,

thanks for pushing this PR. I think it's a duplicate of https://github.com/thelastpickle/cassandra-reaper/pull/1704.
Would you mind reviewing it and let us know if you think your PR has some valuable bits that would be missing in Manoj's PR?

Thanks

PR #1699, opened 2026-07-09
Bumped jackson-databind to 2.21.6, httpclient5 to 5.6.4, httpcore5  to 5.4.3 and logback to 1.5.35 to resolve CVEs

dheerajmogaveera:
### Changes (`src/server/pom.xml`)

| Artifact | Before | After | Addresses |
|---|---|---|---|
| `com.fasterxml.jackson.core:jackson-databind` | 2.21.4 | **2.21.6** | CVE-2026-68497 (High), CVE-2026-83557, CVE-2026-19032, CVE-2026-77310, CVE-2026-59889, CVE-2026-54515 |
| `org.apache.httpcomponents.client5:httpclient5` | 5.6.1 (transitive) | **5.6.4** | CVE-2026-71290 (Critical), CVE-2026-64607 |
| `org.apache.httpcomponents.core5:httpcore5` / `httpcore5-h2` | 5.4.2 (transitive) | **5.4.3** | CVE-2026-54399, CVE-2026-54428 |
| `logback.version` | 1.5.33 | **1.5.35** | CVE-2026-10532 |

PR #1713, opened 2026-07-31
Bump lz4-java 1.11.2, amqp-client 5.35.0 and Netty 4.2.17.Final

dheerajmogaveera:
Bumps vulnerable transitive dependencies in `src/server/pom.xml`.

### lz4
- Exclude legacy `org.lz4:lz4-java`
- Pin `at.yawk.lz4:lz4-java` **1.11.2**
- Covers **CVE-2026-59949**, **XRAY-1046408**, **XRAY-1046409**

### amqp-client
- Pin `com.rabbitmq:amqp-client` **5.35.0**
- Covers **CVE-2026-61634**, **CVE-2026-63335**, **CVE-2026-63336**, **CVE-2026-63337**, **CVE-2026-75516**, **XRAY-1059086**

### Netty
- Introduce `${netty.version}` (**4.2.17.Final**) and use it for codec modules plus direct `netty-handler` / `netty-transport-native-epoll`
- Covers **CVE-2026-59901** (and later Netty 4.2.17 findings)

Also merged `master` so this branch is current. Duplicate `amqp-client` 5.33.1 from master was dropped so only 5.35.0 remains.

Jackson / httpclient5 are out of scope — see #1699.

Verified with `mvn validate` (BanDuplicatePomDependencyVersions passed) and `mvn dependency:tree`.

### DIFF
--- a/src/server/pom.xml
+++ b/src/server/pom.xml
@@ -32,6 +32,7 @@
         <dropwizard.version>5.0.2</dropwizard.version>
         <jetty.version>12.1.10</jetty.version>
         <jersey.version>3.1.10</jersey.version>
+        <netty.version>4.2.16.Final</netty.version>
         <logback.version>1.5.33</logback.version>
         <cxf.version>3.4.5</cxf.version>
         <prometheus.version>0.16.0</prometheus.version>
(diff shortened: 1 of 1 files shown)


src/server/pom.xml
```diff
@@ -438,27 +444,38 @@
...
+            <dependency>
+                <groupId>at.yawk.lz4</groupId>
+                <artifactId>lz4-java</artifactId>
+                <version>1.11.1</version>
+            </dependency>
             <dependency>
                 <groupId>io.netty</groupId>
                 <artifactId>netty-codec</artifactId>
-                <version>4.2.15.Final</version>
+                <version>4.2.16.Final</version>
```
adejanovski: **Suggestion:** should we turn this into a variable to avoid repeating the version 6 times?

dheerajmogaveera: Thank you @adejanovski  for the suggestion! I've updated the PR accordingly — introduced a netty.version property and replaced the repeated hardcoded versions with it.

PR #1692, opened 2026-06-25
This PR provides fix for the following Netty vulnerabilities: CVE-2026-45416, CVE-2026-44249, CVE-2026-50010 (netty-handler) and CVE-2026-45536 (netty-transport-native-epoll)

dheerajmogaveera:
This PR provides fix for the following vulnerabilities by upgrading Netty to 4.2.15.Final:

### io.netty:netty-handler (INS-42540)
- CVE-2026-45416 (High) — DoS via eager buffer allocation in SslClientHelloHandler.decode()
- CVE-2026-44249 (High) — IPv6 subnet rule bypass in IpSubnetFilterRule.compareTo()
- CVE-2026-50010 (High) — Missing hostname verification via X509TrustManagerWrapper

### io.netty:netty-transport-native-epoll (INS-42541)
- CVE-2026-45536 (Medium) — File descriptor leak in netty_unix_socket_recvFd

### Changes
- Bump `netty-handler` to 4.2.15.Final
- Add explicit `netty-transport-native-epoll` 4.2.15.Final and exclude the transitive version pulled in by `cassandra-all`

PR #1715, opened 2026-08-05
This PR provides fix for the following Netty vulnerabilities: CVE-202…

gmanojDS:
## Summary

Backports the Netty security fixes from #1692 to the `4.2` branch.

The current `4.2.5-ubi` image contains vulnerable Netty versions:

* `netty-handler:4.2.4.Final`
* `netty-transport-native-epoll:4.1.130.Final`

## Changes

* Upgraded `netty-handler` to `4.2.15.Final`
* Excluded the vulnerable transitive `netty-transport-native-epoll` dependency from `cassandra-all`
* Added a direct dependency on `netty-transport-native-epoll:4.2.15.Final`

This remediates:

* CVE-2026-44249
* CVE-2026-45416
* CVE-2026-50010
* CVE-2026-45536

## Validation

* Confirmed all core `io.netty:netty-*` dependencies resolve to `4.2.15.Final`
* Confirmed no core Netty `4.1.x` dependencies remain
* All tests passed
* Full Maven build completed successfully
* `netty-tcnative:2.0.70.Final` remains unchanged because it uses an independent versioning stream and is not affected by these CVEs

The rebuilt UBI image will require a follow-up Twistlock scan to confirm that the vulnerabilities are no longer reported in the final artifact.

### DIFF
--- a/src/server/pom.xml
+++ b/src/server/pom.xml
@@ -123,6 +123,10 @@
             <version>5.0.7</version>
             <exclusions>
                 <exclusion>
+                    <groupId>io.netty</groupId>
+                    <artifactId>netty-transport-native-epoll</artifactId>
+                </exclusion>
+                <exclusion>
                     <groupId>org.slf4j</groupId>
                     <artifactId>slf4j-log4j12</artifactId>
                 </exclusion>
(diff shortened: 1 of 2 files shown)


src/server/pom.xml
```diff
@@ -371,6 +380,111 @@
         </dependency>
     </dependencies>
 
+    <dependencyManagement>
+        <dependencies>
+            <!-- ============================================================ -->
```
rzvoncek, line 385: Nit: do we really need these comments?

rzvoncek (approved):
Looks all right, except we perhaps don't need the commens in pom.xml.

PR #1719, opened 2026-08-28
Fix CVEs

ShreyashD333:
Fixes following CVEs 
CVE-2026-59901, CVE-2026-54428, CVE-2026-63337, CVE-2026-69219, CVE-2026-69220, CVE-2026-54399

### DIFF
--- a/src/server/pom.xml
+++ b/src/server/pom.xml
@@ -193,17 +193,17 @@
         <dependency>
             <groupId>io.netty</groupId>
             <artifactId>netty-handler</artifactId>
-            <version>4.2.16.Final</version>
+            <version>4.2.17.Final</version>
         </dependency>
         <dependency>
             <groupId>io.netty</groupId>
             <artifactId>netty-transport-native-epoll</artifactId>
-            <version>4.2.16.Final</version>
+            <version>4.2.17.Final</version>
         </dependency>
         <dependency>
             <groupId>io.netty</groupId>
             <artifactId>netty-codec-compression</artifactId>
-            <version>4.2.16.Final</version>
+            <version>4.2.17.Final</version>
         </dependency>
         <dependency>
             <groupId>org.apache.httpcomponents.core5</groupId>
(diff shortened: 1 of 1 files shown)


src/server/pom.xml
```diff
@@ -193,12 +193,32 @@
...
+            <artifactId>netty-codec-compression</artifactId>
+            <version>4.2.16.Final</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.httpcomponents.core5</groupId>
+            <artifactId>httpcore5</artifactId>
+            <version>5.4.3</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.httpcomponents.core5</groupId>
```
burmanm: issue: Why are you bringing so many new dependencies?

ShreyashD333: These are transitive dependencies, I am adding explicit version overrides

burmanm: Why? Also, is there a reason you used 4.2.16 and 4.2.17?

burmanm: Also, this isn't the way to define versions - this is a `<dependencies>` section, this adds dependencies.

ShreyashD333: > Why? Also, is there a reason you used 4.2.16 and 4.2.17?

I simply used the version with available fix the CVE, I will update to 4.2.17

> Also, this isn't the way to define versions - this is a <dependencies> section, this adds dependencies.

Could you suggest the way to define versions ?

burmanm: > I simply used the version with available fix the CVE, I will update to 4.2.17

Always update to the latest version. Otherwise your next scan would have returned all the CVEs fixed in the 4.2.17. This buys us at least a little bit of time before we need yet another round.

Transitive versions and overrides are done in the `<dependencyManagement>` section. 

But why do you need them?

ShreyashD333: These transitive package has vulnerabilities reported, so I am overriding them to the version which has vulnerabilities fixed.
The parent libraries that carry the vulnerable transitives have not yet released new versions that bundle the fixed transitive versions

ShreyashD333: Moved transitive package version overrides to `<dependencyManagement>`

burmanm: > The parent libraries that carry the vulnerable transitives have not yet released new versions that bundle the fixed transitive versions

Which package is that?

ShreyashD333: io.dropwizard:dropwizard-client
io.dropwizard:dropwizard-metrics-graphite

burmanm (approved):

